PR #70, opened 2013-03-25
IGNORE_TEST correctly output in JUnit output

UlkUpp:
When using IGNORE_TEST and output as JUnit, the output will be "skipped"
to make such test cases show up correctly in CI systems, e.g. Jenkins.


Andne:
I suspect that this pull request can be closed, I submitted #463 that got merged a a while ago that I think had the same effect.

basvodde:
Ok, everyone ok with closing this then?

PR #458, opened 2014-12-26
Test for -p option

arstrube:
This should fail.


basvodde:
I think it did fail before :)

arstrube:
Is it okay then to test without CppUTestExt for now?

basvodde:
Yeah, I guess so :(

arstrube:
Any other reason why this pull request should not be merged?

basvodde:
Other than that I didn't know it was ready... uhm... no :)

PR #463, opened 2015-01-10
Added support for outputting skipped tests to junit

### LINKED ISSUE #462: JUnit Outputs not reporting IGNORE_TEST as skipped/ignored
So I've been working on the other issue I reported (#461), and set up a build on teamcity and I found that teamcity isn't reporting any tests as ignored (it parses the junit output) even though I can see ignored tests in the normal output.  It looks like the JUnitTestOutput class doesn't recognize when a test is an ignored test instead of a running test, so it can't flag the test as a result.

Andne:
Fixes #462.  It's not a very good solution, just checks what character the progress indicator is, but it works.  It looks to me like there is some unneeded coupling going on between the TestOutput logic and the UTestShell class since it knows the character that gets printed.  I think that it should probably just know it's status and let the TestOutput class decide which character that means gets printed, or in the case of JUnit Output, what to put into the file.

### DIFF
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -136,7 +136,7 @@ void JUnitTestOutput::printCurrentTestStarted(const UtestShell& test)
         impl_->results_.tail_ = impl_->results_.tail_->next_;
     }
     impl_->results_.tail_->name_ = test.getName();
-    if (*(test.getProgressIndicator()) == '!') {
+    if (!test.willRun()) {
         impl_->results_.tail_->ignored_ = true;
     }
 }
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -34,13 +34,14 @@
 struct JUnitTestCaseResultNode
 {
     JUnitTestCaseResultNode() :
-        execTime_(0), failure_(0), next_(0)
+        execTime_(0), failure_(0), ignored_(false), next_(0)
     {
     }
 
     SimpleString name_;
     long execTime_;
     TestFailure* failure_;
+    bool ignored_;
     JUnitTestCaseResultNode* next_;
 };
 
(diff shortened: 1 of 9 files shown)


src/CppUTest/JUnitTestOutput.cpp
```diff
@@ -135,6 +136,9 @@ void JUnitTestOutput::printCurrentTestStarted(const UtestShell& test)
         impl_->results_.tail_ = impl_->results_.tail_->next_;
     }
     impl_->results_.tail_->name_ = test.getName();
+    if (*(test.getProgressIndicator()) == '!') {
```
basvodde: This is kind of an ugly way of discovering ignored. Would it be better to add an willRun method or so to the uTest class?

src/CppUTest/Utest.cpp
```diff
@@ -596,6 +596,11 @@ IgnoredUtestShell::IgnoredUtestShell()
 {
 }
 
+IgnoredUtestShell::IgnoredUtestShell(const char* groupName, const char* testName, const char* fileName, int lineNumber) :
```
basvodde: Wasn't this there yet?

src/CppUTest/JUnitTestOutput.cpp
Andne: Probably, I was at the time just trying to make it work.  Had thoughts of adding a status to Utest that would indicate not run, passed, failed or skipped, but never tried to do anything with it.

src/CppUTest/Utest.cpp
Andne: I got linker errors about it missing until I added it, so I guess it wasn't.

basvodde: Interesting.

src/CppUTest/JUnitTestOutput.cpp
basvodde: Could you change it still? This way the JUnitOutput would break down if we change the ignored indicator on the command line. I'd say that is a weird side-effect :)

Andne:
I changed it to depend on a willRun function.  I also removed the progress indicator from UtestShell and TestResult since it seemed to be just a function of the console test output and not actually part of the test itself, so having one outputter being told what to print and the other knowing it based on data didn't seem quite right to me.

basvodde:
Ok, interesting.

PR #468, opened 2015-01-14
Fix VC2008 Debug and Release configurations

arstrube:
TO-DO:

- Debug target should be linked with CppUTestd.lib, but this results in linker errors
- Release target has linker errors
- Inclusion of AllTests.h leads to linker problems (and linking the tests into a separate library doesn't seem to solve them)

@Andne - could you perhaps help me fix this up?


examples/AllTests/AllTests.cpp
```diff
@@ -54,4 +54,4 @@ int main(int ac, char** av)
     return CommandLineTestRunner::RunAllTests(ac, av);
 }
 
-#include "AllTests.h"
+// #include "AllTests.h"
```
basvodde: Adding a comment here doesn't seem good ?

Andne: There's an extra test in AllTests.h that isn't included in the project - HelloTest.cpp.  Looks like the file got removed when you were renaming the project file?

Andne:
Oye, MSVC's method of deciding which libraries to include is really confusing, ended up having to change a preprocessor define at one point (examples/AllTests/AllTests.vcproj has _DEBUG in Release mode).  One thing that I think is always going to cause a problem is which library to build against.  Unless all files were compiled with the same option (Multi-threaded, multi-threaded debug, dll, debug dll), then the linker is going to error out.  I think I've narrowed some of the logic down to this:
- Standard C Library - based on the option in the Code Generation pane, can be suppress through the /Zl option.
- Standard C++ Library - static vs. DLL based on the option in the Code Generation pane, Debug version based on the existence of _DEBUG in the preprocessor.  Isn't suppressed by /Zl, looks like it comes from pragmas in use_ansi.h (seems like a funny name for that file), which comes from yvals.h, which comes from a number of C++ headers.  Could always define _USE_ANSI_CPP to suppress that I suppose, but it's a bit hackish ;)

Any opinions on what the correct library to build against should be?  I'm not sure what your thoughts are for a release process, is the assumption that people will always build the library themselves?

arstrube:
Thanks a lot, @Andne - especially for the hint with HelloTest.cpp (duh - could have spotted that one myself).

I cleaned up the project files as far as I could. I made them use ./$(ConfigurationName)/ consistently, so as to avoid linking the Debug ApplicationLib.lib instead of the Release  ApplicationLib.lib.

Both Debug and /O2 Release compile, link and run without failures now. So I hope this can be merged

There are a few things I still want to do:
- Put tests into a separate AllTestsLib.lib (which is what the #include "AllTests.h" is really for)
- Make the Debug build link CppUTestd.lib (as it should)
- Make additional project files & solution for VS2010

I hope the above will help me to find the problem with the CppUTestd.lib

jwgrenning:
Can you add Visual Studio 20113 to the list too?

Andne:
Visual Studio 2013 can use the Visual Studio 2010 files, the only difference is what version of the toolset it tries to compile against.  One concern I'd have with trying to create files for every version of Visual Studio is the difficultly in maintaining all them, which seems to be proving to be a challenge already.

arstrube:
@jwgrenningI am sorry; I don't do VS2013.  ;-).

I installed VS2012 once. It was hate on the first sight, plus I am still struggling with errors in VS2008/2010 that were caused by the installation (and are not fixed by deinstallation).

@Andne - you are very right. I would say though, for practical reasons, we should always provide the "last version that can be imported into the current version". Since the CppUTest solution is also in 2010 format, I think that would be good for the examples, also.

Andne:
I agree that a 2010 example would be good to have.  I think that because 2012 and 2013 (and probably 2015, but I haven't used that version yet) can use 2010 files that there is little reason to create specific ones for it.  The only change that needs to happen to the file to use the newer version is a prompt to change the default PlatformToolset, which Visual Studio prompts you for the first time you open the project.  Should we maybe rename the solution file to reflect this?  The project files have a different file extension, so they do not have anything about version in the name.

arstrube:
I was thinking about calling it CppUTestExample_VS2010.sln and distinguishing the projects by extension. Personally, I would opt not to check in any 2012++ stuff. 

Were you thinking of a name like ..._VS201x.sln?

Andne:
I agree with both thoughts, nothing specific to 2012, but change the VS2010 name to indicate that it can be used by the newer versions as well.  Letting the extension distinguish the projects makes sense to me.

arstrube:
@basvodde should I change all names to ..._VC201x when adding the example solution (also for the CppUTest solution)? I think they should be consistent.

basvodde:
You did so already, right?

arstrube:
Yeah, I figured you'd probably say I should....

PR #470, opened 2015-01-16
Vs201x

arstrube:
I forgot to commit the .vcxproj files.

I also tried to fix warning LNK4217 but didn't succeed yet. At least, the runtime library options should now be more consistent with the Debug / Release configurations. Any idea, @Andne?

### DIFF
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -94,7 +94,7 @@ copy $(OutDir)vc100.pdb lib\vs2010
       <InlineFunctionExpansion>OnlyExplicitInline</InlineFunctionExpansion>
       <PreprocessorDefinitions>WIN32;NDEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <StringPooling>true</StringPooling>
-      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
       <FunctionLevelLinking>true</FunctionLevelLinking>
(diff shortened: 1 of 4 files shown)


Andne:
One, we apparently could use a CI build for the visual studio projects too.... ;)  I'm kind of working on that in with my fork, have it running on a teamcity instance at home, been trying to create a set of builds for the different versions so I can bundle it up into a nuget package.  Unfortunately that has some other challenges around the file tree structure too.  That's why I was asking about the runtime libraries in the other pull request, if releasing as an artifact like that, what to do there becomes more important.

I'll take a look when I get a chance, but it probably won't be until tomorrow at least.

arstrube:
I totatlly agree. But would it even be possible to do on Travis? BTW I am doing the VC stuff on my work computer. Couldn't get myself to install it on my MacBook, although I suppose it would do fine in VirtualBox... I don't won't to bloat my current, extremely lean XP VM and I haven't been willing to take the time to set up a dedicated VM for Studio...

arstrube:
Oh and don't worry about timing. I will only get back to this on Monday :)
@basvodde it's okay to merge as is; the previous pull request actually lacked the .vcxproj's, which is no good. Apologies for that.

Andne:
No, I think Travis can only do Linux (and apparently OSX from their docs) builds.  It does look like Appveyor is basically the Windows version of Travis, so maybe there's opportunity there?  They charge I think, but it says free for open source projects too.

I'm the opposite of you, have more visual studio versions and tools on my home desktop than my work computer.  For the most part I stay in Windows, except my Teamcity master and Plex server are Ubuntu VM's running under Hyper-V.

PR #471, opened 2015-01-18
Added the release folders to the ignore list

Andne:
When building the VS201x projects, I noticed that git status shows the release folders as being changes, so I added them to the ignore list.

@arstrube I don't see your linker warnings, I built both Debug and Release configurations of CppUTest and the Examples.  One thing I did just remember, there was a known issue with VS2010 and VS2012 coexisting that was fixed in VS2010SP1, so if you didn't have the service pack installed before installing VS2012, that may be why you had problems.  One possible way to clean things up may be to install Service Pack 1 if you don't already have it installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,9 @@ ErrorLogs
 testResults
 .metadata
 Debug
+Release
 */Debug/*
+*/Release/*
 *.exe
 *.obj
 *.ncb


arstrube:
I am glad this was only a local problem on my machine.

I'll have to talk to our IT guy about the update because apparently it's tied into Windows update, which is administrated by the IT department. I got most of the really nasty warnings to go away using some magic that someone suggested on the Web, but installing SP1 would definitely be better.

Adding the Release folder to the ignore list is an excellent idea. Thanks.

Andne:
I don't think I've ever seen VS2010SP1 come down via Windows Update, I just download it from msdn and run it - now I have a copy of the ISO version on my network, so I just use that.

arstrube:
Ah, good to know. I clicked some link in VS2010 and was sent to the Windows Update site :(

PR #473, opened 2015-01-26
Use IGNORE_TEST() on incompatible systems.

arstrube:
Up to now, the 64 bit pointer printing test would have been silently omitted on 32-bit systems.

I suggest replacing this with an empty IGNORE_TEST() on incompatible systems.

Note: This drops my hack for 64 bit Windows. I consider it a hack, rather than a test of anything useful. So I am rather glad to see it gone.

The number of testcases should now be identical (with regard to this particular conditional compilation), regardless of target system.

### DIFF
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -560,20 +560,15 @@ TEST(SimpleString, CollectionWritingToEmptyString)
     STRCMP_EQUAL("", col[3].asCharString());
 }
 
-#ifdef CPPUTEST_64BIT
-#ifdef CPPUTEST_64BIT_32BIT_LONGS
+#if !defined(CPPUTEST_64BIT) || \
+     defined(CPPUTEST_64BIT_32BIT_LONGS)
 
 /*
  * Right now, the 64 bit pointers are casted to 32bit as the %p is causing different formats on
- * different platforms. However, this will need to be fixed in the future and then this test ought
- * to be deleted.
+ * different platforms. However, this will need to be fixed in the future.
  */
-TEST(SimpleString, _64BitAddressPrintsCorrectly)
+IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
 {
-    char* p = (char*) 0xffffffffu;
-    SimpleString expected("0x23456789");
-    SimpleString actual = StringFrom((void*)&p[0x2345678A]);
-    STRCMP_EQUAL(expected.asCharString(), actual.asCharString());
 }
 
 #else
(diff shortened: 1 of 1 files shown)


tests/SimpleStringTest.cpp
```diff
@@ -560,20 +560,15 @@ TEST(SimpleString, CollectionWritingToEmptyString)
     STRCMP_EQUAL("", col[3].asCharString());
 }
 
-#ifdef CPPUTEST_64BIT
-#ifdef CPPUTEST_64BIT_32BIT_LONGS
+#if !defined(CPPUTEST_64BIT) || \
+     defined(CPPUTEST_64BIT_32BIT_LONGS)
```
basvodde: Curious. Shouldn't there be another ! ? (or additional parenthesis?)

Andne: I don't think so, from what I can tell the first block is for systems that don't support 64bit at all or still have a 32bit long on a 64bit platform (windows x64).  It has an empty IGNORE_TEST in it.  The #else condition has the functional test in it that check the 64-bit long behavior.

arstrube: That's the idea.

I didn't like to different #else block for essentially the same thing,

basvodde: Ok

arstrube: @basvodde -- on second thoughts, you were right. This felt off even to me in the end. Things are getting mixed that shouldn't.....

basvodde: *grin*. Ok, well please do another pull request to fix it ;)

jwgrenning:
Sorry to not have looked all the messages in this thread.  instead of 
conditional compilation, how about segregating 64 bit test in their own 
file, and building it or not based on platform.  It won't preserve the 
test count, but will get rid of objectionable conditional compilation.

James

--------------------------------------------------------------------------
James Grenning - Author of TDD for Embedded C - 
pragprog.com/titles/jgade/
wingman-sw.com
wingman-sw.com/blog
twitter.com/jwgrenning
facebook.com/wingman.sw
[wingman 
software](http://wingman-sw.com)

arstrube:
I have asked the question in another thread, more generally, about testing platform specific code in platform specific test cases... This one doesn't quite fit that idea, though, because we are not testing platform code. We have segregated the isNAN tests, if I remember correctly.

However, personally, I find the changing number of test cases disconcerting and would therefore prefer conditional defines cum IGNORE_TEST().

For example, I compile under Cygwin -> 368 tests. I compile under MinGW ->367 tests. Hmm. I compile under Visual Studio -> 369 tests (or whatever). 

So, much as I hate conditional defines in gereral, I would prefer to see them here.. (Not inside of test cases, though).

basvodde:
Agree, as long as there aren't too much. So far it is very limited.

PR #474, opened 2015-01-27
Handling the 64 bit test case in a cleaner way

arstrube:
Rewrote test so it will compile on all 32 bit and 64 bit systems.

We now have the following distinct scenarios:
1. Test will pass on 64 bit systems with 64 bit longs
   -> TEST().
2. Test should pass on 64 bit systems with 32 bit longs, but fails
   due to a known implementation problem. Needs to be fixed.
   -> IGNORE_TEST() for now.
3. Test cannot pass on 32 bit systems
   -> IGNORE_TEST() always.

### DIFF
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -573,10 +573,11 @@ TEST(SimpleString, _64BitAddressPrintsCorrectly)
 
 #else
 /*
- * The above test case should also pass on 64 bit systems with 32 bit longs, 
- * but would actually fail due to implementation problems. Right now, the 64  
- * bit pointers are casted to 32bit as the %p is causing different formats
- * on different platforms. However, this will need to be fixed in the future.
+ * This test case should pass on 64 bit systems with 32 bit longs, 
+ * but actually fails due to an implementation problem: Right now,
+ * the 64 bit pointers are casted to 32bit as the %p is causing 
+ * different formats on different platforms. However, this will
+ * need to be fixed in the future.
  */
 
 IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
@@ -586,7 +587,7 @@ IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
 #endif
 #else
 /*
- * The above test case would necessarily fail on 32 bit systems.
+ * This test case necessarily fails on 32 bit systems.
  */
 IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
 {
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -565,9 +565,9 @@ TEST(SimpleString, CollectionWritingToEmptyString)
 
 TEST(SimpleString, _64BitAddressPrintsCorrectly)
 {
-    char* p = (char*) 0xffffffff;
-    SimpleString expected("0x100000000");
-    SimpleString actual = StringFrom((void*)++p);
+    char* p = (char*) 0x0012345678901234;
+    SimpleString expected("0x12345678901234");
+    SimpleString actual = StringFrom((void*)p);
     STRCMP_EQUAL(expected.asCharString(), actual.asCharString());
 }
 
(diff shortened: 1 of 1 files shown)


basvodde:
Should we then get the other test back that you deleted?

arstrube:
Why? 

Would you then (assuming the 32 bit long problem was fixed) use the deleted one for 64/64 and the one in this pull request for 64/32?

E.g. I could then also flesh out the ignored test case that *should* pass. Using the new code there, it would compile, but fail if it wasn't ignored.

Hmm. Maybe not at all a bad idea.

arstrube:
We would then have:
```c++
TEST() // Original test case for pure 64 bit systems
IGNORE_TEST() // 32 bit long compatible test case
IGNORE_TEST() // Always empty for 32 bit systems
```

arstrube:
Should I do that?

basvodde:
Sorry, I'm slow. In this commit: https://github.com/cpputest/cpputest/commit/a4df3a62293d43bdf02c0039ed1c3aef16239ffe#diff-82d4bffd3899d43667ae495ad4b04a9a

You removed one test. Shouldn't we have that one back?

arstrube:
Yes, I think we should have it back.

But then, we can give pure 64 bit systems the old test back also (without the funky 32 bit max initialization constraint workaround).

arstrube:
P.S. Right now, I don't have a 64 bit Windows gcc environment ready for testing (we changed all back to 32 bit at work) and my Mac is 64bit (32bit Windows VM). To be frank, I find that combination detestable, so I don't use it normally (the only reason to use 64bit apps on 64 bit Win is if one of them needs that much memory all on its own :-)

arstrube:
Okay, I managed to test it -- sort of.
1. un-ignore test
2. compile with mingw 64 bit
3. hack away the pthread stuff
4. "fix up" Simplestring (see below) for Windows
5. Test passes.
```c++
SimpleString HexStringFrom(const void* value)
{
    return StringFromFormat("%llx", (long long)value);
}
```
So, this can be considered fully tested 
1. 32 bit Windows (my usual setup) - I?LP32
2. 64 bit Windows with 32 bit longs (MinGW 64 bit) - LLP64 / IL32P64
3. Travis - LLP64 / I32LP64

basvodde:
Originally the test wasn't IGNOREd, but now it is. Why is this?

arstrube:
Could you maybe comment in the code? I am not sure I understand what exactly you are referring to here.

If you look at the new code in context (not the diff), the comments would make pretty clear what's intendend (or else, I failed miserably with this pull request :-)

arstrube:
Ah, now I get it. The diff is a bit confusing.

Before, the testcase was used for L64P64 (although the test is compatible with L32P64).

Now I am distingishing L64P64 with its own test case.

The old L32P64 test case is now used for L32P64 systems (Windows), and it needs to be ignored until SimpleString is fixed (one such fix would be using `%llx` and `(long long)`, as detailed by the comment.

As I mentioned above, take a look at the new code in context (not the diff).

arstrube:
This way we make explicit that 

1. One `IGNORE_TEST()` is out of necessity (no way this will ever pass on L32P32 systems)
2. The other `IGNORE_TEST()` is for code that needs fixing

Bunching these together like in my last pull request makes for less code and "fewer" conditional defines, but contradicts the principle of making explicit why we ignore certain tests, and what ultimately needs to be fixed.

basvodde:
But it wasn't IGNOREd before? So, why does it need to be ignored now?

arstrube:
The conditional define logic is different.

Before, L32P32 and L32P64 was bunched together (which it shouldn't, in my opinion) and totally IGNOREd.

So the not-IGNORED test was only run for L64P64 (now covered by the new test case).

Now, the not-IGNORED test is run on L32P64 systems, where it will compile but fail (until I twist your arm to use `long long` in SimpleString). Hence it now needs to be IGNOREd.......

basvodde:
Ok, I'll just merge it then :)

arstrube:
Thx :)

This is the test failing when built against MinGW-64:

fail

and now it is ignored:

ignore

PR #489, opened 2015-03-02
Coverage

arstrube:
Bring up statement coverage to 95+ % by adding missing tests and removing some dead code.

Another small improvement will be achieved when I add tests for running tests in a separate process.

However, 100% cannot be reached because of some methods failing unconditionally.

Also, defaultCrashMethod cannot be exercised without, well, crashing ;-).

### DIFF
--- a/tests/SetPluginTest.cpp
+++ b/tests/SetPluginTest.cpp
@@ -82,6 +82,22 @@ TEST(SetPointerPluginTest, testUnoverridden_testBody)
     delete tst;
 }
 
+class defaultCreateTestUtestShell: public UtestShell
+{
+public:
+};
+
+TEST(SetPointerPluginTest, testUnoverridden_createTest)
+{
+    defaultCreateTestUtestShell *tst = new defaultCreateTestUtestShell();
+    ;
+    myRegistry_->addTest(tst);
+    myRegistry_->runAllTests(*result_);
+    LONGS_EQUAL(0, result_->getFailureCount());
+    LONGS_EQUAL(0, result_->getCheckCount());
+    delete tst;
+}
+
 class FunctionPointerUtest : public Utest
 {
 public:
--- a/tests/SetPluginTest.cpp
+++ b/tests/SetPluginTest.cpp
@@ -49,28 +49,36 @@ TEST_GROUP(SetPointerPluginTest)
     }
 };
 
-class nonoverriddenTestBodyUtest : public Utest
+class defaultTestBodyUtest : public Utest
 {
 public:
+    void setup() _override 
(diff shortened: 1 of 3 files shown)


Andne:
Not sure if the scripts would support this or not, but some of the parsing scripts I've seen for gcov include the ability to flag a line as excluded from coverage, so then it doesn't get included In any metrics.  From what I remember, cpputest has some of its own scripts for coverage parsing, but possibly they could be updated to support such a mechanism?  Kind of wish gcov had some ability like that natively.

arstrube:
I've never heard of such a thing, except for commercial tools. I am very intersted in seeing how this could be done.

arstrube:
Apparently lcov is needed to achieve such a feat. I have so far tried to avoid it....

As for CppUTest, I dug into the Travis output a bit and came up with the following. My own log result (gcc 4.8.2)
```
Lines executed:95.25% of 316
```
Travis output (clang):
```
94.26%   ../src/CppUTest/Utest.cpp
```
Travis output (gcc):
```
 92.08%   ../src/CppUTest/Utest.cpp
```

arstrube:
The differences are mostly due to Automake building with -O2. Generally not necessary, in my opinione, and definitely not a good idea when looking at code coverage. Some functions have been inlined and optimized away.

arstrube:
The last percent goes to me running the tests with -v, and Automake not.

tests/SetPluginTest.cpp
```diff
@@ -49,16 +49,65 @@ TEST_GROUP(SetPointerPluginTest)
     }
 };
 
+class defaultTestBodyUtest : public Utest
+{
+public:
+    void setup() _override 
+    {
+        CHECK(true);
```
basvodde: Why this CHECK(true) ?

tests/SetPluginTest.cpp
```diff
@@ -49,16 +49,65 @@ TEST_GROUP(SetPointerPluginTest)
...
+};
+
+TEST(SetPointerPluginTest, testUnoverridden_testBody)
+{
+    defaultTestBodyUtestUtestShell *tst = new defaultTestBodyUtestUtestShell();
+    ;
+    myRegistry_->addTest(tst);
+    myRegistry_->runAllTests(*result_);
+    LONGS_EQUAL(0, result_->getFailureCount());
+    LONGS_EQUAL(2, result_->getCheckCount());
```
basvodde: Ah, never mind above question :)

basvodde:
Nice, thanks

arstrube:
You are welcome :-)

PR #491, opened 2015-03-03
1. refactor: Fix and simplify misplaced test

arstrube:
Couldn't live with the original test :-)

### DIFF
--- a/tests/UtestTest.cpp
+++ b/tests/UtestTest.cpp
@@ -153,9 +153,8 @@ class defaultUtestShell: public UtestShell
 {
 };
 
-TEST(UtestShell, coverageForDefaultMethods)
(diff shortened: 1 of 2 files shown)


tests/UtestTest.cpp
```diff
@@ -149,6 +149,19 @@ TEST(UtestShell, TestStopsAfterSetupFailure)
     LONGS_EQUAL(0, stopAfterFailure);
 }
 
+class defaultUtestShell: public UtestShell
+{
+};
+
+TEST(UtestShell, coverageForDefaultMethods)
+{
+    // This test covers UtestShell::createTest() and Utest::testBody()
```
basvodde: Could you change the test name to include this knowledge instead of using a comment?

arstrube: I was trying to avoid that monstrosity :P....

PR #492, opened 2015-03-03
3. fork: Tests for PlatformSpecificRunTestInASeparateProcess()

arstrube:
This is ready so far ... I'm still thinking about a few missing test cases and fleshing out the IGNORE_TESTs.

I would like some feedback on handling the MinGW angle. If tests fail with -p on MinGW, it is a lot easier to test. Currently they print an additional message. From a users perspective, I like neither version.

As a user, I would the test like to inform me once that -p isn't supported and refuse to run altogether (what's the point in running anyway?)


arstrube:
Also, I'm not sure about a location. Maybe put this with the other tests? It should work fine for Mac, Cygwin and Unix derivates but what about Visual Studio? What about more exotic platforms, where I can't even run these tests (anymore)?

arstrube:
It wouldn't break the Travis build. We'd have to rely on users to tell us if the tests misbehave on one of the rarer platforms. Currently, they are only run for Gcc.

arstrube:
Actually, as I'm doing these tests I have come to realize that in most cases, I wouldn't want to run an entire test suite with -p anyway. I would want to enable running in a separate process per test group or even test cases (and yes, I will test the DefaultCrashMethod, too).

Would this be hard to implement, without a TestTestingFixture? It would then be more appropriate for a test to fail, if it can't do it (MinGW).

Also, it is a bit wasteful to run 500 tests with -p, when doing so for one or two test cases would do.

-p should still remain, I think.

arstrube:
@basvodde
(1) Any idea why the cmake build appears to hang? It runs the first test suite to completion, then ?
(2) Would you create tests for each platform individually (as in this pull request) or would you rather have one test along with the other tests, with possibly a few more conditional compilations in it?

I actually tend to favor the second approach, because this would be a nice indicator how platform-independant we really are. Even though the implementations are different, the tests should be mostly the same, since the interface is the same.

arstrube:
```
1: ........................
1: OK (424 tests, 412 ran, 722 checks, 12 ignored, 0 filtered out, 12 ms)
1: 


No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

basvodde:
1) I have no idea, haven't looked at it :(
2) Yes, not have platform specific tests. I'd like to avoid that as long as possible.

arstrube:
Okay, will change it accordingly.

I will #ifdef the -p process so they run only for gcc at this point.

As for the hang, having slept over it, I think killing the stopped process before the test ends would be a good idea anyway. Just strange how this only effects cmake - under cmake, the test seems to wait for the child process to finish, unlike any other build envirionment I have tested.

arstrube:
I put the offending test in a separate pull request.

basvodde:
There are still platform specific tests here? Weren't you going to remove those or am I confused? :)

arstrube:
Well yes, although I was going to move, rather than remove them. There will be one UTestPlatformTest for all and not several as is still the case in this pull request.

Haven't come round to it yet though. My bad.

arstrube:
I do believe the new run in separate process code should be tested.

basvodde:
Its ok, just wasn't sure I should merge it or not :)

PR #496, opened 2015-03-05
Coverage

arstrube:


basvodde:
Hrm, ok. So the variable change is purely to add the test for the coverage. I'm a bit uncomfortable with it, but ok. I'll merge it.

arstrube:
Thanks :) I could up it to 1000 though. I think that would still be barely noticable.

We changed the 0.00001 percent chance of this happening with no test to 0.01 with test :D

If you can think other of other ways to inject the value of n-1 into the counter, that would probably greatly improve my testing skills ;-). 

Or we could improve the list search algorithm, so looping 99,999 times would be no problem - and then re-use it to make mocks more effient, too. I recall an issue reporting problems with 10,000 mock cals....

basvodde:
I guess 100 will do.

Yeah, the Mocks still have an issue with many mocks. This need to be resolved at some point.

PR #498, opened 2015-03-08
Applied change from #494 to Visual C++

Andne:
PlatformSpecificRunTestInASeperateProcess was modified to be a function pointer as part of pull request #494.  That change wasn't also applied to the VisualCpp copy of UtestPlatform.cpp, so I applied it to this file as well.


arstrube:
Thanks @Andne - much appreciated!

PR #501, opened 2015-03-09
These tests will bring up statement coverage for UTestPlatform.cpp (GCC) to 100%

arstrube:
They are rather simple-minded and don't do any checking (yet?). Some observations:

1. `PlatformSpecificFOpen()`, `PlatformSpecificFPuts()` - it seems like those functions probably don't get used anywhere in CppUTest. One could write some stuff in an actual temp file, read it back and have the test compare it. However, temporary files (that get automatically unlinked) are not supported under Cygwin, as far as I know.

2. `GetPlatformSpecificTimeString()` - this doesn't seem to be used, either. The interface does not allow for any kind of sensible test.

3. `PlatformSpecificMutexLock` and `PlatformSpecificMutexLock` - better tests might be possible for this. Personally, I have no experience so I left it at this.


basvodde:
Commented on this change in another pull request. I'm uncomfortable with this because the tests don't test anything and they are purely created for coverage sake

PR #503, opened 2015-03-09
Add TEST_IN_SEPARATE_PROCESS() macro

arstrube:
Using this macro, a single test can easily be run in a separate process on platforms that support fork() including Cygwin.

In this way, a branch that crashes can be safely tested without terminating the tests.

It seems less useful than I thought, though, because an anticipated crash would still cause the test to fail.

### DIFF
--- a/tests/UtestPlatformTest.cpp
+++ b/tests/UtestPlatformTest.cpp
@@ -30,6 +30,23 @@
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTest/PlatformSpecificFunctions.h"
 
+TEST_GROUP(UTestPlatformsTest)
+{
+};
+
+TEST(UTestPlatformsTest, PlatformSpecificFileFunctionsDontCrash)
+{
+    PlatformSpecificFile file;
+    file = PlatformSpecificFOpen("/dev/null", "rw");
+    PlatformSpecificFPuts("Hello", file);
+    PlatformSpecificFClose(file);
+}
+
+TEST(UTestPlatformsTest, GetPlatformSpecificTimeStringIsCalledOnce)
+{
+    GetPlatformSpecificTimeString();
+}
+
 TEST_GROUP(UTestPlatformsTest_PlatformSpecificRunTestInASeperateProcess)
 {
     TestTestingFixture fixture;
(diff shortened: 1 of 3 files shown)


tests/UtestPlatformTest.cpp
```diff
@@ -30,6 +30,31 @@
...
+{
+    PlatformSpecificFile file;
+    file = PlatformSpecificFOpen("/dev/null", "rw");
+    PlatformSpecificFPuts("Hello", file);
+    PlatformSpecificFClose(file);
+}
+
+TEST(UTestPlatformsTest, GetPlatformSpecificTimeStringIsCovered)
+{
+    GetPlatformSpecificTimeString();
```
basvodde: This is not a test, of course :)

tests/UtestPlatformTest.cpp
```diff
@@ -30,6 +30,31 @@
...
+TEST_GROUP(UTestPlatformsTest)
+{
+};
+
+TEST(UTestPlatformsTest, PlatformSpecificFileFunctionsDontCrash)
+{
+    PlatformSpecificFile file;
+    file = PlatformSpecificFOpen("/dev/null", "rw");
+    PlatformSpecificFPuts("Hello", file);
+    PlatformSpecificFClose(file);
```
basvodde: No test here, just written for coverage. I'd prefer not having coverage than having meaningless tests...

basvodde:
Yeah, I'm very hesitant to add this macro, as it might make it look like a good idea... whereas I've never seen when this is a good idea.

If we add it, I'd suggest to add it to the extensions (like ordered test) and not to the CppuTest core.

tests/UtestPlatformTest.cpp
arstrube: I see your point. But then, it's about as much of a test as the IGNORE_TEST() one I was complaining about a while back ;-). Also, it proves that it doesn't crash :-/

arstrube: I'd prefer changing it so it gets passed seconds and can be genuinely tested.

Also, you forgot about option #3: All of these functions can be removed entirely. They are dead code; not used anywhere.

arstrube:
Okay. Yes, it would definitely be better in the extensions. I'm not so sure anymore about its usefulness myself...

arstrube:
Actually, I came to realize that this feature would only be really useful in connection with the TestTestingFixture. 

But if I use the TestTestingFixture, then I don't need this :-(

basvodde:
Ok, I sort-of suspected that. Good to know that it isn't hard to implement though :)

arstrube:
I'll keep it around in my repo for reference :-)

Issue #480, opened 2015-02-06
Successful CHECK_THROWS does not add to check count


klusark:
When using CHECK_THROWS, it only increments the "checks" count when the test is successful. For example:

    .................................................
    OK (49 tests, 49 ran, 44 checks, 0 ignored, 0 filtered out, 21 ms)

And here we have one exception test failing.

    .................................
    filename.cc:128: error: Failure in TEST(test_group, test_name)
            expected to throw std::exception
    but threw nothing

    ................
    Errors (1 failures, 49 tests, 49 ran, 45 checks, 0 ignored, 0 filtered out, 50 ms)

Notice that it has 44 checks in the first case, and 45 in the second.

arstrube:
I looked into this one. The reason for this bug is easy enough to see (use of `UtestShell::fail()` rather than `UtestShell::failWith()`, which unconditionally increments the check count, but only in case of failure.

I tried to use `UtestShell::failWith()` instead, but that is tricky since the whole thing is inplemented as a macro rather than a method of UtestShell.

@basvodde can you see any solution for this dilemma?

arstrube:
Not to worry. I found another way...

arstrube:
I believe I fixed this one, too.

basvodde:
Closing.

PR #504, opened 2015-03-09
Fix issue #480

arstrube:


basvodde:
I'll merge this, but it is hard to do so as the commits are all build on each other, so we'll first need to check the functions purely added for coverage...

arstrube:
I'll rebase it. The only other commit it really needs is the fix to cmake.

Issue #168, opened 2013-08-23
weird behavior in JUnitOutput


simshi:
I have a file contains something like:

> TEST(groupA, case1) { ... }
> TEST(groupA, case2) { ... } 
> TEST(groupB, case1) { ... } 
> TEST(groupA, case3) { ... } 

output xml of groupA (cpputest_groupA.xml) will be something like:

``` xml
<testsuite ... tests="1" ...>
<testcase classname="groupA" name="case3">...</testcase>
<system-out></system-out>
<system-err></system-err>
</testsuite>ssname="groupA" name="case2">...</testcase>
<system-out></system-out>
<system-err></system-err>
</testsuite>
```

The same file 'cpputest_groupA.xml' was written twice, first case1 and case2, then overwritten with case3...

A funny bug :-)

The root cause is in TestRegistry::endOfGroup(..):

```
return (test->isNull() || test->getGroup() != test->getNext()->getGroup());
```

Andne:
I've been chasing why my CI build reports a different number of tests than I see on my computer, and it appears to be related to this issue (I use JUnit output for CI servers).  I was able to correct a couple by re-arranging some tests, but one is not quite that simple.  It appears that there's an interaction between this issue and `TEST_ORDERED` that causes trouble when when a test group has a mixed combination of `TEST` and `TEST_ORDERED`, which is the case in OrderedTestTest.cpp.  My test order appears to be such that `TEST(TestOrderedTestMacros, NormalTest)` runs at the very beginning of the test list and the `TEST_ORDERED(TestOrderedTestMacros, Test*)` tests run at the very end of the test list.  The end result is that my CI server never notices the existence of `TEST(TestOrderedTestMacros, NormalTest)`

I'm not quite sure how to fix this at the moment, but wanted to let people know that it existed.

itavero:
@basvodde Any idea what the status of this issue is?

basvodde:
AFAIK, nobody picked up this issue. It would require some nice refactoring of the CppUTest core to make TestGroup a class by itself :)

itavero:
I'm pretty sure there are other ways to fix this.
But, having looked at how the JUnit output is currently generated, I do agree that refactoring is probably the right solution (instead of adding more logic to the output class).

PR #507, opened 2015-03-10
Workaround to improve junit test count

Andne:
I swapped the order of a few tests around to working around issue #168 and get the junit test count closer to the actual test count.  There's still one test that I cannot easily get because of the `ORDERED_TEST` macros.

### DIFF
--- a/tests/MemoryLeakOperatorOverloadsTest.cpp
+++ b/tests/MemoryLeakOperatorOverloadsTest.cpp
@@ -260,6 +260,20 @@ TEST_GROUP(OutOfMemoryTestsForOperatorNew)
 
 #if CPPUTEST_USE_STD_CPP_LIB
 
+TEST_GROUP(TestForExceptionsInConstructor)
+{
+};
+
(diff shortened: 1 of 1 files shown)


basvodde:
I *can* merge this, but we ought to solve the real problem, of course :)

So far, I've assumed that solving the real problem means to create a TestGroup class and elevate TestGroup to a real entity in CppUTest. The TestRegistry can then just run TestGroups, though for OrderedTest, this would be a bit harder.

Merge this or do it right?

Andne:
I vote for do it right, and I suspect that one or two other things that have seemed a little odd to me may be possible to clean up with such a change (like how the verbose output is generated).  I think that we probably do need test group to be a real entity, and calling the test group to run other tests is probably a good idea.  I suspect there may be other systems out there that prefer all tests from a group running as a single block.  Given that most newer environments (Java, .NET) use classes to create groups of tests, it's pretty much a given that the tests in a group would all be run together without any other groups occurring in-between them.

basvodde:
Close this one then?

PR #510, opened 2015-03-11
Put back vc6 files

arstrube:
Must have been over-zealous :-(


arstrube:
The *.dsp will need updating.

When all is said and done, the *.mak should be regenerated from Visual Studio.

PR #513, opened 2015-03-13
VC6

arstrube:
With this, VC6 will compile and run successfully all tests for CppUTest and CppUTestExt -- without memory leak detection.

If someone other than me knows how to resolve this operator new[] stuff, maybe we could get memory leak detection to work, too.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,5 +9,7 @@
 *.bat text eol=crlf
 *.dsp text eol=crlf
 *.dsw text eol=crlf
+*.dep text eol=crlf
+*.mak text eol=crlf
 *.vcproj text eol=crlf
 *.vcxproj text eol=crlf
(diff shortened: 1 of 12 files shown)


arstrube:
Oops. Looks like there's a crlf problem with the .mak files...

arstrube:
With this, VC6 will compile and run successfully all tests for CppUTest and CppUTestExt -- without memory leak detection.

If someone other than me knows how to resolve this operator new[] stuff, maybe we could get memory leak detection to work, too.

arstrube:
@basvodde - sorry, I really don't know how to change lf -> crlf without the diff looking like this. Since .mak and .dep are generated files anyway, maybe you could tolerate them just this once? .gitattributes is changed, so the next diff should show up normally.

The rest of this pull request is good stuff (and took a lot of debugging to figure out :-)

arstrube:
In MockSupportTest, `teardown()` actually checks for no mock failure.

Some tests leave open expectations, and called `mock().clear()` prior to actual cleanup, in order to get rid of those expectations. This can be confusing.

This commit fulfils any open expectations before calling `teardown()`. `mock().clear()` called in `teardown()` takes care of overall cleanup.

arstrube:
I'm breaking this up in smaller pull requests.

PR #514, opened 2015-03-14
Removed some redundant methods

arstrube:
These are never used anywhere (and also, not tested, of course).

Since it is not likely they would be used by anyone outside of CppUTest, I removed them (rather than adding tests for them).


basvodde:
Good, less stuff :)

PR #517, opened 2015-03-14
Vc6 2 - Fixup tests that crash / fail under VC6

arstrube:
This pull request is the second replacement for #513. It fixes up the tests.

It is interesting that Clang and Gcc produced code that didn't crash / fail. Nevertheless, these fixes make for cleaner code even when we are not compiling them with VC6.

This code will compile, link, run and succeed under VC6, as well as the Travis build.


arstrube:
Closed, as it turned out that we need to use the main() from tests/CppUTestExt/, which will load the MockSupportPlugin to clean up automatically after each test.

PR #519, opened 2015-03-15
Simplestring

arstrube:
Additional tests and a small fix for SimpleString

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -172,7 +172,6 @@ bool SimpleString::containsNoCase(const SimpleString& other) const
     return lowerCase().contains(other.lowerCase());
 }
 
-
 bool SimpleString::startsWith(const SimpleString& other) const
 {
     if (StrLen(other.buffer_) == 0) return true;
@@ -192,6 +191,8 @@ bool SimpleString::endsWith(const SimpleString& other) const
 
 size_t SimpleString::count(const SimpleString& substr) const
 {
+	if(substr.isEmpty()) 
+		return 1;
     size_t num = 0;
     char* str = buffer_;
     while ((str = StrStr(str, substr.buffer_))) {
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -191,11 +191,9 @@ bool SimpleString::endsWith(const SimpleString& other) const
 
 size_t SimpleString::count(const SimpleString& substr) const
 {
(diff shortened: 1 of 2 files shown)


tests/SimpleStringTest.cpp
```diff
@@ -257,6 +257,12 @@ TEST(SimpleString, countTogether)
     LONGS_EQUAL(4, str.count("ha"));
 }
 
+TEST(SimpleString, countEmptyString)
+{
+    SimpleString str("hahahaha");
+    LONGS_EQUAL(1, str.count(""));
```
basvodde: Uh, why?

arstrube: What would you rather expect? How many times is the empty string contained in a string? :P

Behavior so far was some very large unsigned long integer value, possibly MAX_ULONG. This lead to problems with the empty string in replace().

arstrube: This lead to a noticable delay (30 seconds or so), as well as the replace function not working properly for empty strings (the else part was never executed).

arstrube: Returning zero would be fine by me and work okay. But I could't give you any better justification for returning zero than for returning 1...

After all, how many times is the empty string contained in a string ;-).

basvodde: I would have thought 0 :) What does strstr do?

arstrube: strstr returns the entire first string, if the string to search for is the empty string, apparently. Knowning that doesn't really help, does it... I'm fine with zero.

arstrube: This behavior of strstr spells trouble for the while loop in count, I guess.

src/CppUTest/SimpleString.cpp
```diff
@@ -194,7 +193,7 @@ size_t SimpleString::count(const SimpleString& substr) const
 {
     size_t num = 0;
     char* str = buffer_;
-    while ((str = StrStr(str, substr.buffer_))) {
+    while (*str && (str = StrStr(str, substr.buffer_))) {
```
basvodde: strstr returns the same string as it passed. That would suggest that the count would be equal to size(). What do you think about that? Which means the *str shouldn't be needed....

basvodde: Nevermind above comment, str isn't the substr but the buffer_ It is early morning (5:10am)

arstrube: :-)

PR #520, opened 2015-03-16
Testregistry

arstrube:
Increasing statement coverage to 100%

### DIFF
--- a/tests/TestRegistryTest.cpp
+++ b/tests/TestRegistryTest.cpp
@@ -238,10 +238,10 @@ TEST(TestRegistry, findTestWithNameDoesntExist)
 
 TEST(TestRegistry, findTestWithName)
 {
-    test1->setTestName("SomeOtherTest");
-    test2->setTestName("NameOfATestThatDoesExist");
-    myRegistry->addTest(test2);
+    test1->setTestName("NameOfATestThatDoesExist");
+    test2->setTestName("SomeOtherTest");
     myRegistry->addTest(test1);
+    myRegistry->addTest(test2);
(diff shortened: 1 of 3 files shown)


tests/TestRegistryTest.cpp
```diff
@@ -300,3 +304,22 @@ TEST(TestRegistry, CurrentRepetitionIsCorrectTwo)
...
+
+TEST(TestRegistry, ResetPluginsWorks)
+{
+    MyTestPluginDummy plugin1("Plugin-1");
+    MyTestPluginDummy plugin2("Plugin-2");
+    myRegistry->installPlugin(&plugin1);
+    myRegistry->installPlugin(&plugin2);
+    LONGS_EQUAL(2, myRegistry->countPlugins());
+    myRegistry->resetPlugins();
+    LONGS_EQUAL(0, myRegistry->countPlugins());
```
basvodde: Comment on test style. Usually it is better to make this 2 tests and follow the AAA test pattern :)

arstrube: I totally agree with you. Yet there are situations (such as this one) where I want to _assert_ that my Arrange is the way I think it is.

Actually, I started out installing the NullPlugin twice. Guess what? The first assertion failed, because that is used as "no plugin"... The test case would have passed, but without proving the point it was supposed to prove.

Would you consider asserting the validity of your arrangements before the Action always "wrong"? I'm not opposed to splitting this in two tests, if you prefer. If set up identically, the first test could prove the validity of my Assertion (two plugins are indeed installed)

basvodde: I've merged it already, but usually I'd prefer two tests, it makes the purpose a lot clearer.

PR #525, opened 2015-03-17
Regenerated and test solution (VS2005)

arstrube:
Project files and solution for VS2005.

I imported the new VC98 workspace into VS2005.

I also imported this solution into VS2008, which compiled and ran successfully.The same is true of VS2010. 

In the interest of less maintenance, I was wondering whether we should maybe drop the xx_VS2008 (and possibly _VS201x) files and provide xx_VS2005 only instead. 

What does everybody think about that? The only snag is that we would need a relatively ancient copy of VS around for maintenance. I am willing to do that (and relatively unwilling to set up a VM running VS2008/10 on my Mac). 

Looking at the diff for the converted vcxproj's, it seems that it contains quite a bit of redundant stuff, but it works.

### DIFF
--- a/CppUTest.sln
+++ b/CppUTest.sln
@@ -1,0 +1,29 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 9.00
+# Visual C++ Express 2005
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "AllTests", "tests\AllTests.vcproj", "{E66A12BB-1E17-4CFE-A358-9E0FA85E0F15}"
+	ProjectSection(ProjectDependencies) = postProject
+		{EC28C821-4AB8-458F-A821-C6E65607B781} = {EC28C821-4AB8-458F-A821-C6E65607B781}
+	EndProjectSection
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CppUTest", "CppUTest.vcproj", "{EC28C821-4AB8-458F-A821-C6E65607B781}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Win32 = Debug|Win32
+		Release|Win32 = Release|Win32
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{E66A12BB-1E17-4CFE-A358-9E0FA85E0F15}.Debug|Win32.ActiveCfg = Debug|Win32
+		{E66A12BB-1E17-4CFE-A358-9E0FA85E0F15}.Debug|Win32.Build.0 = Debug|Win32
+		{E66A12BB-1E17-4CFE-A358-9E0FA85E0F15}.Release|Win32.ActiveCfg = Release|Win32
+		{E66A12BB-1E17-4CFE-A358-9E0FA85E0F15}.Release|Win32.Build.0 = Release|Win32
+		{EC28C821-4AB8-458F-A821-C6E65607B781}.Debug|Win32.ActiveCfg = Debug|Win32
+		{EC28C821-4AB8-458F-A821-C6E65607B781}.Debug|Win32.Build.0 = Debug|Win32
+		{EC28C821-4AB8-458F-A821-C6E65607B781}.Release|Win32.ActiveCfg = Release|Win32
+		{EC28C821-4AB8-458F-A821-C6E65607B781}.Release|Win32.Build.0 = Release|Win32
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+EndGlobal
--- a/CppUTest.vcproj
+++ b/CppUTest.vcproj
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="Windows-1252"?>
 <VisualStudioProject
 	ProjectType="Visual C++"
-	Version="9.00"
+	Version="8,00"
 	Name="CppUTest"
-	ProjectGUID="{F468F539-27BD-468E-BE64-DDE641400B51}"
-	TargetFrameworkVersion="0"
+	ProjectGUID="{EC28C821-4AB8-458F-A821-C6E65607B781}"
 	>
 	<Platforms>
 		<Platform
@@ -26,7 +25,6 @@
 			>
 			<Tool
 				Name="VCPreBuildEventTool"
-				CommandLine=""
 			/>
 			<Tool
 				Name="VCCustomBuildTool"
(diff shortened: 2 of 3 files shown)


arstrube:
The diff of the converted VS2008 vcproj is quite clean; basically only version info is chanted when imported.

Andne:
I agree with the thought of less maintenance, and all of the different project files does feel a bit cluttered to me as well.  However, I wonder a bit if it would put people off to only appear to be supporting a 10 year old version of Visual Studio.  My other concern is that occasionally there are some oddities around specific versions of Visual C++ where functions have been deprecated or warnings added, so if the project files are only checked against an old compiler, these things may not be caught.  Granted, I'm not sure how often the project files are being checked lately anyways, but that may be helped if I can ever get AppVeyor running like I started to.  Which apparently AppVeyor doesn't support 2005, only 2008 and newer.

That said, in my fork I don't actually use the project files as they currently are in the directory tree, I wanted to match the library file structure that CMake and autotools generate, so I redid the project files to split CppUTest from CppUTestExt.  Makes my merges a bit messier and I use cherry-pick a lot to create pull requests, but seems to be functional otherwise.

arstrube:
Yes, I know.... I was put off myself initially...

Interestingly, the macro for "safe" functions was needed / useful in VS2005 also. I found no issue with VS2010 at all, using the converted VS2005 files. 

Talk about VS98 - it is 17 years old :D

As for Cmake, I was disappointed to find that all it does is create Visual Studio projects, not makefiles. So I still need the IDE to convert it to makefiles, which in my opinion is slightly pointless...

On the other hand, Cmake will currently create any version of VS projects....

Pros and Cons of Cmake:
+ No clutter of VS projects
+ Easiest to maintain? (maybe the opposite -- see below)
– Force VS users to take an extra step building
– Force VS users to install an extra piece of software
– The Cmake version produces pretty odd (mostly useless) "standard" build targets
– I ran cmake for VS2008 and ended up with a lot of build errors
? How does one tell cmake to define a variable for VS6 but not for any of the others?
   (STDC_WANT_SECURE_LIB)

All in all, as a VS user I guess I'd rather import an "ancient" solution and see it working out of the box. I'm not sure though if this approach works with 2013, or whether it will "always" work in the future.

arstrube:
As for Appveyor – no need to "support" VS2005, but, could the conversion of the old solution be made an integral step of the build process? It seems a bit bothersome to supply two versions of .sln / .vproj which are *almost* identical.

arstrube:
As for the library structure, I have often wondered why the VS projects are so different. But then, they have always been that way...

arstrube:
@Andne - I added STDC_WANT_SECURE_LIB to the defines, which allowed me to build it. I changed the call to ctest so it uses -V rather than --force-new-ctest-process in order to get more reasonable output.

If we wanted to use this, some more work would have to go into it, in my opinion:
+ add STDC_WANT_SECURE_LIB but not to VC6
+ get ctest to run so at least the dots for the tests are shown
+ find some way to use the CppUTest -v option, when desired.

basvodde:
I guess VS users need to use the project to build CppUTest or is there some kind of package manager? If there is not (which I suspect), then I guess it will be better to maintain all these files :(

Andne:
Oddly, I don't think I've added STDC_WANT_SECURE_LIB to my projects since I pulled those files, yet they still build fine, hm... I need to investigate that, maybe I should double-check my warnings settings...

I'd have to see how passing an old project file to a slightly newer version of vcbuild works out.  Off the top of my head, I'd assume that passing it to VS2008 shouldn't be too bad, I'd worry a bit more about passing it to VS2010 or newer since the project file gets rewritten completely.  I think I saw something about that once, but don't remember any more.

I thought that one of the CMake outputs was nmake files, which should be able to call VC6 I would think.

There's a helper file for CMake I remember seeing once - FindCXX11.cmake - that appears to run an executable to detect certain things about the compiler.  Could something like that be used to determine if a given version of Visual C++ was being used?

My biggest problem with CMake has been with the set of projects it generates, it's just always felt like it generates a bunch of extra projects that served no purpose and made it harder to use.  Also, I know how to do a fair amount in msbuild and it seems like the way CMake works means that you can't take advantage of any of those capabilities and even if you do change the project files in Visual Studio, those changes just get lost anyways, so why even bother with having the project files in the first place?

I think we may be going a bit off topic here, maybe this discussion should move to a separate issue?

Andne:
@basvodde Newer versions of visual studio (2010+ I think, I know 2012+) have support for NuGet, which while it isn't perfect when it comes to working with native projects, is capable of being made to work at least.  There are some issues with Visual C++ and native code around release/debug runtimes, debug information format and one or two other options, but those can usually be handled through one mechanism or another.

basvodde:
Is there a CppuTest nuget package available?

Andne:
I was working on one, but was trying to use a feature of NuGet 3.0 that isn't actually ready yet, so not at the moment, no.  If I quite messing with the prerelease version identifier / continuous delivery idea, it wouldn't take much to create one.

Here's where I have my builds running, the nuget build is still broken, but it does build libraries for every version of Visual C++ from 2008-2013.  https://home.andne.net/TeamCity  Guest logins should work properly.

basvodde:
Nice, does it build from github main cpputest repository?

Andne:
No, it builds from the master branch of my fork.  The code files are a match to cpputest/cpputest, but I modified the project files so that I didn't have to figure out the build for each version of Visual C++, instead it has one set of project files that can build against any version (relies on some functionality that only exists with .vcxproj files, 2010 has the files with it to still use 2008's compiler if it is installed)

basvodde:
Ok, too bad :(

arstrube:
@Andne - the supplied projects don't have warnings as errors, therefore they should compile (though grudgingly ;-). cmake sets warnings as errors, therefore, the resulting project (CppUTest) doesn't compile with the current head unless you define STDC_WANT_SECURE_LIB .

> Oddly, I don't think I've added STDC_WANT_SECURE_LIB to my projects since I pulled those files, yet they still build fine, hm... I need to investigate that, maybe I should double-check my warnings settings...

arstrube:
@Andne - would you mind pulling this branch of my fork and see what happens when you load the sln into newer versions of VS, like 2010 and later? As far as I could see, 2010 worked fine except for some redundancy (it put a lot of entries with each source file that serve no purpose, as far as I can see.

My suggestion would be to make the 'normal' SLN VS2005 (not label it as such). Users wold notice they have to convert it, but it does so with no hitch up to VS2008. We can keep the _201x.sln in addition. So we'd have only two sets of solution/project files.

As for VC98, it would be nice to get rid of the .mak / .dep files, because they can always be regenerated from the project. But then, it's no big deal maintaining them, either.

Andne:
I pulled the project into VS2012, and it seems to mostly work.  There are some warnings, mostly about no-longer-supported tools (which aren't used anyways), and one about TargetPath not matching the librarian tool's output path.  It doesn't cause an issue that I can see, but it shows up during every build.

Looks like a few code files are missing from AllTests, I only had 647 tests with 643 of them running.

I looked at the generated project file, it seems pretty clean in the file list, it's just the compiler and linker settings are specifying a lot of intermediate and output files instead of using the default values that are identical.  I think we may be able to sort out some of those settings and possibly the TargetPath warning if the project settings regarding output were a little more structured in the first place.  Instead of relying on the copy build step, why not output directly to the lib folder to start with?  I'm not sure where CMake outputs to, but does it use Debug and Release the same as visual studio by default, or is there another folder we can mimic the behavior of instead?

arstrube:
Yeah - never thought of it but yes, 647 tests is a lot less than 714 for my standard automake gcc build. I did my best to get hold of all test files. Some will not compile because memory leak warnings are disabled.

arstrube:
In VC6, the best I can get is 687 test. So I must have missed last time when I brute-force added them....

arstrube:
I need to remove the disabling of memory leak protection in the VS2005 project. Then it will be 722 tests.

Also, I need to fix the invalid allocation size, because it needs to be a little less for MSC or it will throw rather than letting CppUTest handle it...

PR #532, opened 2015-03-18
100% statement coverage for MemoryLeakWarningPlugin.cpp

arstrube:


tests/MemoryLeakWarningTest.cpp
```diff
@@ -54,6 +54,36 @@ static MemoryLeakWarningPlugin* memPlugin;
...
+
+TEST(MemoryLeakWarningLocalDetectorTest, localDetectorIsTheOneSpecifiedInConstructor)
+{
+    MemoryLeakDetector localDetector(&dummy);
+    MemoryLeakDetector globalDetector(&dummy);
+    MemoryLeakWarningPlugin memoryLeakWarningPlugin("TestMemoryLeakWarningPlugin", &localDetector);
+    POINTERS_EQUAL(&localDetector, memoryLeakWarningPlugin.getMemoryLeakDetector());
+}
+
+IGNORE_TEST(MemoryLeakWarningLocalDetectorTest, localDetectorIsGlobalDetector)
```
basvodde: Ignore ?

arstrube: Yeah.

Not sure why setting the global detector to a local variable temporarily results in several "memory leaks" and a crash later on:
 1. A local variable is not allocated on the heap at all.
 2. Original detector and reporter are saved and restored
 3. The MemoryLeakDetectorPlugin instance goes out of scope after the test

 So, what is wrong in this picture?

arstrube: Help would be appreciated. This is a 'real' test case... Of course, I could cheat and not set the global detector at all and compare against getGlobalMemoryLeakDetector() ;-)

arstrube:
Latest commit and Travis build have become incongruent.

PR #535, opened 2015-03-18
100% statement coverage for MemoryLeakWarningPlugin.cpp

arstrube:


arstrube:
The failing test actually indicates a real problem. Clang -O2 will optimize code in such a way that the operator new overloads do not get called, if they are not the debug versions (with additional __FILE__ and __LINE__ arguments).

GCC on the other hand works just fine.

Taking out the optimization (-O0) will succeed on Clang as well.

So, how do we resolve this one?

Andne:
We learned a while back that optimizations cause some of the tests to fail (or even crash) under Visual C++, so maybe we just need to say that it's a really bad idea to build with optimizations for unit testing?  I think they can cause some weird behavior when checking coverage too, where the results don't always make sense.

arstrube:
I tend to agree with that. However, if we didn't have the tests with optimization, we (or certainly I) wouldn't have found out about the problem.

Ideally, I would like to handle it like a "known failure" somehow. That is
+ the test remains unchanged
+ the test fails on Travis when compiled with clang and -O2
+ the failure is harmless to the success of the build

Realistically, conditional compilation and IGNORE_TEST_ seems to be the only way to handle this.

The Visual C++ problem was harder in that it wouldn't even link.

@basvodde can we please get rid of the -O2 in the coverage run in cmake / automake? This is really nonsensical. It produces weird and rather useless results. Myself, I use:
```bash
../configure --enable-coverage CFLAGS="-g -O0" CXXFLAGS="-g -O0"
```
the -g being optional.

arstrube:
I tried 
`#ifdef __clang__`
`__attribute__((optnone))`
`#endif`
but, unfortunately, it doesn't help. I would prefer to deal with the problem at its root, rather than use IGNORE_TEST.

arstrube:
I looked at the bug that was submitted to LLVM (status: WONTFIX). They reference a document where they claim this behavior is specified. However, the same document also says:

"A C++ program may provide the definition for any of eight dynamic memory allocation function signatures declared in header <new> (3.7.4, 18.6):
 • operator new(std::size_t)
 • operator new(std::size_t, const std::nothrow_t&)
 • operator new[](std::size_t)
 • operator new[](std::size_t, const std::nothrow_t&)
 • operator delete(void*)
 • operator delete(void*, const std::nothrow_t&)
 • operator delete[](void*)
 • operator delete[](void*, const std::nothrow_t&)"

What is the point of these, if they will be silently optimized away, introducing subtle bugs? To me, this kind of optimization should have to be specified explicitly

arstrube:
At this point, I see two possible solution

tests/MemoryLeakWarningTest.cpp
```diff
@@ -385,8 +440,49 @@ TEST(MemoryLeakWarningWarningThreadSafe, turnOnThreadSafeNewDeleteOverloads)
...
 }
 
+#ifdef __clang__
+
+IGNORE_TEST(MemoryLeakWarningThreadSafe, turnOnThreadSafeNewDeleteOverloads)
+{
+    /*  Clang misbehaves with -O2 - it will not overload operator new or
+     *  operator new[] no matter what. Therefore, this test is must be ignored.
+     */
+}
```
basvodde: I actually think that I once made a bug report to clang about this. Don't think they ever fixed it, neither remember seeing it closed.

basvodde:
I'll merge this as is now. I'm of 2 minds related to removing the -O2. On one hand, you don't need to optimize the tests. On the other hand, it will expose differences in compiler behavior (such as these).

tests/MemoryLeakWarningTest.cpp
arstrube: You totally did. And the status totally is "Wontfix" :-(

arstrube:
Oki. And you are right. We did remove (I forgot) the -O2 for coverage, which is better. For the rest, with hindsight I must agree - it is better if these kinds of differences and bugs are exposed, even though it can be pretty annoying at times ;-)

arstrube:
By the way, in one of their answers they claim that using the option -ffreestanding will prevent this behavior. I could verify this with a small non-CppUTest test program. However, when I tried to apply this to the Travis build, it messed up the Gtest-17 build in ways that are beyond me. It also forces you to declare main as extern "C"....

Practically speaking, it is easier to tell our users to always use the ...NewMacros.h, and they'll be on the safe side.

basvodde:
Yup, lets do that for now and we'll see when this hits us again :)

Andne:
I agree, it's good to build with the ...Macros.h files when you can, but that also requires a bit of extra work.  For the projects at work that use this, I've actually had to set up 2 builds on our CI server.
- First I run my 'Release' build, which doesn't include the ...Macros.h files, since it will be linked with other projects later on, I need to have references to new, delete, malloc, etc... be to the C/C++ runtimes.
- Second is my 'Metrics' build, this one enables code coverage and force includes ...MallocMacros.h on all files and ...NewMacros.h on C++ files.

I maybe need to do a bit more work to ensure that traceability between the builds is better, but I've found that when working with native code, these are the hoops you have to jump through that you maybe wouldn't if you were working in a newer language.  I guess I wrote this mostly for awareness that while it is possible to use the ...Macros.h files most of the time, at some point there probably needs to be a build without them.

Note: CppUTest and mock/stub libraries I find to be an exception to this, given their usage (you always have the unit test framework around when using them) there isn't much reason to not always force the ...Macros.h files in those builds.

PR #536, opened 2015-03-20
Real fix for Clang problem -- experimental

arstrube:
NOTES:

+ This is still a bit of a hack; should probably be done in configure.ac rather than in Makefile.am

+ Maybe we should change the build generally so that it is mostly done with -O0 and use -O2 ony in specific runs to check whether it works

+ Doing so would limit the "special treatment" for Clang to those runs only

+ Wondering how we should treat -g? I would like to see it in the 'normal' build, but do we need it when the builds just run on Travis?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ set(CppUTest_version_minor 7dev)
 # 2.6.3 is needed for ctest support
 cmake_minimum_required(VERSION 2.6.3)
 
+set(CPPUTEST_CXX_FLAGS "${CPPUTEST_CXX_FLAGS} -ffreestanding")
+
 option(STD_C "Use the standard C library" ON)
 option(STD_CPP "Use the standard C++ library" ON)
 option(CPPUTEST_FLAGS "Use the CFLAGS/CXXFLAGS/LDFLAGS set by CppUTest" ON)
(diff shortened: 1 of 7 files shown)


arstrube:
Hmm. The diff for MemoryLeakWarningTest.cpp looks weird on Github - nothing like my local diff, which just removes the #ifdef __clang__ ... #else ... #endif bits.

arstrube:
Ignore that. Now it looks okay.

arstrube:
I got a mysterious error in the gtest build that I can't relate to the addition of -ffreestanding at all:
```bash
In file included from src/gtest-all.cc:44:
./src/gtest-filepath.cc:110:12: error: use of undeclared identifier
      '_POSIX_PATH_MAX'; did you mean '_PC_PATH_MAX'?
  char cwd[GTEST_PATH_MAX_ + 1] = { '\0' };
           ^~~~~~~~~~~~~~~
           _PC_PATH_MAX
./src/gtest-filepath.cc:58:26: note: expanded from macro 'GTEST_PATH_MAX_'
# define GTEST_PATH_MAX_ _POSIX_PATH_MAX
                         ^
/usr/include/x86_64-linux-gnu/bits/confname.h:36:5: note: '_PC_PATH_MAX'
      declared here
    _PC_PATH_MAX,
```
Any idea, anyone?

arstrube:
I don't think we want to go down this road :-(

PR #543, opened 2015-03-24
Surprisingly, this too is not (and cannot be) used...

arstrube:


basvodde:
This isn't true :)

arstrube:
Please enlighten me, how it is used? 

I can see from the Travis result that it is *somehow* used, but it is private, and it is not like there is a place where it is passed instead of output_ or something. So what's going on?

basvodde:
It is new-ed in the code you deleted :)

arstrube:
I know. But nobody can use it outside of CommandLineTestRunner, and CommandLineTestRunner doesn't reference it anywhere? If it did, the code shouldn't have compiled?

src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -112,12 +111,6 @@ int CommandLineTestRunner::runAllTests()
 bool CommandLineTestRunner::parseArguments(TestPlugin* plugin)
 {
     if (arguments_->parse(plugin)) {
-        if (arguments_->isJUnitOutput()) {
-            output_ = jUnitOutput_ = new JUnitTestOutput;
```
basvodde: Look, it makes output and jUnitOutput the same. Then it will mostly use output_

arstrube:
Ooops. Sorry. Blind spot :-( 

But - this line:

```c++
output_ = jUnitOutput_ = new JUnitTestOutput;
```
could probably be improved?

basvodde:
How improved, I always felt it worked pretty well. We just create the type of output, but we keep the pointer seperately to deal with the memory management issues

arstrube:
Hhhm. It certainly misled me. But of course, it will be intuitively clear to more brilliant minds ;-)

I'd say it's a bit of a hack to accomodate the package name for JUnit (which was added later, if memory serves me).

Anyway since the code needs to be there, it lacks a test :D.

PR #544, opened 2015-03-24
Add test case for generating JUnit style output

arstrube:


tests/CommandLineTestRunnerTest.cpp
```diff
@@ -94,3 +95,25 @@ TEST(CommandLineTestRunner, NoPluginsAreInstalledAtTheEndOfARunWhenTheArgumentsA
...
+struct TestOutputCheckingCommandLineTestRunner : public CommandLineTestRunner
+{
+    TestOutputCheckingCommandLineTestRunner(int ac, const char** av, TestOutput* output, TestRegistry* registry) :
+        CommandLineTestRunner::CommandLineTestRunner(ac, av, output, registry)
+    {
+    }
+
+    bool hasJUnitTestOutput(void)
+    {
+        return NULL != dynamic_cast<JUnitTestOutput*>(output_);
```
basvodde: Can we not use a dynamic_cast here. You should be able to check the CommandLineArguments class.

arstrube: Could you explain how? My point is to assert that output_ has been instantiated to a JUnitTestOuput rather than StringBufferTestOutput. How could the CommandLineArguments class help with that?

basvodde: It has an isJUnitOutput option :)

We don't use dynamic_casts in CppUTest due to backwards C++ compatibility.

arstrube: Okay. I'll think about alternatives. The problem is, that CommandlineArguments would happily assert it's junit (because the commandline param was such) but I could remove the code I'm testing here (as I mistakenly did before) and the test would then happily pass...

One approach would be asserting that print() doesn't print anything.

It would be nice though if output_ itself could be queried for it's junit-ishness...

basvodde: You can just make the junitoutput variable protected and compare whether output and junitoutput have the same value.

arstrube: Great idea. That will do :-)

PR #551, opened 2015-03-27
VS201x should define the STDC_WANT_SECURE_LIB macro

arstrube:
This will prevent the warning about using unsafe functions.


arstrube:
This one just stopped compiling in mid-air somewhere after the gtest run. How can I get this to be run again?

PR #553, opened 2015-03-28
Fixed base class constructor call

Andne:
Last pull request caused errors in my VC build.  Looks like it doesn't like treating the constructor as a static function as part of an initializer list.  I'm a bit surprised that other compilers didn't mind this, I don't think I've ever seen it before.


arstrube:
My bad.... glad you caught this one. MS to the rescue!

PR #557, opened 2015-03-30
mac - 100% Coverage for MockActualCall & Co.

arstrube:


tests/CppUTestExt/MockActualCallTest.cpp
```diff
@@ -133,3 +142,56 @@ TEST(MockCheckedActualCall, MockIgnoredActualCallWorksAsItShould)
     CHECK_FALSE(actual.hasReturnValue());
     CHECK(actual.returnValue().equals(MockNamedValue("")));
 }
+
+TEST(MockCheckedActualCall, remainderOfMockActualCallTraceWorksAsItShould)
+{
```
arstrube: Break this one up in smaller tests? It started out like MockIgnoredCall, except that there is a lot more interesting stuff to check here.

src/CppUTestExt/MockActualCall.cpp
```diff
@@ -643,6 +639,7 @@ unsigned int MockActualCallTrace::returnUnsignedIntValueOrDefault(unsigned int)
 
 MockActualCall& MockActualCallTrace::onObject(void* objectPtr)
 {
+    traceBuffer_ += " onObject:";
```
arstrube: The address of the object used to be stuck onto the CallOrder like so:
`withCallOrder: 10x4536123`

One might use newlines for every parameter (with an indent) or stick on everything seperated by spaces like now. Also, a space might be after the colon, but if so, I feel it should be done consistently.

basvodde:
Seems ok for now :)

PR #558, opened 2015-03-31
mec - 100% Coverage for MockExpectedCall & Friends

arstrube:


Andne:
Merged these changes into my build tree and MockExpectedCallComposite.isOnObject is failing.  Here's the error message:
```
AssertionFailedError: MockExpectedCallTest.cpp:533:
expected [(object address: 0x49ee00)::name -] no parameters]
but was  [(object address: 0049EE00)::name -] no parameters]
difference starts at position 18 at: [address: 0049EE00)::]
```

Apparently whatever is doing formatting under Visual C++ likes uppercase for hex digits.

arstrube:
I will change it to use the NOCASE macro - makes sense with hex numbers anyway. 
Or should we rather change the code to ensure lower / upper case always?
vsnprintf() doesn't seem to follow much of a standard ;-).

arstrube:
interesting that this never popped up with the SimpleString tests...

PR #562, opened 2015-04-01
Removing two lines from MemoryLeakDetector

basvodde:
I think these 2 lines are *probably* not needed anymore. They were there due to the earlier unusual creation of the MemoryLeakDetector to deal with early static initialization and late destruction.


arstrube:
The build didn't break ;-)
Also, I tend to agree. Who would write e.g.:
```C++
MemoryLeakDetectorTable memoryTable_;
memoryTable_ = MemoryLeakDetectorTable();
```
:P

PR #573, opened 2015-04-02
made commandline argument -oJunit compatible with -v 

WilcoBT:
By referring everywhere to the parent functions and adding support for the flushing is it possible to combine -oJunit with -v.

### DIFF
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -282,6 +282,7 @@ void JUnitTestOutput::flush()
 
 void JUnitTestOutput::print(const TestFailure& failure)
 {
+    TestOutput::print(failure);
     if (impl_->results_.tail_->failure_ == 0) {
         impl_->results_.failureCount_++;
         impl_->results_.tail_->failure_ = new TestFailure(failure);
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -98,24 +98,31 @@ void JUnitTestOutput::resetTestGroupResult()
 
 void JUnitTestOutput::printTestsStarted()
 {
+    TestOutput::printTestsStarted();
 }
 
-void JUnitTestOutput::printCurrentGroupStarted(const UtestShell& /*test*/)
+void JUnitTestOutput::printCurrentGroupStarted(const UtestShell& test)
 {
+    TestOutput::printCurrentGroupStarted(test);
 }
 
 void JUnitTestOutput::printCurrentTestEnded(const TestResult& result)
 {
+    TestOutput::printCurrentTestEnded(result);
+
     impl_->results_.tail_->execTime_
             = result.getCurrentTestTotalExecutionTime();
 }
 
-void JUnitTestOutput::printTestsEnded(const TestResult& /*result*/)
+void JUnitTestOutput::printTestsEnded(const TestResult& result)
 {
+    TestOutput::printTestsEnded(result);
 }
 
 void JUnitTestOutput::printCurrentGroupEnded(const TestResult& result)
 {
+    TestOutput::printCurrentGroupEnded(result);
+
     impl_->results_.groupExecTime_ = result.getCurrentGroupTotalExecutionTime();
     writeTestGroupToFile();
     resetTestGroupResult();
@@ -123,6 +130,8 @@ void JUnitTestOutput::printCurrentGroupEnded(const TestResult& result)
 
 void JUnitTestOutput::printCurrentTestStarted(const UtestShell& test)
 {
+    TestOutput::printCurrentTestStarted(test);
+
     impl_->results_.testCount_++;
     impl_->results_.group_ = test.getGroup();
     impl_->results_.startTime_ = GetPlatformSpecificTimeInMillis();
@@ -241,20 +250,33 @@ void JUnitTestOutput::writeTestGroupToFile()
 
 // LCOV_EXCL_START
 
-void JUnitTestOutput::printBuffer(const char*)
+void JUnitTestOutput::printBuffer(const char* buffer)
 {
+    TestOutput::printBuffer(buffer);
+    if( verbose_ ) {
+        while( *buffer ) {
+            PlatformSpecificPutchar( *buffer );
+            buffer++;
+        }
+        flush();
+    }
 }
 
-void JUnitTestOutput::print(const char*)
+void JUnitTestOutput::print(const char* values)
 {
+    TestOutput::print(values);
 }
 
-void JUnitTestOutput::print(long)
+void JUnitTestOutput::print(long value)
 {
+    TestOutput::print(value);
 }
 
 void JUnitTestOutput::flush()
 {
+    if( verbose_ ) {
+        PlatformSpecificFlush();
+    }
 }
 
 // LCOV_EXCL_STOP
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -252,7 +252,6 @@ void JUnitTestOutput::writeTestGroupToFile()
 
 void JUnitTestOutput::printBuffer(const char* buffer)
 {
-    TestOutput::printBuffer(buffer);
     if( verbose_ ) {
         while( *buffer ) {
             PlatformSpecificPutchar( *buffer );
--- a/include/CppUTest/JUnitTestOutput.h
+++ b/include/CppUTest/JUnitTestOutput.h
@@ -76,4 +76,42 @@ protected:
 
 };
 
+///////////////////////////////////////////////////////////////////////////////
+//
+//  StringBufferJUnitTestOutput.h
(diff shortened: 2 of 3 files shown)


basvodde:
Thanks for the pull request. I noticed that there aren't any tests with the pull requests? I'm afraid I won't accept pull requests without tests (and I'd recommend to start with the tests).

@arstrube Do you think its a good idea to have -v and -ojunit still output to console? I guess it can't harm, right? Might be useful even!

arstrube:
From a users point of view it is definitely desirable - we are currently running the tests twice to obtain both. If the changes in code are straightforward (haven't looked at it yet) I would be in favor of it. In our case, we are actually thinking about dropping junit output altogether, because we don't seemt to use the xml for anything. Both our reporting script and the CppUTest Eclipse plugin rely on console output.

As a matter of interest, what was the original reason to disable all output when writing junit output?

As for tests, I am not too surprised the Travis build failed because, if I remember correctly, there is at least one test to make sure there is nothing written to console when junit output is active :-).

WilcoBT:
Alright, these tests seem to succeed now.

basvodde:
Wow, lots of code. I'll need to think about it a bit though. I'm wondering if we are not better off making the console output a decorator of the JUnitOutput than to intermingle it. That way the code and tests would be a lot easier.

@WilcoBT Do you know how it would look like as decorator?

WilcoBT:
I should but I don't think I ever have played with one. Now grabbing the gang of 4 book.

basvodde:
Ok, what might be best is to create a tee TestOutput.

So, you derive a class from TestOutput and it passes 2 TestOutput in the constructor. Each call, you then forward to both the TestOutput.

Then when the -v and -ojunit is passed, you will create the TeeTestOuput and pass a JUnitOutput and a ConsoleOutput. That way, you just need to test one seperate class and your code should be a lot easier...

Andne:
I was thinking about bit about that this morning, though my thought wasn't so much an outputter that knows two outputs as one that could handle an arbitrary number of outputs.  -v could be made an alias for -overbose or something and the implicit default could be -oconsole, and then use a factory to get the instance pased on the -o option and keep adding outputters as additional -o options are found.  I think that it would be more flexible in the future if we added any other outputters than what we have today.

So to get 'normal' console output with junit output the command line would look like this:
`CppUTestTests.exe -oconsole -ojunit`

Verbose output with junit output would look like this:
`CppUTestTests.exe -overbose -ojunit`
or this:
`CppUTestTests.exe -v -ojunit`

And 'normal' console output with verbose output (why?) would look like this:
`CppUTestTests.exe -oconsole -overbose`

Now if there was only a way to dynamically register outputters so they could be discovered

arstrube:
That sounds very useful.

arstrube:
Did you know there already is -oeclipse, which is functional, and code for vistualStudio (sic!) which has been tested but can't be activated due to the lack of a -ovistualStudio option ;-)

arstrube:
@basvodde this is what happens when you test features that are actually inaccessible :P

It seems we haven't needed anything special for VS for the error recognition to work? The eclipse working environment is actually the default all platforms. @Andne except for -ojunit the same TestOutput class was used to cover all, just with slightly different print methods.

arstrube:
@Andne I think we are looking at three distinct types of options here:
+ Verbosity (-v), currently only affects console output
+ Working environment (-onormal, -oeclipse, vistualStudio) which affect slightly different printing of test failures
+ distinct outputters (-ojunit), currently junit vs. console
maybe use -w { eclipse | vistualStudio | ... }, -o { console | junit | } and -v (as before). Then -o could occur more than once and invoke more than one outputter, if desired.

WilcoBT:
Which brings us back to the decorator pattern.
Looking at it, I see most of the current implementation of TestOutput to go to consoleTestOutput, which can be initialised with parameters 'color' and 'verbose' both default false.
The StringbufferTestOutput should then inherit from this one.
TestOutput itself will just remain as pure virtual.
The biggest question I haven't answered for me is if we actually need a "ConcreteComponent". I don't really see it as the decorator can just handle the empty component case.

There are then a couple of things that don't make sense such as multiple TestOutputs in the chain that want access to the console and there some kind of generator/factory would serve nicely to only create sane combinations. But maybe this can also be handled by making everything that needs the console a "ConcreteComponent" such that there can be only one as it terminates the chain.

arstrube:
@WilcoBT just wondering whether your immediate problem couldn't be solved easily with an appropriate main() function?

WilcoBT:
Unless you are back to suggesting running it twice I do not see what you meen @arstrube

arstrube:
Oh, I see. No, definitely not running twice.

What I mean is this:

1. Make a file called something like MainForEveryTest.cpp

2. In that file, define main() such that it does your wchar_t conversions before passing stuff to RUNALLTESTS().

3. In every VS project, remove whatever main.cpp you are currently using, and add MainForEveryTest.cpp.

4. As an added bonus, you are dealing with the wchar_t stuff at the earliest possible point ;-)

(In my experience, there is little need to have specialized main() functions for every test suite. If you are currently doing that, maybe there is a better way).

arstrube:
Oops. Sorry. Wrong thread. I meant your problem with wchar_t. Of course it won't work for the output.

WilcoBT:
:)

arstrube:
So, could your solve the wchar_t problem in that manner?

WilcoBT:
yes

src/CppUTest/JUnitTestOutput.cpp
```diff
@@ -241,26 +250,39 @@ void JUnitTestOutput::writeTestGroupToFile()
 
 // LCOV_EXCL_START
 
-void JUnitTestOutput::printBuffer(const char*)
+void JUnitTestOutput::printBuffer(const char* buffer)
 {
```
arstrube: Actually, I had to exclude this from test coverage because they currently serve no useful purpose. Nice to see these put to good use ;-)

src/CppUTest/JUnitTestOutput.cpp
```diff
@@ -241,26 +250,39 @@ void JUnitTestOutput::writeTestGroupToFile()
 
 // LCOV_EXCL_START
```
arstrube: Actually, I had to exclude this from test coverage because they currently serve no useful purpose. Nice to see these put to good use ;-)

In your version, the // LCOV... comments can (and should) be removed

WilcoBT: I'll try to keep it in mind as I overthrow it once more

arstrube:
@basvodde - not _that_ much code really :P -- most of it is tests.

arstrube:
If the decorator pattern will help us not have derived classes full of unused methods, I'm all for it. Otherwise, this solution is quite adequate, no? It actually makes more sense to me than the current implementation with its empty methods.

src/CppUTest/JUnitTestOutput.cpp
arstrube: :-)

src/CppUTest/JUnitTestOutput.cpp
arstrube: It would be nice not to have to reimplement printBuffer()...

WilcoBT:
It will as the Decorator definition will have code for each function and thus everything inheriting from it has no need anymore to implement more then what they extend.
It will hit a lot more code. I have at the moment prepared the commandlinetestrunner and refactored the outputtest. This already hit a lot of tests.
Now starting on the decorator so I can hook up the implementations.

WilcoBT:
ow, practical question, is "#pragma once" allowed?

arstrube:
Sounds good :-)

arstrube:
I don't think so. The one exception being UtestPlatform.cpp. But ultimately, that's for @basvodde  to say.

arstrube:
In my opinion though, it makes no sense to use `#pragma once` just _once_ ;-)

WilcoBT:
:) I was tempted to use it in the new files I am creating or just touching very seriously as I was running once again into the copy paste messyness if #ifndef

arstrube:
Don't I know... well consider it a tribute to supporting really old, really cranky compilers :D

BTW Adding files to CppUTest carries a penalty...

WilcoBT:
Please, tell me more. I think I'm at 6 new files now.

WilcoBT:
Ow, is this the mess with adding them to all the different build system scripts?

arstrube:
Yup. Exactly that. Without Makefile.am and the various cmake files, you won't get past Travis CI. And then several VS projects for sure. Less is more here :-)

WilcoBT:
As doing this with the decorator pattern involved a lot of reverting it became to messy so I started a new branch.
Have JunitTestOutput as a decorator now and the tests are still running! One slight ugly thing with incorrect commandline arguments test actually printing to the real commandline.
Next step, update the commandlinetestrunner to initialise a chain when needed and add some tests.

WilcoBT:
Unless we still want to do something with this first attempt I think this can be closed now the 2nd is offered.

PR #576, opened 2015-04-03
Add countCheck() to UtestShell rather

arstrube:


src/CppUTestExt/MockSupport.cpp
```diff
@@ -137,7 +137,7 @@ MockExpectedCall& MockSupport::expectOneCall(const SimpleString& functionName)
 {
     if (!enabled_) return MockIgnoredExpectedCall::instance();
 
-    countCheck();
```
arstrube: or have
```
void MockSupport::countCheck() 
{
    UtestShell::getCurrent()->countCheck();
}
```
perhaps, so this line would stay?

PR #579, opened 2015-04-03
Change the test group to match the new name

Andne:
Error introduced by PR 575 (I think), the test group inside the preprocessor #if block was changed, but the test group inside the #else block was not changed.


arstrube:
Thank you Andy for catching this one. My IDE greys out the #else's in such a way that it is hardly readable, so I missed it. I do plead guilty for not running every PR past VC :-(.

Andne:
That would be why I need to finish getting appveyor to work properly ;).  I did get one build to run, but it could definitely use some improvement yet.

PR #580, opened 2015-04-04
New test for crash on failure with mock_c()

arstrube:
This test fails because crashOnFailure_ is never passed to MockFailureReporterTestTerminatorForInCOnlyCode. It is only passed to the default test terminator, which isn't called here.

The Travis CI build is expected to fail at this point ;-)

### DIFF
--- a/include/CppUTestExt/MockSupport_c.h
+++ b/include/CppUTestExt/MockSupport_c.h
@@ -128,6 +128,7 @@ struct SMockSupport_c
     int (*expectedCallsLeft)(void);
 
     void (*clear)(void);
+    void (*crashOnFailure)(void);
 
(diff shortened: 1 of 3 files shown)


arstrube:
This is now working.

But -- it introduces subtle differences in the semantics of mock_c() vs. mock(): 
- mock().clear() doesn't clear everything that mock_c().clear() does
- mock().crashOnFailure() doesn't with mock_c() but mock_c().crashOnFailure() does (this is not new, however, it was never tested before)

This is a bit of a conondrum -- we could take out the crash on failure functionality from mock_c() instead. This wouldn't change anything, as it was hitherto inaccessible. However, that too would be a subtle semantic difference that users wouldn't expect, right?

arstrube:
I guess I lean towards taking it out.

basvodde:
Ok!

PR #581, opened 2015-04-04
Mockc nocrash

arstrube:
Take out this functionality because crashOnFailure is never passed as true

### DIFF
--- a/src/CppUTestExt/MockSupport_c.cpp
+++ b/src/CppUTestExt/MockSupport_c.cpp
@@ -35,9 +35,6 @@
 class MockFailureReporterTestTerminatorForInCOnlyCode : public TestTerminatorWithoutExceptions
 {
 public:
-    MockFailureReporterTestTerminatorForInCOnlyCode(bool)
-    {
-    }
 
     virtual void exitCurrentTest() const _override
     {
@@ -55,7 +52,7 @@ public:
     void failTest(const MockFailure& failure) _override
     {
         if (!getTestToFail()->hasFailed())
-            getTestToFail()->failWith(failure, MockFailureReporterTestTerminatorForInCOnlyCode(crashOnFailure_));
+            getTestToFail()->failWith(failure, MockFailureReporterTestTerminatorForInCOnlyCode());
     }
 };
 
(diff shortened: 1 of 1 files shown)


arstrube:
The problem with crashOnFailure in mock_c() is that
1. it is never set
2. if we did set it, it will stay set forever, because the test terminator is static

It seems that mock_c() works fine for everything that is stored inside the mock() class itself, but it doesn't work for values stored in aggregates that become static in mock_c().

basvodde:
This change doesn't make sense. This would disable the crash on failure in C code. Not understand why we would want that.

On your comments:
1) It is passed in the constructor and set in the base class, right?
2) Of course it stays forever, that is the point of crashOnFailure :) If it is true, it will crash.

arstrube:
The reason this makes sense is that this code cannot be executed under any circumstances :-)

1. Nope. Wrong. I can submit a test that proves it (actually, was part of PR #580 that I closed). It _might_ be set if I first make a call to mock_c() and only then call mock().crashOnFailure. I wouldn't call that proper behavior though. It is unlikely to happen in this way, especially if I use mock() in test cases and mock_c() in mocks written in C. It would be impossible to activate if using pure mock_c(), since crashOnFailure is not part of the mock_c interface.

2. Nope. Wrong again. For mock(), the setting stays only for the current TEST, because the test terminator isn't static. I see your point though; but if staying forever was a feature we actually wanted indeed, it would become very difficult to test, as it would be active in each and every subsequent TEST and has no matching operation to reset it.

Please take a look at the closed PR #580 to see the other option. Also, that PR would take considerably more changes to work properly, where with "properly" I mean that mock() calls and mock_c() calls are 100% exchchangeable under all circumstances.

basvodde:
When you use CrashOnFailure, you just want it to CrashOnFailure, so the scope of the test isn't really that relevant. It is a debugging technique and not something that would ever be permanently in a test. So, I do not find it very useful to make it work 100% exchangable, as long as you can let it crash on failure.

Related to #1, so you mean we ought to add a crashOnFailure method to the mock_c() ?

arstrube:
Yes, if it is to work at all, we would definitely need a crashOnFailure method in mock_c().

I recommend my addition to mock_c()::clear() also, if for no other good reson, the for keeping our sanity when testing CppUTest. In other words, close this PR and open #580 again ;-)

But, as I said, I wasn't entirely happy with that one.

arstrube:
P.S. If you can make it work without mock_c().crashOnFailure(), I'd be very interested to see how you did it.

arstrube:
Rejected

PR #582, opened 2015-04-06
100% Coverage for MockSupport_c - Part 1

arstrube:
because crashOnFailure_ is never passed to
MockFailureReporterTestTerminatorForInCOnlyCode.


arstrube:
Ooops. Something went wrong re-opening #580, so this is now its replacement.

arstrube:
#5448F86 - An unrelated test fails on TravicCI because the crash function remains set

arstrube:
Decided to split this PR into smaller ones. This one now only contains one additional test case and additions to another.

PR #584, opened 2015-04-07
100% Coverage for MockSupport

arstrube:


include/CppUTestExt/MockFailure.h
```diff
@@ -48,6 +48,7 @@ class MockFailureReporter
     virtual UtestShell* getTestToFail();
 
     virtual void crashOnFailure() {crashOnFailure_ = true; }
+    virtual void clear() {crashOnFailure_ = false; }
```
basvodde: I don't think I'd like crashOnFailure_ to be false in clear. This is something that I would even put in the main so that I can easily turn it on/off. It might be good to make it a command line parameter, but definitively not something you want to turn off every clear.

arstrube: Okay. Because when using it, we might not know which test case is concerned?

How about turning it off just before we crash? In a real run, we have crashed anyway. In the test, we could be sure of no side effects.

arstrube: That doesn't work. I will drop the second test case and not need to use TEST_ORDERED, since this behavior is not intended.

However, I am a bit worried about possible side-effects...

arstrube: It is very deeply encapsulated. If it wasn't, I think turning it off at the beginning of a test run would seem like a good idea.

tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -1720,24 +1750,39 @@ TEST(MockSupportTestWithFixture, mockExpectationShouldIncreaseNumberOfChecks)
...
+    cpputestHasCrashed = false;
     mock().crashOnFailure();
-    fixture.registry_->setRunTestsInSeperateProcess();
+    UtestShell::setCrashMethod(crashMethod);
     fixture.setTestFunction(crashOnFailureTestFunction_);
     fixture.runAllTests();
-    fixture.assertPrintContains("Failed in separate process - killed by signal 11");
+    CHECK(cpputestHasCrashed);
+    UtestShell::resetCrashMethod();
+    mock().clear();
```
basvodde: Why not just use mock().crashOnFailure(false) here? Then leave clear out of it.

arstrube:
It seems to be okay if I don't explicitly test that behavior. Let's see the results of the Travis CI build though.

tests/CppUTestExt/MockSupportTest.cpp
arstrube: So we change

`void crashOnFailure(void)` to `void crashOnFailure(bool)`, right?

BTW, for MockSupport, the TestTerminator is thrown away at some point, so the setting doesn't seem to survive too long. For MockSupport_c, on the other hand, it causes real trouble because the test terminator is static.

arstrube:
Do we have a standard way of representing `bool` on the mock_c() side?

PR #585, opened 2015-04-07
mnv - 100% Coverage for MockNamedValue

arstrube:


include/CppUTestExt/MockNamedValue.h
```diff
@@ -85,7 +85,6 @@ class MockNamedValue
     virtual void setObjectPointer(const SimpleString& type, const void* objectPtr);
     virtual void setSize(size_t size);
 
-    virtual void setComparator(MockNamedValueComparator* comparator);
```
basvodde: Don't think we want to remove this, do we?

arstrube: It is never used. Comparaters are always set via installComparator().

If I understand what this is for, I can write a test for it and not remove it ;-)

basvodde: It allows to change the comparitor for a particular NamedValue. When removing this, the whole member comparitor_ would also be possible to be removed.

I can't argue I've used the setting of the comparitors in the MockNamedValue, but it feels like something that shouldn't be removed ;)

arstrube:
Hmm. I'v put it back and added a test for it, but I couldn't think of any useful application for it, even after adding the test case.

I mean, there isn't any way I can set a comparator on a MockNamedValue and then pass that to the mocking framework in any useful way. The packagin of objects into MockNamedValue's happens all behind the scenes, no?

Neither am I going to use this in any production code.

I don't mind keeping it, but I'd like to see some useful application for it, at least theoretically. One such application would be a one-off comparison, without the need to install a comparator for it. Can that be done?

arstrube:
I tried removing `comparator_` but wound up with rather ugly zero-pointer-checking constructs in at least two places, so I think it would be better to keep it.

With the setter gone, it is clear that `comparator_` is only ever set from the "inside", using defaultRepository_. I find this much less confusing.

Should we ever actually need the setter, it can be put back very easily :P

basvodde:
Bah, I like the setter but have no argument for it, I'll merge it.

PR #586, opened 2015-04-07
Add a boolean argument to crashOnFailure()

arstrube:
So we can de-activate crashOnFailure after testing it


arstrube:
@basvodde good suggestion. I like it.

arstrube:
On second thoughts - this might break users' existing tests. Wouldn't adding something like `dontCrashOnFailure()` perhaps be less invasive?

basvodde:
Make it a default =true argument

arstrube:
Will do.

PR #589, opened 2015-04-08
Issue571 attempt2: made commandline argument -oJunit compatible with -v Via decorator pattern.

WilcoBT:
Clean branch for clarity after feedback on the pull request " made commandline argument -oJunit compatible with -v #573"

### DIFF
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -157,6 +157,7 @@ copy $(OutDir)CppUTest.lib lib\CppUTest.lib
     <ClCompile Include="src\CppUTest\TestHarness_c.cpp" />
     <ClCompile Include="src\CppUTest\TestMemoryAllocator.cpp" />
     <ClCompile Include="src\CppUTest\TestOutput.cpp" />
+    <ClCompile Include="src\CppUTest\TestOutputDecorator.cpp" />
     <ClCompile Include="src\CppUTest\TestPlugin.cpp" />
     <ClCompile Include="src\CppUTest\TestRegistry.cpp" />
     <ClCompile Include="src\CppUTest\TestResult.cpp" />
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -146,6 +146,7 @@ copy $(OutDir)CppUTest.lib lib\CppUTest.lib
     <ClCompile Include="src\CppUTestExt\OrderedTest.cpp" />
     <ClCompile Include="src\CppUTest\CommandLineArguments.cpp" />
     <ClCompile Include="src\CppUTest\CommandLineTestRunner.cpp" />
+    <ClCompile Include="src\CppUTest\ConsoleTestOutput.cpp" />
     <ClCompile Include="src\CppUTest\JUnitTestOutput.cpp" />
     <ClCompile Include="src\CppUTest\MemoryLeakDetector.cpp" />
     <ClCompile Include="src\CppUTest\MemoryLeakWarningPlugin.cpp" />
@@ -184,6 +185,7 @@ copy $(OutDir)CppUTest.lib lib\CppUTest.lib
     <ClInclude Include="include\CppUTestExt\OrderedTest.h" />
     <ClInclude Include="include\CppUTest\CommandLineArguments.h" />
     <ClInclude Include="include\CppUTest\CommandLineTestRunner.h" />
+    <ClInclude Include="include\CppUTest\ConsoleTestOutput.h" />
     <ClInclude Include="include\CppUTest\JUnitTestOutput.h" />
     <ClInclude Include="include\CppUTest\MemoryLeakDetector.h" />
     <ClInclude Include="include\CppUTest\MemoryLeakDetectorMallocMacros.h" />
@@ -193,12 +195,14 @@ copy $(OutDir)CppUTest.lib lib\CppUTest.lib
     <ClInclude Include="include\CppUTest\SimpleMutex.h" />
     <ClInclude Include="include\CppUTest\SimpleString.h" />
     <ClInclude Include="include\CppUTest\StandardCLibrary.h" />
+    <ClInclude Include="include\CppUTest\StringBufferTestOutput.h" />
     <ClInclude Include="include\CppUTest\TestFailure.h" />
     <ClInclude Include="include\CppUTest\TestFilter.h" />
     <ClInclude Include="include\CppUTest\TestHarness.h" />
     <ClInclude Include="include\CppUTest\TestHarness_c.h" />
     <ClInclude Include="include\CppUTest\TestMemoryAllocator.h" />
     <ClInclude Include="include\CppUTest\TestOutput.h" />
+    <ClInclude Include="include\CppUTest\TestOutputDecorator.h" />
     <ClInclude Include="include\CppUTest\TestPlugin.h" />
     <ClInclude Include="include\CppUTest\TestRegistry.h" />
     <ClInclude Include="include\CppUTest\TestResult.h" />
--- a/Makefile.am
+++ b/Makefile.am
@@ -47,7 +47,6 @@ lib_libCppUTest_a_SOURCES = \
 	src/CppUTest/TestOutput.cpp \
 	src/CppUTest/TestOutputDecorator.cpp \
 	src/CppUTest/ConsoleTestOutput.cpp \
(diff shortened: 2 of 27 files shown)


basvodde:
Wow, too much change at ones still :)

I browsed through it and it looks pretty good, but I'd prefer to split it up in smaller changed that we can look at. It is hard to see the larger change because of the size.

Is it true that the first step you are taking is to move functionality out of TestOutput to ConsoleOutput? If so, can we take this change as a first pull request?

WilcoBT:
Yes, I broke it up to get the TestOutput free of stuff the decorator line does not need. Besides the makefiles that I did as an afterthought that can indeed be done pretty stand alone.

arstrube:
Also, if you extract ConsoleOutput.h from TestOutput.h in a separate step, you don't have to initially change the half-dozen or so files that include it. 

Perhaps there is no need for a separate header at all? This is not Java, after all.... Just wondering.

arstrube:
I hope that was clear. I mean extracting into a different header file, as opposed to extracting the functionality to a new class in the same header.

WilcoBT:
No I'm not following you but maybe that is a style difference. I'm used to a class being in a file with the same name, this to be acutely aware of the class you are looking at, not that you accidentally got in the next when you scrolled. 
So yes, maybe it wouldn't ripple less to keep it in the same files. I don't believe in trying to avoid changes if you pay for that in clarity as I believe that is bad for the maintainability/learningcurve.

basvodde:
@WilcoBT Could you make the changes to extract functionality out of TestOuput to ConsoleOutput in a separate pull request? That way I can look at that change (and the effect on tests) separately from the Decorator.

Sorry for requesting the extra work. Its just that large pull requests make me nervous as it is hard to see all the changes :(

WilcoBT:
Yeah I can do that.

Do you happen to have an idea/solution for the error on autotools:
"if [ -s junit_run_output ]; then echo "JUnit run has output. Build failed!"; exit 1; fi
JUnit run has output. Build failed!"

arstrube:
Yup. There is a check in the build script that makes sure -ojunit produces no output :D

arstrube:
I guess you'll have to remove that as soon as you produce actual output with -ojunit active.

WilcoBT:
The issue is the CommandlineTestRunnerTests. To verify it creates the right chain of outputs I think I need it to create the real thing, which gives noise on the commandline that I dislike and I think that test fails on.
I have been looking at it but don't see a solution I like so stuck with this one.

arstrube:
@WilcoBT I guess CppUTest is trying to take a middle road. It does keep closely related classes in the same file at times. Never mind that, I found this irritating myself at times, but by delaying the extraction in a separate step you can lower the changes-per-pull-request ratio significantly and still end up where you want to go. Just an idea...

arstrube:
```bash
./CppUTestTests -ojunit > junit_run_output
```
It writes the console output into a file and then barks if it contains anything. Rmember, this used to be intentional. It should still be fine for the first pull request that Bas asked for, no?

Later, you will need to remove the `if ... ; fi` line from ../scripts/travis_ci_build.sh .

arstrube:
Better even, we might later change it to something that verifies it DOES contain output.

BTW, I am a bit surprised about this failure. Just specifying -ojunit should not always give console output, or? Shouldn't we be explicitly telling CppUTest to do both? And then the TravisCI script should still pass, because we didn't explicitly tell it to write to the console in addition to JUnit output?

WilcoBT:
Not sure, already seeing more work then I hoped with trying to get that 2nd commit free of the first. Fat collisions on the commandlinetestrunner. The problem is wanting to initialise after the parsing. This might just be related to that output story

Maybe I see it tomorrow, time to go home soon.

WilcoBT:
I can't put much time in it anymore though, can't endlessly fold it under the bigger project I'm working on here in the office.

arstrube:
On second thoughts, I think the failure on Travis CI is justified and you'd better _not_ follow my advice above... -ojunit should behave as always, shouldn't it.

Only if we specify -v -ojunit or perhaps -oconsole -ojunit we should have both. Or what do you think, @basvodde, @Andne ?

arstrube:
I understand about the time constraints...

WilcoBT:
I think the source of my problem comes from the fact that I see the color and verbose settings as very consoleoutput specific and thus have more place as specific to them then in the TestOutput interface.
From there the most logical thing for me is to put them in the constructor, but that makes the injecting from the tests impossible as it requires the parameters to be parsed etc.

I'm also wondering if the tests that give this mess are also maybe involving to much components for a true unit test. I'll think more about that tomorrow

arstrube:
Agreed about color and verbose settings... You may well be right about those tests, too.

WilcoBT:
Ah and when I walk to the train I realise just what these symptoms mean.
I think the solution should be to first parse the inputs. Then have some kind of factory create the testoutput chain, which can then be given to the commandlinetestrunner.
This way the factory can be tested on its own and the commandlinetestrunner can be stubbed.
I do remember trying some small step in that direction and already running into issues. It seems that pluginsystem has some side effects or something on the parsing. I'll have to see if it is possible to get that to work before creating a commandlinetestrunner.

basvodde:
Wow, you guys are amazing at commenting :) Lets try to keep things simple though and one step at the time. First, lets extract the Console specific content from the TestOutput and make it more like an interface. This should be a small step, right? It can be in the same file or in a seperate one, doesn't really matter to me.

WilcoBT:
Well, it could be a reasonably small step, the one commit for this line + some makefile stuf. 
But then first CommandlineTestRunner needs to be refactored differently then I did before to make it better testable. That is the step that worries me as I fear weird stuff coming from the plugin system.

arstrube:
@basvodde - apologies for even more discussion ;-) but there are a few points we need to consider before going too deep into coding.

1. If JUnitTestOutput is to be a decorator (which I believe we agree it should be), then the base class `TestOutput` can't contain any methods that are not used by it, since the whole point of the decorator pattern is that clients only ever use this interface (and not specialized methods that decorators might contain for their own private use).

2. Given (1) such methods as `print(<various basic types>)` cannot be called by clients.

3. Currently, we are heavily relying on such methods, namely:
```
output_->verbose()    CommandLineTestRunner.cpp:90
output_->color()      CommandLineTestRunner.cpp:91
output_->print(char*) CommandLineTestRunner.cpp:124
output_.print(char*)  TestResult.cpp:64
many                  TestOutputTest.cpp:many`
```
The first two would probably go away if we supply ColorConsoleOutput and VerboseConsoleOutput decorators

I would suggest thinking about the various direct uses of print() (as opposed to printTestRun() etc, or even print(TestFailure). If we can clean up those,

Finally, and unfortunately, there are numerous violations in TestOutputTest.cpp, so that would need major rework before we evens start.

I'm afraid if we don't solve this issue first, we are going to have little joy with decorators. Maybe as a _very_ first step, we should do decorators for verbose and color to the existing ConsoleOutput only and properly integrate those with CommandLineTestRunner.cpp?

basvodde:
Not sure why it would need to be refactored, but lets take small steps :) Lets first pull out the console things and then take the next step.

basvodde:
@arstrube Actually, I think we don't need to make JUnitOutput a decorator, but we just need a TestOutput Composite, and that would do already.

WilcoBT:
Hmm, maybe I do see a point, would be even more classes and I wonder if it wouldn't be very order dependend but just adding the verbose and color as seperate decorators. Could be messy with the current flushing etc.

basvodde:
@WilcoBT @arstrube Suggest to keep verbose and color just in the ConsoleOutput for now, we want to merge code and not make it perfect now :)

WilcoBT:
Well, I'll give it a slight shot this morning to get my current changes separated in a few steps that can be pulled sequentially but I do not have much hope that I can do it in that time.

WilcoBT:
I think I have it :) Got the TestOutput refactoring working locally on a separate branch :)

arstrube:
I'm fiddling with a decorator just for ConsoleOutput/StringBufferOutput just for the color aspect (verbose would have been easier. I'll push that now-now; it looks promising as a small step, in principle

WilcoBT:
Just for completeness, the requested separate refactoring of TestOutput is in #592 but depends on #591.

PR #590, opened 2015-04-08
Print test group names

mmidor:
* added commandline switch -lg for printing all test group names instead of running them
* added switch -ln printing list of all group.test pairs
* switch -ln use groups filter
* tests added for new command line arguments

### DIFF
--- a/src/CppUTest/TestRegistry.cpp
+++ b/src/CppUTest/TestRegistry.cpp
@@ -74,28 +74,34 @@ void TestRegistry::runAllTests(TestResult& result)
 
 void TestRegistry::listTestGroupNames(TestResult& result)
 {
-    SimpleString lastgroup = "";
+    SimpleString groupList;
 
     for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
         SimpleString gname = test->getGroup();
-        if (gname != lastgroup) {
-            lastgroup = gname;
-            result.print(gname.asCharString());
-            result.print(" ");
+        if (!groupList.contains(gname)) {
+            groupList += gname;
+            groupList += " ";
         }
     }
+    result.print(groupList.asCharString());
 }
 
 void TestRegistry::listTestGroupAndCaseNames(TestResult& result)
 {
+    SimpleString groupAndNameList;
+
     for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
         if (testShouldRun(test, result)) {
-            result.print(test->getGroup().asCharString());
-            result.print(".");
-            result.print(test->getName().asCharString());
-            result.print(" ");
+            SimpleString groupAndName = test->getGroup();
+            groupAndName += ".";
+            groupAndName += test->getName();
+            groupAndName += " ";
+            if (!groupAndNameList.contains(groupAndName)) {
+                groupAndNameList += groupAndName;
+            }
         }
     }
+    result.print(groupAndNameList.asCharString());
 }
 
 bool TestRegistry::endOfGroup(UtestShell* test)
--- a/src/CppUTest/TestRegistry.cpp
+++ b/src/CppUTest/TestRegistry.cpp
@@ -72,6 +72,32 @@ void TestRegistry::runAllTests(TestResult& result)
     currentRepetition_++;
 }
 
+void TestRegistry::listTestGroupNames(TestResult& result)
+{
+    SimpleString lastgroup = "";
+
+    for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
+        SimpleString gname = test->getGroup();
+        if (gname != lastgroup) {
+            lastgroup = gname;
+            result.print(gname.asCharString());
+            result.print(" ");
+        }
+    }
+}
+
+void TestRegistry::listTestGroupAndCaseNames(TestResult& result)
+{
+    for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
+        if (testShouldRun(test, result)) {
+            result.print(test->getGroup().asCharString());
+            result.print(".");
+            result.print(test->getName().asCharString());
+            result.print(" ");
+        }
+    }
+}
+
 bool TestRegistry::endOfGroup(UtestShell* test)
 {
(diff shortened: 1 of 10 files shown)


src/CppUTest/TestRegistry.cpp
```diff
@@ -72,6 +72,32 @@ void TestRegistry::runAllTests(TestResult& result)
     currentRepetition_++;
 }
 
+void TestRegistry::listTestGroupNames(TestResult& result)
+{
+    SimpleString lastgroup = "";
+
+    for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
+        SimpleString gname = test->getGroup();
+        if (gname != lastgroup) {
```
basvodde: This sucks :) @arstrube more prove we need to make TestGroup in a separate thing.

basvodde:
It looks pretty good, but the TestRegistry functionality doesn't have any tests around it ?

mmidor:
No, it hasn't. I wondered for some time how to run external app, catch it's output and check if returned list is what I expected, but it is too complicated. In this case I need to prepare separate app containing known groups and tests.
Functionality is simple and works pretty good, which is the most important to me. 
If you have some suggestions how to do more reliable tests, I'll try to implement them.

basvodde:
You create a TestRepository in a test, you add some tests to it and check the output.

You can see this done similarly in the TestRegistryTest.cpp

mmidor:
Hi,
can you look at the latest changes I made?

arstrube:
I looked at the Ctest output on Travis CI and it is indeed a lot more useful than before. The code is neat and the test are all there - what do you think @basvodde?

mmidor:
Last improvement added. Nothing more to do from my side.

tests/TestRegistryTest.cpp
```diff
@@ -323,3 +323,38 @@ TEST(TestRegistry, ResetPluginsWorks)
...
+    test2->setGroupName("GROUP_2");
+    myRegistry->addTest(test2);
+    test3->setGroupName("GROUP_1");
+    myRegistry->addTest(test3);
+
+    myRegistry->listTestGroupNames(*result);
+    SimpleString s = output->getOutput();
+    STRCMP_CONTAINS("GROUP_1 ", s.asCharString());
+    STRCMP_CONTAINS("GROUP_2 ", s.asCharString());
+    LONGS_EQUAL(1, s.count("GROUP_1 "));
```
basvodde: All looks good, but here I'm wondering... why do you not assert on the exact string and instead use STMCMP_CONTAINS?

mmidor: Because the order of the groups and tests is not very important, in fact, the tests are listed backward (so the tests are executed from the most advanced to the simplest, defined at the file beginning, which was confusing me initially).
I found this approach more universal. 

If you think different, point me the best solution. 
And BTW I hope it will be merged soon into the trunk, otherwise we can't use such a unofficial modification.

Regards, 
MM

arstrube: @mmidor I agree. Otherwise, one would have to employ TEST_ORDERED() or rely on the unspecified (backward) ordering of tests, which is bad. There is no good enough reason to use TEST_ORDERED() in this case.

basvodde: Not really. The test order is quite fixed in this case and you have only one test. If we would decide to change the test order and this test fails, ok then we'll fix it. But now you have 3 'weak' asserts where you can have one 'strong' assert. So, I'd suggest to change it to a stonger assert that asserts the exact output.

mmidor:
Done.

tests/TestRegistryTest.cpp
mmidor: Done.

tests/TestRegistryTest.cpp
```diff
@@ -323,3 +323,36 @@ TEST(TestRegistry, ResetPluginsWorks)
...
+    test1->setGroupName("GROUP_1");
+    myRegistry->addTest(test1);
+    test2->setGroupName("GROUP_2");
+    myRegistry->addTest(test2);
+    test3->setGroupName("GROUP_1");
+    myRegistry->addTest(test3);
+
+    myRegistry->listTestGroupNames(*result);
+    SimpleString s = output->getOutput();
+    STRCMP_EQUAL("GROUP_1 GROUP_2 ", s.asCharString());
```
basvodde: Better. Now, I see there is the extra space at the end. Do we want to do something about that? I assume it can lead to nasty surprises when people use this functionality?

tests/TestRegistryTest.cpp
```diff
@@ -323,3 +323,36 @@ TEST(TestRegistry, ResetPluginsWorks)
...
+    myRegistry->addTest(test2);
+    test3->setGroupName("GROUP_C");
+    test3->setTestName("test_c");
+    myRegistry->addTest(test3);
+
+    myRegistry->listTestGroupAndCaseNames(*result);
+    SimpleString s = output->getOutput();
+    STRCMP_CONTAINS("GROUP_A.test_a ", s.asCharString());
+    STRCMP_CONTAINS("GROUP_B.test_b ", s.asCharString());
+    STRCMP_CONTAINS("GROUP_C.test_c ", s.asCharString());
```
basvodde: Can you here also assert on the whole string?

tests/TestRegistryTest.cpp
mmidor: There is no one best solution about trailing delimiters. A lot of people rely on them and I was changing my software in the past to gave clients what they wanted.

After all it's easier to add trailing delimiter than remove it so I will change it.

tests/TestRegistryTest.cpp
mmidor: Never ending story - Okay ;)

basvodde: Should be ending pretty soon as I don't think I got much more comments! Thanks for your patience.

basvodde:
Thanks!

mmidor:
Thanks. That was a good lesson of achieving a perfection :)

basvodde:
Ooohh.... that wasn't perfection, that was still pragmatic :) Perfection is a lot further away still :)

PR #591, opened 2015-04-09
Issue571, step 1: Refactor CommandlineTestRunner 

WilcoBT:
Refactor CommandlineTestRunner such that the parsing of arguments and creation of Testoutput based on this result is extracted.

"int CommandLineTestRunner::RunAllTests(int ac, const char** av)"
Is getting a bit long, in a later step where the decorator pattern is applied the initialising of the "output" variable should be put in a separate function.

### DIFF
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -31,16 +31,14 @@
 #include "CppUTest/JUnitTestOutput.h"
 #include "CppUTest/TestRegistry.h"
 
-CommandLineTestRunner::CommandLineTestRunner(int ac, const char** av, TestOutput* output, TestRegistry* registry) :
-    output_(output), jUnitOutput_(NULL), arguments_(NULL), registry_(registry)
+CommandLineTestRunner::CommandLineTestRunner( TestOutput* output, TestRegistry* registry, CommandLineArguments* arguments ):
+    output_(output), arguments_(arguments), registry_(registry)
 {
-    arguments_ = new CommandLineArguments(ac, av);
+
 }
 
 CommandLineTestRunner::~CommandLineTestRunner()
 {
-    delete arguments_;
-    delete jUnitOutput_;
 }
 
 int CommandLineTestRunner::RunAllTests(int ac, char** av)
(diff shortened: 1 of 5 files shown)


basvodde:
I'm not quite understanding why we need to do this?

WilcoBT:
Because otherwise I can't cleanly instantiate the correct ConsoleTestOutput in the next step.

basvodde:
I'm still not understanding, but let me make a counter proposal tomorrow :)

WilcoBT:
This is basically a cleaner solution for commit 1 in  "Issue571 attempt2:" This whole thing got started as you asked commit 2 to be a separate merge. As it need commit 1 and the goals was to keep the changes in one merge as small as possible I gave this also one of its own.

But I'm curious to your proposed solution.

basvodde:
Yup, but I was first expecting the small change of pulling out the ConsoleOutput stuff from the TestOutput...

WilcoBT:
Small is relative :) it still says +587 −421 in that pull request :) but I think a lot of the difference is the fact that more files means more copyright statements.

basvodde:
Did you make that one already?

WilcoBT:
And about why we need this. This means that you can separately test if given a set of arguments, what the TestOutput will become.

Yes but broke it: https://github.com/cpputest/cpputest/pull/589

basvodde:
Yup... but... :)

WilcoBT:
hmm, referenced the wrong one, meant #592 
And being able to separately test the TestOutput generation also fixes my issue in #589 of the CommandlineTestRunnerTest testing to much at once because I couldn't stub the TestOutput.

src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -51,21 +49,44 @@ int CommandLineTestRunner::RunAllTests(int ac, char** av)
...
 
     if (result == 0) {
-        output << memLeakWarn.FinalReport(0);
+        *output << memLeakWarn.FinalReport(0);
     }
-    TestRegistry::getCurrentRegistry()->removePluginByName(DEF_PLUGIN_MEM_LEAK);
+    registry->removePluginByName( DEF_PLUGIN_MEM_LEAK );
+
+    delete output;
+
```
basvodde: The reason why the original code was the way it was is that we want *all* dynamic memory allocation to be within the memLeakWarn and FinalReport to be done. That is why the original ConsoleTestOuput was created as a stack object, so that it doesn't allocate memory and doesn't interfere with the memory leak checking.

So, the design here will probably need to keep the RunAllTests to be the same (minimal) while being able to change the TestOutput.

basvodde:
The other pull requests was baselined on this one, which means I can't merge it separately :(

WilcoBT:
I also failed cherry picking it and getting it compiling and testing

src/CppUTest/CommandLineTestRunner.cpp
WilcoBT: It sounds sensible and the moment the 3rd step of the decorator I think I see again how that is possible but I did not know how to do it now. Well, maybe with that if else within the if succesfullyParsed block and then create the testrunner with the one or the other.

WilcoBT:
Alright, no "new"s anymore :)

basvodde:
How about the change in #595. I guess this change is enough as now you can set the secondaryOutput_ to whatever you want, while always keeping the primaryOutput to be a TestConsoleOutput.

PR #592, opened 2015-04-09
Issue571, step 2: Refactor TestOutput preparing for inserting decorator.

WilcoBT:
Moved everything to specialised classes leaving TestOutput only as the interface so lateron a TestOutputDecorator can be placed right below TestOutput in the inheritance chain(s)

### DIFF
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -146,6 +146,7 @@ copy $(OutDir)CppUTest.lib lib\CppUTest.lib
     <ClCompile Include="src\CppUTestExt\OrderedTest.cpp" />
     <ClCompile Include="src\CppUTest\CommandLineArguments.cpp" />
     <ClCompile Include="src\CppUTest\CommandLineTestRunner.cpp" />
+    <ClCompile Include="src\CppUTest\ConsoleTestOutput.cpp" />
     <ClCompile Include="src\CppUTest\JUnitTestOutput.cpp" />
     <ClCompile Include="src\CppUTest\MemoryLeakDetector.cpp" />
     <ClCompile Include="src\CppUTest\MemoryLeakWarningPlugin.cpp" />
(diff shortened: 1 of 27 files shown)


WilcoBT:
Hmm, maybe I have teased this system to much with forcing changes... It complains about unused variables that in the diff are already disabled...

WilcoBT:
Is there a way to fix this or should I just go for another pull request?

basvodde:
@WilcoBT I'm not sure you followed the other pull requests. I did 3 pull requests to implement this feature. The last one finishes it, I left it open for now at #610 

With these, I guess the feature is done and these can be closed, correct?

WilcoBT:
I had been wondering about it indeed as the conversation in #595 seemed to have stopped in the middle but not had the time to check how it is going and only get alerts for those things I started.
Now catching up.

basvodde:
Ok, let me give you the pull request numbers so you can see the solution I proposed:

- Step 1. Refactored the CommandLineRunner in #595
- Step 2. Introduced CompositeTestOutput in #604
- Step 3. Introduced the -v -ojunit in CommandLineRunner in #610

The last one is still open for comments.

WilcoBT:
Ow hey, I wasn't subscribed to that first one! Seems that doesn't happen automatically.

basvodde:
Yeah, I figured you were either not monitoring the new pull requests or had dropped out due to other things to do :)

WilcoBT:
Well, I'm happy the combination now works and I see you managed to do it with much smaller changes. The difference in familiarity with the code and that I'm not one for being conservative.

I guess the next steps for me will be to close my mess of pull requests and finally get cpputest as a real submodule of our code instead of the copy pasta we have now. Which will bring me into the interesting world of git mirrors and such.

basvodde:
Yeah, also why I decided to do the steps myself instead of commenting forever on the pull request ;)

WilcoBT:
Thanks :)

PR #593, opened 2015-04-09
Suggestion for first step: Extract color into decorator.

arstrube:
NOTE:
1. This still needs improvement - for example, should probably be initialized in the same placeas junit output
2. For some reason, the combination -v -c doesn't work. This will probably go away once -v is also handled via decorator.
3. Slight changes to the expected output -- the newlines are now in color ;-)
4. Even this small change affects five files :-(

### DIFF
--- a/include/CppUTest/CommandLineTestRunner.h
+++ b/include/CppUTest/CommandLineTestRunner.h
@@ -56,6 +56,7 @@ public:
 
 protected:
     TestOutput* output_;
+    ColoredTestOutput* colorOutput_;
     JUnitTestOutput* jUnitOutput_;
 
 private:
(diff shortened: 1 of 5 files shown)


arstrube:
O yeah - function implementation probalby shouldn't stay inline, either...

arstrube:
Hmm. We are obviously missing a test case ;-) Even Travis CI didn't catch that -v and -c don't work together...

basvodde:
But why would we do this? There is no benefit at the moment for pulling out the coloring, it is always in the console.

arstrube:
To not have it in  __TestOutput__, from where it will be inherited by JUnitOutput, which has no use for it.
Same goes for verbose().

arstrube:
Truth told, I figured that extracting the ConsoleOutput stuff from TestOutput is _not_ a small step, really... that's why I attempted an even smaller one :-) 

It was a good learning exercise.

PR #595, opened 2015-04-10
Refactored CommandLineRunner to allow for new TestOutput

basvodde:
Small refactoring to allow for pulling out the ConsoleOutput further.

### DIFF
--- a/include/CppUTest/CommandLineTestRunner.h
+++ b/include/CppUTest/CommandLineTestRunner.h
@@ -56,6 +56,8 @@ public:
 
 protected:
     TestOutput* output_;
+    TestOutput* primaryOutput_;
+    TestOutput* secondaryOutput_;
     JUnitTestOutput* jUnitOutput_;
 
 private:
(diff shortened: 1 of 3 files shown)


arstrube:
That's a minimal step for sure. Looks good to me.

I did an actual refactoring of TestOutput (no PR yet). I'm not entirely happy with it. It also tends to affect a lot more files (that's in the nature of things). I will combine yours with that tongight; the should complement each other well.

I'm haveing problems with StringButterTestOutput vs. ConsoleTestOutput, needing at least one cast. No casts were necessary before, so I don't like that.

arstrube:
P.S. I guess it wasn't so much for me to comment, but it looks good anyway :)

basvodde:
Simplified the CommandLineRunner further. Now it doesn't have the complication it had before and less parameters

arstrube:
Hmm. What are the advantages over my PR #596? Does this already print -v and -junit concurrently, or is just a preparatory step?

arstrube:
@basvodde okay, I merged this with my PR #596 and it merges with no problems and works as expected. Although this is kind of optional ;-). I also discovered an oversight in that failures weren't printed, which I corrected.

PR #596, opened 2015-04-11
cso - print verbose output with -junit

arstrube:
This will print verbose output along with junit output, if -v -ojunit is specified
+ uses composition
+ only minimal refactoring
+ doesn't break any existing tests
+ @basvodde sorry, had no good reason to use your suggestion, but it could be used without breaking anything

### DIFF
--- a/include/CppUTest/JUnitTestOutput.h
+++ b/include/CppUTest/JUnitTestOutput.h
@@ -37,7 +37,7 @@ struct JUnitTestCaseResultNode;
 class JUnitTestOutput: public TestOutput
 {
 public:
-    explicit JUnitTestOutput();
+    explicit JUnitTestOutput(ConsoleTestOutput* output = NULL);
     virtual ~JUnitTestOutput();
 
     virtual void printTestsStarted() _override;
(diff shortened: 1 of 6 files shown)


arstrube:
Sorry, the diff for TestOutput.h didn't come out too good. The changes are not as dramatic as it looks :)

include/CppUTest/JUnitTestOutput.h
```diff
@@ -74,6 +74,9 @@ class JUnitTestOutput: public TestOutput
     virtual void writeFailure(JUnitTestCaseResultNode* node);
     virtual void writeFileEnding();
 
+private:
+    ConsoleTestOutput* currentConsoleOutput_;
+    ConsoleTestOutput defaultConsoleOutput_;
```
arstrube: @basvodde I have to admit I think your solution is more general and thus, better. I was wondering though, whether using a default output might be advantageous (that would make two for CompositeTestOutput)

1. new and delete would alway happen where they're supposed to
2. No need to test whether currentOtuptu is NULL because it won't be

The default would probably be StringBufferTestOutput.

basvodde: Without having an actual case for using it, I'd probably not want to add the complexity. This was also why I eventually simplified the CommandLineTestRunner

arstrube: I see. Here, the use case is that I change JUnitTestOutput, so it provides a way for the existing constructor to work, as well as the one that passes a console output.

I guess I'm not seeing the whole picture because it hasn't been integrated yet with the CLR.

arstrube:
Closing, as this functionality has been implemented by PR #604.

arstrube:
Closing, as this feature has been impelmented by PR #604.

arstrube:
Closing, as this functionality has been implemented by PR #604.

PR #601, opened 2015-04-20
cso2 - print verbose output with -junit

arstrube:
Same as cso, except there are no refactorings whatsoever

+ This sports much fewer changes
+ the result does feel somewhat messier (at least to me)


basvodde:
Can be closed?

arstrube:
Sure.

PR #604, opened 2015-04-21
Composite Test Output

basvodde:
Merged the Composite in the existing header and source. Also added some functionality.

### DIFF
--- a/include/CppUTest/TestOutput.h
+++ b/include/CppUTest/TestOutput.h
@@ -156,12 +156,50 @@ public:
         return output;
     }
 
-private:
+protected:
     SimpleString output;
 
+private:
     StringBufferTestOutput(const StringBufferTestOutput&);
     StringBufferTestOutput& operator=(const StringBufferTestOutput&);
 
+};
+
+class CompositeTestOutput : public TestOutput
+{
+public:
+    virtual void setOutputOne(TestOutput* output);
+    virtual void setOutputTwo(TestOutput* output);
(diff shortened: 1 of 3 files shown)


arstrube:
1. Looking good.
2. Might default outputs work here (see my PR #596 )? It made a lot of sense there, but I'm not sure yet how you plan to hook it up with CommandLineTestRunner.

basvodde:
It was actually 4 additional lines in CommandLineTestRunner. I'll put them in a pull request tomorrow.

arstrube:
There is a strange test failure I noticed working on something else in VC6:
```bash
D:\DEV\05_CPPUTEST\CPPUTEST\tests\TestOutputTest.cpp(381): error: Failure in TEST(CompositeTestOutput, PrintTestFailure)
 expected <
file:10: error: Failure in TEST(Group, Name)
 failed
>
 but was  <
file(10): error: Failure in TEST(Group, Name)
 failed
>
 difference starts at position 5 at: <     
file(10): erro>
                                           
```

arstrube:
Ah, I see. This bug is a feature ;-)

Looks like you need to:
1) set the working environment to eclipse in setup()
2) do an additional test with the working environment explicitly set to vistualStudio

basvodde:
Only 1) and only in one test. Let me do that later.

arstrube:
....well not that I disagree but that's how the existing tests are done :-)

PR #605, opened 2015-04-21
MEMCMP_EQUAL assertion

KisImre:
I've implemented the MEMCMP_EQUAL feature in my fork. Feel free to use any part of my code if you find it useful. It's output look like this:
```
../mainTest.cpp:16: error: Failure in TEST(MemCmpDemo, Demo)
	expected <01 23 45 67>
	but was  <01 23 00 67>
	difference starts at position 2 at: <   01 23 00 67      >
	                                               ^
```

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -189,6 +189,13 @@
 #define DOUBLES_EQUAL_LOCATION(expected,actual,threshold,file,line)\
   { UtestShell::getCurrent()->assertDoublesEqual(expected, actual, threshold,  file, line); }
 
+//Check two memory areas for equality
+#define MEMCMP_EQUAL(expected,actual,size)\
+  MEMCMP_EQUAL_LOCATION(expected,actual,size,__FILE__,__LINE__)
+
+#define MEMCMP_EQUAL_LOCATION(expected,actual,size,file,line)\
+  { UtestShell::getCurrent()->assertBinaryEqual(expected, actual, size, file, line); }
+
 //Fail if you get to this macro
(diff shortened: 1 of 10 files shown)


include/CppUTest/UtestMacros.h
```diff
@@ -189,6 +189,13 @@
 #define DOUBLES_EQUAL_LOCATION(expected,actual,threshold,file,line)\
   { UtestShell::getCurrent()->assertDoublesEqual(expected, actual, threshold,  file, line); }
 
+//Check two memory areas for equality
```
basvodde: Would you mind removing the comment, as it is pretty obvious :)

basvodde:
Hey, this looks pretty good. Thanks! Finally someone who has made the output look good and tests it reasonably ;)

Would you mind removing the one comment and adding some tests for the macro still? Then I think it is ok.

@arstrube Any comment on this still?

arstrube:
Nice. Looks good to me :-)

KisImre:
I'm almost done with the tests for the macro. However I was wondering that what should the assertion do if both pointers are NULL. The current implementation doesn't fail, so it works the same way as the string assertions. Is this the required behavior of the MEMCMP_EQUAL?

arstrube:
Whether it is required behavior, I cannot say. However, this is what I would personally expect from CppUTest, given the behavior of the string comparisons. Because of that, I would also expect that it doesn't crash if either pointer (or both) are NULL. 

I am aware that this behavior cannot be taken for granted; normally something might try to dereference the pointer, and it might crash.

basvodde:
Thanks, appreciated!

PR #606, opened 2015-04-21
ts - Testing platform-specific code with a wrapper

arstrube:
This one is for discussion.

It exemplifies a method that would allow to test actual platform-specific code, where the tests test the interface and the wrapper is implemented in a platform-specific way.

The wrapper allows stubbing stuff that otherwise cannot be stubbed. For example, PlatformSpecificTimeString() internally call time(), which you would need to stub to make sure the rest of the function does what it should.

The method employed is *slightly* border-line. I have personally used it in some very tricky situations where production code could not be changed.

It is possible to test the file operations in this way, however, I have not included that here, since there are some tests in the Travis CI build that rely on actual file output being produced.

### DIFF
--- a/CppUTest.dsp
+++ b/CppUTest.dsp
@@ -90,54 +90,6 @@ SOURCE=.\src\CppUTestExt\CodeMemoryReportFormatter.cpp
 # End Source File
 # Begin Source File
 
-SOURCE=.\src\CppUTestExt\MemoryReportAllocator.cpp
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MemoryReporterPlugin.cpp
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MemoryReportFormatter.cpp
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockActualCall.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockExpectedCall.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockExpectedCallsList.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockFailure.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockNamedValue.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockSupport.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockSupportPlugin.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\MockSupport_c.cpp"
-# End Source File
-# Begin Source File
-
-SOURCE=.\src\CppUTestExt\OrderedTest.cpp"
-# End Source File
-# Begin Source File
-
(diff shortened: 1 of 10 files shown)


CppUTest.dsp
```diff
@@ -90,71 +90,71 @@ SOURCE=.\src\CppUTestExt\CodeMemoryReportFormatter.cpp
 # End Source File
 # Begin Source File
 
-SOURCE=.\src\CppUTestExt\MemoryReportAllocator.cpp
```
arstrube: The changes in the .dsp are a bit annoying. Would have to be edited manually. Essentually, we add one file to AllTests. That's all.

include/CppUTest/SimpleString.h
```diff
@@ -150,7 +150,6 @@ SimpleString VStringFromFormat(const char* format, va_list args);
 #if CPPUTEST_USE_STD_CPP_LIB
 
 #include <string>
-#include <stdint.h>
```
arstrube: This is a separate pull request

basvodde:
I do not see the use of this. The PlatformSpecific method should already be small wrappers and now you are testing the wrapper by adding a thinner wrapper. Whereas the problems in the PlatformSpecific methods almost always is in that the behavior of the OS is different, not in what the wrapper code is wrong.

arstrube:
I do agree with you where _real_ thin wrappers are concerned.
```c++
static const char* TimeStringImplementation()
{
    time_t tm = time(NULL);
    static char dateTime[80];
    struct tm *tmp = localtime(&tm);
    strftime(dateTime, 80, "%Y-%m-%dT%H:%M:%S", tmp);
    return dateTime;
}
```
This is not just a wrapper. It contains functionality that deserves testing, and that cannot be tested due to the assignment of time() stuck in there.

arstrube:
How about we get rid of GetPlatformSpecificTimeString() altogether:

1. It is never used anywhere in CppUTest
2. It is not useful, because it is limited to one particular format that the user cannot influence
3. Because of 1), it is completely untested
4. It is also untestable (well, testable only by means that @basvodde disapproves ;-)

I think the method has merit in itself, when it comes to testing really tough pieces of code. But it might be better placed in a book on testing legacy code than here :-)...

basvodde:
1) It is used in JUnitTestOutput
2) It is useful in the JUnit output
3) It is in wide use
4) It is quite a thin wrapper around a few PlatformSpecific calls
5) But it is used :)

In this case, I do not consider the effort going through to test the wrapper worth it.

arstrube:
Hmm. That's interesting, because in that case it is stubbed out everywhere, for every test, always. 
Which, in my book, would be a good reason to test it, not to omit the test... Since when is the relative simplicity of a function a reason not to test it at all ?

basvodde:
Well, the answer is more complex. Most of my "testing" is not for testing purpose but to test-drive my code. Which is why I never go back to add tests to places that aren't tested as it rarely is worth the effort.

A common problem is unit testing is 3rd party dependencies, such as platform. So this is usually stubbed out so you can test-drive the logic without the dependency. But then you need a small wrapper, which I usually do not test-drive but just manually check whether it is ok. There should be no major logic in that wrapper, otherwise ... it isn't a thin wrapper anymore.

So, for me it is the wrapper-nature that causes simplicity and it is that wrapper-nature that causes me to not write a test as it is *only* dependency with the OS :)

In this case, it is 5 lines of which 3 have a OS call. One is a return statement and the other is a variable declaration :) The only logic is the format. We could pull that out, but don't see a huge need for that at the moment. Other than that, it is just OS calls :)

PR #610, opened 2015-04-22
Verbose and JUnit

basvodde:
Step #3 of getting verbose and JUnitOutput without ever touching the JUnitOutput class.

### DIFF
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -137,6 +137,14 @@ TestOutput* CommandLineTestRunner::createConsoleOutput()
     return new ConsoleTestOutput;
 }
 
+TestOutput* CommandLineTestRunner::createCompositeOutput(TestOutput* outputOne, TestOutput* outputTwo)
+{
+  CompositeTestOutput* composite = new CompositeTestOutput;
+  composite->setOutputOne(outputOne);
+  composite->setOutputTwo(outputTwo);
+  return composite;
+}
+
 bool CommandLineTestRunner::parseArguments(TestPlugin* plugin)
 {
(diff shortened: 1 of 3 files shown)


src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -145,8 +153,11 @@ bool CommandLineTestRunner::parseArguments(TestPlugin* plugin)
...
 
-  if (arguments_->isJUnitOutput())
+  if (arguments_->isJUnitOutput()) {
     output_= createJUnitOutput(arguments_->getPackageName());
+    if (arguments_->isVerbose())
+      output_ = createCompositeOutput(output_, createConsoleOutput());
+  }
   else
     output_ = createConsoleOutput();
   return true;
```
arstrube: I guess I dont like the new which is not balanced by a delete.
I also don't really like that a different output is created depending on the arguments.
My solution required no changes to CommandLineTestRunner....
In isolation, I like your CompositeTestOutput.
However, I do feel that my solution integrates more nicely into the existing code.

I there some way that you can avoid new'ing in one place and delete'ing in another?

basvodde: In case of a Composite, the new/delete mismatch makes sense as the Composite must take ownership of it for it to be treated as one.

I don't like the changes needed for JUnitOutput as the functionality requested isn't related to jUnitOutput, but it is related to CommandLineRunner. So, to me, this change makes a lot more sense and JUnitOutput doesn't know about ConsoleTestOutput... as it shouldn't.

arstrube: You are right about that JUnitOutput shouldn't know about it, which is why I prefer your solution, too.

I just wish we could do the test runner side more elegantly....

PR #614, opened 2015-04-23
Fix examples for VC6 .. VS2005 etc.

arstrube:
1. Made the .dsp files usable again
2. Removed confusing path to platform specific code
3. Got rid of linker problems

@Andne, @PaulBussmann please double-check

P.S. If VC6 is not available, can be imported into VS2008 and should behave properly. Can be imported into VS201x via VS2008.


arstrube:
Got rid of the messy diff for CppUTest.dsp.
Undoing the mis-merge looks messy in itself; look at individual commits to get a clearer picture of the changes.
Both Debug and Release targets now working properly.

Issue #615, opened 2015-04-24
PlatformSpecificRunInSeparateProcess() Fails in Debugger under Mac OS X


arstrube:
@basvodde reports that, when run in the debugger under Mac OS X Yosemite, all test cases that actually run in a separate process fail, with waitpid() returning -1 and ERRNO set to EINTR

The beginning of the discussion can be found in #608.

As this bug will involve fixing PlatformSpecificRunInSeparateProcess(), I started a new issue.

arstrube:
As for looping "forever" -- actually we are talking about a new loop here, not the existing one.

The existing one will end when the child process returns.

The new "forever" loop would end when fork() doesn't return -1 with errno==EINTR. Which may or may not happen.... We'll see.

arstrube:
@basvodde -- as an aside, how do you actually debug on your Mac? I tend to use the Code::Blocks IDE under Windows, but the Mac version has a bug preventing it from starting the debugger properly. Using the commandline debugger to me feels a bit like driving a car over a bumpy road in the dark with no headlights on ;-).

basvodde:
I use a command line debugger :) I love it :)

arstrube:
Oookaaay.... I use the command line for quite a few other things, so ... any good place to get started?

basvodde:
Uhm, donno actually. Cheat sheet? I've used it for so long, it is quite automatic for me. But there is only like 10 commands you frequently use.

arstrube:
Fixed by #616.

arstrube:
Now you mention it -- how come you've even got gdb on your Mac? Mine only has lldb. The little I tried, I definitely find standalone gdb a lot more useful than standalone lldb (tried the Cygwin one).

arstrube:
Okay. I got gdb running. Quite a performance.

arstrube:
which one is your debugger?

basvodde:
gdb

PR #616, opened 2015-04-24
Fix Mac OS X debugger failures

arstrube:
caused by waitpid() failing with errno == EINTR.

Fixes issue #615.

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -44,6 +44,7 @@
 #include <signal.h>
 #ifndef __MINGW32__
 #include <sys/wait.h>
+#include <errno.h>
 #endif
 #include <pthread.h>
 
@@ -83,6 +84,7 @@ static void GccCygwinPlatformSpecificRunTestInASeperateProcess(UtestShell* shell
         do {
             w = PlatformSpecificWaitPid(cpid, &status, WUNTRACED);
             if (w == -1) {
+                if(EINTR ==errno) continue; /* OS X debugger */
                 result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
                 return;
             }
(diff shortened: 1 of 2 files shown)


src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -72,8 +73,13 @@ static void GccCygwinPlatformSpecificRunTestInASeperateProcess(UtestShell* shell
     cpid = PlatformSpecificFork();
 
     if (cpid == -1) {
-        result->addFailure(TestFailure(shell, "Call to fork() failed"));
-        return;
+        if(errno == EINTR) { /* Retry once if interrupted */
```
arstrube: Using `goto` would be really neat here -- but only if we knew for sure that the fork would eventually succeed...

basvodde:
I don't understand. It is the waitpid that you need to retry and not the fork ?

basvodde:
And you do want to use a forever loop :)

arstrube:
You are right. But then I am not quite sure what they mean with "restart"? Simply keep waiting?

arstrube:
Okay. Will change test and code accordingly.

arstrube:
At least in that case, `continue` will work fine

basvodde:
The easiest is to write:

while ((errorCode = wait(..)) != EINTR)
   ;

arstrube:
Just pushed the fix to the correct problem.

The test sets EINTR only ten times. It will be interesting to see what the debugger does.

basvodde:
Ok, looks good. I'll merge when travis is done.

arstrube:
I think the test doesn't have do return EINTR 10 times, maybe? How about using a `bool` instead and only returning EINTR once? This should be sufficient to exercise the new code properly, and still fail with the old code?

basvodde:
It worked. Tests don't fail in debugger.

arstrube:
Glad to hear :)

For the record - these tests helped discover and fix a real problem...

PR #617, opened 2015-04-24
Fixed the problem of optimizing away division by zero

basvodde:
Moved the division in a separate compilation unit so that the compiler can't optimize it away anymore. HAH :)


arstrube:
Cunning :)

I like the FAIL() in case it goes wrong, so we will know what happened, should compilers ever get even clevererer than that.

PR #619, opened 2015-04-27
Cleaned up the project output folders

Andne:
Please do not merge until I say so, still need to make changes and want to see if appveyor is handling the pull requests.

I still need to make some changes to VS2008 to match, but this should fix the VS2010 build at least.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ gcov*.html
 ErrorLogs
 testResults
 .metadata
+/ipch
 Debug
 Release
 */Debug/*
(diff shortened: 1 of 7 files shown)


Andne:
And now cleaning up some other settings has caused me to break the 2008 build... I'll fix it once I get back to a computer that has 2005/2008 installed on it (don't have a computer like that at work).

basvodde:
Merge it?

Andne:
Probably, sure.

I think I have VS2005/VS2008 and VS2010 working fine.  Looks like I could clean up the files a bit more, but this is pretty much the minimum to get a working build.  I also added to appveyor.yml the settings to archive the library file so that it can be downloaded for a given build.  Not that I'm sure if anyone would want to do that or not.

PR #627, opened 2015-05-01
Test JUnitOutput properly again

arstrube:
Due to CommandLineTestRunnerWithStringBufferOutput overriding some methods for testing, the original methods weren't tested anymore.

This test tests the original methods.

### DIFF
--- a/tests/CommandLineTestRunnerTest.cpp
+++ b/tests/CommandLineTestRunnerTest.cpp
@@ -139,3 +139,23 @@ TEST(CommandLineTestRunner, JunitOutputAndVerboseEnabled)
     STRCMP_CONTAINS("TEST(group, test)", commandLineTestRunner.fakeJUnitOuputWhichIsReallyABuffer_->getOutput().asCharString());
     STRCMP_CONTAINS("TEST(group, test)", commandLineTestRunner.fakeConsoleOutputWhichIsReallyABuffer->getOutput().asCharString());
 }
+
+TEST(CommandLineTestRunner, listTestGroupNamesShouldWorkProperly)
+{
+    const char* argv[] = { "tests.exe", "-lg" };
+    
+    CommandLineTestRunnerWithStringBufferOutput commandLineTestRunner(2, argv, &registry);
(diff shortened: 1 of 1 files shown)


basvodde:
I'll merge it, but myself think these tests are not really needed.

arstrube:
Thanks :).

If it is not really needed to test this function (because it is kind of wrapper-like and contains only one condtional expression...:P), then where does "not needed" actually stop? E.g.:

```c++
TestOutput* CommandLineTestRunner::createJUnitOutput(const SimpleString& packageName)
{
     JUnitTestOutput* junitOutput = new JUnitTestOutput;
     if (junitOutput != NULL) {
         junitOutput->setPackageName(packageName);
     }
     return junitOutput;
}
```
As opposed to createConsoleOutput(), which is tested indirectly 6 times, this code would no longer be tested at all.

basvodde:
Sure, but I bet you can test the createJUnitOutput separately. I didn't, as it was extracted code and decided not to bother.

arstrube:
Could you explain what you mean by testing createJUnitOutput "separately"? I am actually very curious.

BTW it does get exercised in at least one test run in the TravisCI, but coverage is not recorded in that run. Besides, that would by more of a system test, I guess.

basvodde:
I mean, you call the method and figure out it created "the right object". Because when testing that, it doesn't mean you have to re-test the JUnitOutput class. That causes a piece of code to be tested via many tests, and that has a risk of high maintainability cost.

arstrube:
Sounds good - lemme try whether I can figure out a way to do that.

About the other test -- I would be happy to take out the division-by-zero test and all of its prerequisites entirely, as it doesn't contribute  anything unique to testing CppUTest and rather seems to lead to some sort of odd compiler test :P  . What do you think?

basvodde:
Its merged already. As long as it ain't causing harm, I don't mind integrating tests like these.

arstrube:
So if division-by-zero trips us up again, I'll solve the problem by re moving the test :)

PR #631, opened 2015-05-03
Appveyor mingw - CMake

Andne:
Added support for running the MinGW builds on Appveyor using CMake.

Is there a desire to include autotools as well?  Based on other comments I think that AppVeyor includes msys and could like do it, if people think that it would be helpful.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ environment:
   - PlatformToolset: v100
 
 install:
-- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw }
+- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw -forcex86 }
 
 build_script:
 - ps: scripts\appveyor_ci_build.ps1
--- a/scripts/appveyor_ci_build.ps1
+++ b/scripts/appveyor_ci_build.ps1
@@ -118,12 +118,12 @@ if ($env:PlatformToolset -eq 'MinGW')
     Remove-PathFolder "C:\Program Files (x86)\Git\cmd"
 
     # Add mingw to the path
-    Add-PathFolder "C:\Tools\mingw64\bin"
+    Add-PathFolder "C:\Tools\mingw32\bin"
 
     Write-Host "Building with Path: $env:Path"
 
     Invoke-BuildCommand "cmake -G 'MinGW Makefiles' .." 'cpputest_build'
     Invoke-BuildCommand "mingw32-make all" 'cpputest_build'
 
-    Remove-PathFolder "C:\Tools\mingw64\bin"
+    Remove-PathFolder "C:\Tools\mingw32\bin"
 }
--- a/scripts/appveyor_ci_test.ps1
+++ b/scripts/appveyor_ci_test.ps1
@@ -43,6 +43,6 @@ if ($env:PlatformToolset -ne 'MinGW')
 }
 else
 {
-    Invoke-Tests('.\cpputest_build\CppUTestTests.exe')
-    Invoke-Tests('.\cpputest_build\CppUTestExtTests.exe')
+    Invoke-Tests('.\cpputest_build\tests\CppUTestTests.exe')
+    Invoke-Tests('.\cpputest_build\tests\CppUTestExt\CppUTestExtTests.exe')
 }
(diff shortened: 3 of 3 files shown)


arstrube:
Personally, I think one build is all we need, and since you got cmake to work, that fine.

Now we need to get MinGW with Posix threads...

Andne:
If I understand right, your code change was supposed to be to handle MinGW that does not support posix threads.  So wouldn't we want the current version of MinGW then?

I can see adding another build where MinGW does support Posix threads, but since there are versions in the wild that do not, I think we should keep the current build too.

Andne:
I wish there weren't so many versions of MinGW, turns out the version on my computer (I use a build of MinGW-w64) is not that same as the version on AppVeyor (I think they use the original MinGW).

arstrube:
Sorry for posting on the other thread...

arstrube:
No, the posix thread issue is a separate one. (Obviously, @bortsov did not have that issue, or he would never have seen the bug he pointed out in my code). My change related to not handling fork() (and no version of MinGW can handle that, as far as I know). MinGW versions have become very confusing. Personally, I don't like putting yet another set of conditional code in the GCC/UtestPlatform.cpp. I would rather have to point out to the odd user (not many, as far as I can tell) that they need to install a Posix thread version of MinGW. There is also the option of installing PThread into MinGW separately.

As for AppVeyor, can you try the Chocolatey distro of MinGW, perhaps? I think that would be easier than installing PTthread libraries separately.

Andne:
I'm working on it right now.  Looks like we're going to get GCC 4.8.3.  Hopefully appveyor has more bandwidth than I do, the download is a bit slow.

Andne:
Ok, so the todo list from the other issue:
1. Get a version of MinGW that has Posix threads (via Chocolatey?)
2. Install that into, say C:\MinGW493_32.
3. Take out C:\MinGW from path and put in C:\MinGW493_32
4. Run the build normally

Here are my results:
1. Is in theory done (pthreads.h exists, but the #ifdef __MINGW32__ isn't working now)
2. It gets installed to C:\Tools\mingw64 (that's what --forcex86 was for, to get the 32-bit version instead)
3. Done
4. Build runs with new compiler, but fails because it's not detecting mingw anymore.  Seems like maybe there's a check that should be added to CMake/autotools instead of relying on a compiler default define.

arstrube:
Oh, it's working alright.

 I really wish it were 32 bit though. It's complaining about a long long to int conversion in the mingw stub.... ball's back in my court now. I guess with win 64 bit we can expect all the nasties that's theoretically possible, so presumably that's a good thing...

Will clean up the mess tomorrow. Well done on your part!!!

arstrube:
p.s. I don't do 64 bit myself on Windows. I don't like 32bit long / 64 bit pointer much (you got a demo why) and @basvodde  is opposed to long long (mind you, can't blame him).

Andne:
I agree, 64bit is going to cause some trouble.  I'm looking a bit to see how you are supposed to force 32bit in chocolately, I'm pretty sure I can, but there's a version mismatch between my computer and appveyor that changed some of the command-line options.

arstrube:
Not to worry. In this particular case, I should use pid_t (the int return type is an error no my part)

Andne:
I got the 32-bit build working, just need to clean up the test logic

arstrube:
Actually, maybe we should keep the 64 bit one. This kind of thing has tripped us up in the past. It might be good to have a nasty build like that.... none of the Travis build would pick up that sort of thing, although the VC builds should.

basvodde:
Hum hum. I guess all this got merged as part of the other pull request? Not quite sure why this one is open still? Should it be closed?

basvodde:
Oh, on a very early question. It would be great if we could include autotools :)

Andne:
There are a few late changes I made here that weren't in the other request.  I've rebased it, though that seems to have made the commit list look funny.  Hopefully it'll still merge correctly.

arstrube:
Err, it probably wasn't quite finished. Mine was just for testing my stuff based on @Andne's new work ;-). It did pass, but he hadn't incorporated running the correct test executables yet.

basvodde:
Eh, it passed, so I merged it. But now @Andne added to it, so it is ok now, right?

arstrube:
No problem. It's okay.

PR #632, opened 2015-05-03
Test whether the stubs work right with MinGW64

arstrube:


arstrube:
The Appveyor build shows up as successful, but it seems Appveyor still had two problems:
1. Related to cmake
2. Somehow it seemed like it couldn't find the executable(s)

arstrube:
@Andne - it builds CppUTestTests.exe and CppUTestExtTests.exe, but it attempts to run AllTests.exe. I thought our latest cmake build handles this itself via ctest?

Andne:
The first one is just the results of the cmake generation, I haven't figured out how to keep it from showing up as an error yet, but it doesn't interfer with the build, so I've ignored it.

Yeah, I don't know how ctest does/doesn't work, so for now I was just trying to tie into the test logic I already had running for the Visual C++ builds.

arstrube:
Ah, I see. Well on Travis, ctest is part of running cmake. So maybe we can glean some information from there. Otherwise, simply running the executables would serve just fine. At least it seems I got my error fixed.

Great stuff -- this is really helpful.

Andne:
Part of the reason I was avoiding CTest was since I don't know how to make its results show up on the tests tab.  I have that working for Visual C++ and thought it would be nice to have for MinGW as well.  I don't know if anyone actually pays attention to it, but it seems like a nice feature.

PR #639, opened 2015-05-04
Appveyor multiplatform

Andne:
I've added support to appveyor so that it runs MinGW as both a 32-bit and 64-bit build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,11 +6,14 @@ os:
 environment:
   matrix:
   - PlatformToolset: MinGW
+    Platform: Win32
+  - PlatformToolset: MinGW
+    Platform: x64
   - PlatformToolset: v90
   - PlatformToolset: v100
 
 install:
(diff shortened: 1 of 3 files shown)


arstrube:
Great. Could you perhaps in the lack of failure output? It registers the failure, but does not seem to ouptput the failure string.

arstrube:
As can be seen here: https://ci.appveyor.com/project/arstrube/cpputest/build/3.7.0-ci8/job/msyphrqx8oeo9kpl

Andne:
If you go to the tests tab, it shows you the full error message there.  Do you want the output in the console too?  That is possible I'm sure, I just liked having the test information stay on it's own page.  However, I should probably make the message page show the details as well.

arstrube:
Oh, kewl. No, I didn't know that. In that case, that's much more convenient than console output.

If you still consider it a good idea, maybe we could run the tests with -v -junit. But, as for me, that's not necessary. I just didn't know my way around AppVeyor well enough.

PR #640, opened 2015-05-04
This test demonstrates lost groups and names

arstrube:
when listing test groups / names. This is far more likely to
happen with groups than with names.


arstrube:
No longer needed

PR #641, opened 2015-05-04
Added proper pthread library linking when using CMake and MinGW to compile

jgonzalezdr:
It seems that when compiling using MinGW, the pthread library has to be linked explicitly, so I've modified the CMakeLists to do so when using that compiler.

NOTE: cmake's find_package(Threads) is buggy, at least at version 2.8.12, as ${CMAKE_THREAD_LIBS_INIT} is not set properly (it's actually empty), therefore this variable is not used to add the library to the linker.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -38,8 +38,13 @@ if (MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4723")
 endif (MSVC)
 
+if (MINGW)
+    find_package (Threads REQUIRED)
+    set(THREAD_LIB "pthread")
+endif (MINGW)
+
 add_executable(CppUTestTests ${CppUTestTests_src})
-target_link_libraries(CppUTestTests CppUTest)
+target_link_libraries(CppUTestTests CppUTest ${THREAD_LIB})
 
 if (TESTS)
     if (TESTS_DETAILED)
--- a/tests/CppUTestExt/CMakeLists.txt
+++ b/tests/CppUTestExt/CMakeLists.txt
@@ -18,8 +18,13 @@ set(CppUTestExtTests_src
     OrderedTestTest.cpp
 )
 
+if (MINGW)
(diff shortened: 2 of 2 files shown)


arstrube:
Interesting. We didn't run into that issue when doing the AppVeyor build, however, we did have the issue of first using a version of MinGW that had no support for PThread (when that happens, the missing header causes a compiler error).

Just for interest, could it be that you installed PThread seprately from and after MinGW? That might explain why the library needs to be linked explicitly. Either way, it seems your addition would be helpful for others in the same situation.

Andne:
@jgonzalezdr Can you tell us which version/configuration of MinGW you're using?  There's a few different versions around and it would be helpful to have a better understanding of how each of them interact with the build tools and code.

basvodde:
Looks good though. Merge?

arstrube:
yup

jgonzalezdr:
The version of MinGW that I'm using is the latest one installed using the official installer (http://www.mingw.org/download/installer), which is basically mingw32-gcc v4.8.1-4, mingw32-gcc-g++ v4.8.1-4, and mingw32-libpthreadgc v2.9.1-1. I suppose this is one of the most widespread installations out there.

The pthread library is installed separately.

I expect that this patch doesn't collide with other MinGW versions or installation packages, since MinGW linker seems to silently ignore "-lpthread" if the library is embedded in standard libraries (or always linked btw), and the AppVeyor build using MinGW-w64 was successful.

Anyway, I pledge to fix the patch if a problem arises with a particular version of MinGW or cmake.

basvodde:
@jgonzalezdr Thanks, merged already :)

arstrube:
We use the following builds on Appveyor:

http://downloads.sourceforge.net/mingw-w64/x86_64-4.8.3-release-posix-seh-rt_v3-rev0.7z
http://downloads.sourceforge.net/mingw-w64/i686-4.8.3-release-posix-dwarf-rt_v3-rev0.7z

Neither seems to need the patch, but neither had any problems with it, either.

So I think you'll be fine.

PR #642, opened 2015-05-04
Fix test groups listing feature

mmidor:
* TestRegistry::listTestGroupNames() put each of grup in #name# before
  putting into temporary list. This will prevent missing of GROUP1
  in the list when GROUP11 was already there
* TestRegistry::listTestGroupAndCaseNames() - just like above

### DIFF
--- a/tests/TestRegistryTest.cpp
+++ b/tests/TestRegistryTest.cpp
@@ -114,6 +114,7 @@ TEST_GROUP(TestRegistry)
     MockTest* test1;
     MockTest* test2;
     MockTest* test3;
+    MockTest* test4;
     TestResult *result;
     MockTestResult *mockResult;
     void setup()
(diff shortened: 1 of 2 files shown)


tests/TestRegistryTest.cpp
```diff
@@ -326,16 +329,20 @@ TEST(TestRegistry, ResetPluginsWorks)
 
 TEST(TestRegistry, listGroupNames)
 {
+    // run tests backward, ensure that GROUP_1 will be listed
+    // after GROUP_11, and also that GROUP_2 will be listed only once
```
basvodde: Would you mind adding this text to the name of the test instead of adding it ad a comment ;)

arstrube:
Something along the lines of `shouldListGroupNamesWhereOneNameIsContainedInAnother` :D

mmidor:
Please, do it yourself as I will be unavailable for the next 12 hours.
I'm not sure if you may need write access to my repo in such a case?

arstrube:
No problem. I will do it for you. No need for write access.

PR #644, opened 2015-05-05
Just had to get those 100% back...

arstrube:
cov_2015-05-05
cov_2015-05-05

### DIFF
--- a/tests/UtestTest.cpp
+++ b/tests/UtestTest.cpp
@@ -30,6 +30,15 @@
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTest/PlatformSpecificFunctions.h"
 
+TEST_GROUP(Utest)
+{
+};
+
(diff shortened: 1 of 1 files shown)


basvodde:
Ahaha :)

But isn't this covered via the other test anyway? It does a division where it crashes. Why doesn't that cover it?

arstrube:
Ah, good point. This is true of all methods which get covered in a separate process. The separate process kind of creates a race condition. It wants to write to the same .gcda (or even does) but then it gets overwritten by the parent process. 

If we had a record of its being called, the we'd only need to use `} // LCOV_IGNORE_LINE`.

We could of course ignore the whole function :)

arstrube:
p.s. 1 - that's why UtestPlatformTest.cpp has a pretty lousy coverage
p.s.2 - calling a branch 100% is not a good idea; AppVeyor conked out over it

PR #647, opened 2015-05-09
Remove division-by-zero tests again because

arstrube:
of too many problems with different compilers / architectures.


basvodde:
I feel sad about removing it. Just now we knew how to cheat the compiler :)

arstrube:
No problem. I started a file I call XtremeTest.cpp. I'll keep this test case in there plus some others that you never accepted :-)

I imagine one day we might start a test suite that only gets run on demand, by interested persons, on interesting compilers and architectures. Let me know if you are interested...

PR #648, opened 2015-05-10
Change test so as to only look at the type of JUnitOutput

arstrube:
@basvodde - I know we don't use dynamic_cast<> :P What I don't know is how check for the type of what's returned by getTestOutput() without dynamic_cast<>.

I couldn't find a single method of JUnitTestOutput that doesn't either want to write to a file, or does nothing at all. Subclassing didn't seem to help, because I couldn't find an injection point.

Ideas?

### DIFF
--- a/tests/CommandLineTestRunnerTest.cpp
+++ b/tests/CommandLineTestRunnerTest.cpp
@@ -161,70 +161,6 @@ TEST(CommandLineTestRunner, listTestGroupAndCaseNamesShouldWorkProperly)
     STRCMP_CONTAINS("group.test", commandLineTestRunner.fakeConsoleOutputWhichIsReallyABuffer->getOutput().asCharString());
 }
 
-# if 0
-
-struct FakeOutput
-{
-    FakeOutput() : SaveFOpen(PlatformSpecificFOpen), SaveFPuts(PlatformSpecificFPuts),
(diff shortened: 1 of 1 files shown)


basvodde:
Nope, but we can close this as we won't do the dynamic_cast :)

Andne:
I know it's not ideal, but any thoughts on adding a function that just returns an output type description?  For now it'd probably be only there for the unit test, but there's a couple things that might be able to make use of it in the future.

arstrube:
@basvodde hey wait, that's not how this is supposed to work!
The dynamic_cast is there because I didn't know any better, but you were supposed to have some trick up your sleeve, not close it :P

arstrube:
@basevodde said:
>Sure, but I bet you can test the createJUnitOutput separately
Your bet is still up -- I'm keen to learn...

basvodde:
Ehe :) There is a reason why I hadn't created a test purely around createJUnitOutput :)

I wouldn't go for the dynamic_cast. That would break the CppUTest design rules and is ugly also. So therefore you'll need to test for the behavior, which is what you did. I'd try to find the smallest possible behavior I could find as it is likely to duplicate earlier tests.

Thinking about it. The easiest way to test is is probably to use the createJUnitOutput in the JUnitOutputTest to create the JUnitOutput :) Not test is here.

PR #657, opened 2015-05-28
BITS_EQUAL assertion

KisImre:
I've created an assertion for checking the equality of the specified bits in a value. Here's an example of it's usage:
```
TEST(BitsDemo, Demo) {
    uint16_t expected = 0x1234;
    uint16_t actual = 0xFF74;
    uint16_t mask = 0x00F8;

    BITS_EQUAL(expected, actual, mask)
}
```
The output:
```
TestBits.cpp:10: error: Failure in TEST(BitsDemo, Demo)
	expected <xxxxxxxx 00110xxx>
	but was  <xxxxxxxx 01110xxx>
```

I used the sizeof operator on the actual value for determine the number of bits that the error message should display. I don't know if it's good or it's not. I think it's not so nice when I'm checking a variable with the length of 8 bits then I get an error message containing 64 bits. There can be another solution for this problem, if there are separate test macros for the integers with different length.


basvodde:
Very nice, thanks. Much appreciated.

PR #658, opened 2015-05-29
Fixed eclipse project generation using cmake

jgonzalezdr:
Added headers to cmake's library files to obtain proper projects with proper headers when creating an eclipse project.

Without this patch, when creating an eclipse project using cmake (e.g. cmake -G "Eclipse CDT4 - MinGW Makefiles"), the header files don't appear in the target's headers, which makes indexing not working properly, and finding them a bit harder.


basvodde:
Closed?

jgonzalezdr:
I reopened in a different PR because I though that the PR was linked to the commit, not the branch, and have messed up things with something that will go on another PR when it's finished...

Issue #661, opened 2015-05-30
Build fails with GCC 5


offa:
Building with ***GCC 5.1.0 (x86-64)*** fails with two errors:
```
MockExpectedCallTest.cpp:531:46: Format %p expects argument of type void, but argument 2 has type »MockExpectedCallComposite*« [-Werror=format=]

     info += StringFromFormat("%p", &composite);
                                              ^
```

and
```
Unknown commandlineoption »-Wno-c++98-compat« [-Werror]
```

*(Both translated from my system language)*

Tested with current git version and v3.7.2 (using `cmake -DC++11=ON ..`)

With GCC 4.9.x everything was ok.

offa:
Seems the root cause is the first error, I'll open a PR with a fix.

offa:
Autotools will need some more years, here it is ...

basvodde:
Should the second one be fixed too still? We should be able to do that within the configure.ac

offa:
> Should the second one be fixed too still?

I don't think so, it looks like the error was caused by the type mismatch only. With the cast added the second error is gone.

basvodde:
Ok, strange. Anyways, your other PR is merged.

offa:
Thanks, issue is solved then.

PR #662, opened 2015-05-30
Fixed build with GCC 5 (issue  #661)

offa:
This is a fix for #661.


tests/CppUTestExt/MockExpectedCallTest.cpp
```diff
@@ -528,7 +528,7 @@ TEST(MockExpectedCallComposite, isOnObject)
 {
     composite.onObject(&composite);
     SimpleString info("(object address: ");
-    info += StringFromFormat("%p", &composite);
+    info += StringFromFormat("%p", static_cast<void*>(&composite));
```
basvodde: Could you change this to a C style cast? Within CppUTest we prefer not using the "new" style C++ casts.

offa:
@basvodde cast is replaced.

PR #668, opened 2015-06-09
Fix UtestPlatform.cpp plus tests for IAR

arstrube:
@imr I hope the #define for IAR (`__IAR_SYSTEMS_ICC__`) hits the mark.

Please, let us know when this compiles and tests correctly on your system!

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -57,7 +57,7 @@ static int jmp_buf_index = 0;
 
 static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, TestPlugin*, TestResult* result)
 {
-    result->addFailure(TestFailure(shell, "-p doesn't work on MinGW as it is lacking fork.\b"));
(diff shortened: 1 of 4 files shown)


arstrube:
@basvodde - should we supply empty dummy implementations for the Mutex stuff (as in this PR), or would it be better to have it bark loudly (as before), so new users __know__ the Mutex stuff is not implemented? 

@imr Would a 'real' implementation of some sort for the Cortex M3 be helpful?

imr:
@arstrube It might be helpful, but is not necessary now. I think one approach might be through CMSIS, which would cover the entire Cortex M line.

basvodde:
@arstrube Either way. I usually prefer FAILing, but either way is ok.

arstrube:
@basvodde me, too. But it's merged already ;-). I meant to roll it back to the previous commit if we agreed on FAIL(). The one ugly thing about FAIL is that it occurs in main(), so maybe doing nothing is actually best.....

basvodde:
If you feel a need to add the FAIL, please do. I'll merge it as I have no really strong opinion on this one.

PR #673, opened 2015-06-21
First shot at Custom Copiers

arstrube:
I tried to go for the solution with the fewest changes. There were some tradeoffs:
+ A new MockCopier class vs. a new method in NamedValueComparator. I chose the latter.
+ Reusing some of the original output parameter test vs. creating a totally new one with its own object for copying. I reused the existing one.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -40,7 +40,7 @@ public:
 
     virtual bool isEqual(const void* object1, const void* object2)=0;
     virtual SimpleString valueToString(const void* object)=0;
-    virtual void* copy(const void*, const void*) { return 0; }
+    virtual bool copy(const void*, const void*) { return false; }
 };
 
 class MockFunctionComparator : public MockNamedValueComparator
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -40,6 +40,7 @@ public:
 
     virtual bool isEqual(const void* object1, const void* object2)=0;
     virtual SimpleString valueToString(const void* object)=0;
+    virtual void* copy(const void*, const void*) { return 0; }
 };
 
 class MockFunctionComparator : public MockNamedValueComparator
--- a/src/CppUTestExt/MockExpectedCall.cpp
+++ b/src/CppUTestExt/MockExpectedCall.cpp
@@ -154,6 +154,14 @@ MockExpectedCall& MockCheckedExpectedCall::withOutputParameterReturning(const Si
     return *this;
 }
 
+MockExpectedCall& MockCheckedExpectedCall::withOutputParameterOfTypeReturning(const SimpleString& type, const SimpleString& name, const void* value)
+{
+    MockNamedValue* newParameter = new MockExpectedFunctionParameter(name);
+    outputParameters_->add(newParameter);
+    newParameter->setObjectPointer(type, value);
+    return *this;
+}
+
 SimpleString MockCheckedExpectedCall::getInputParameterType(const SimpleString& name)
 {
     MockNamedValue * p = inputParameters_->getValueByName(name);
@@ -562,6 +570,13 @@ MockExpectedCall& MockExpectedCallComposite::withOutputParameterReturning(const 
 {
     for (MockExpectedCallCompositeNode* node = head_; node != NULL; node = node->next_)
         node->call_.withOutputParameterReturning(name, value, size);
+    return *this;
+}
+
+MockExpectedCall& MockExpectedCallComposite::withOutputParameterOfTypeReturning(const SimpleString& type, const SimpleString& name, const void* value)
+{
+    for (MockExpectedCallCompositeNode* node = head_; node != NULL; node = node->next_)
+        node->call_.withOutputParameterOfTypeReturning(type, name, value);
     return *this;
 }
 
--- a/tests/CppUTestExt/MockSupportTest.cpp
+++ b/tests/CppUTestExt/MockSupportTest.cpp
@@ -645,8 +645,22 @@ TEST(MockSupportTest, threeExpectedAndActual)
 class MyTypeForTesting
 {
 public:
-    MyTypeForTesting (int val) : value(new int(val)) {}
-    ~MyTypeForTesting() { delete value; }
+    MyTypeForTesting () : value(new(int)) {}
+    MyTypeForTesting(int val) : value(new int(val)){}
+    MyTypeForTesting(MyTypeForTesting& rhs)
+    {
+        value = new(int);
+        *value = *rhs.value;
+    }
+    ~MyTypeForTesting()
+    {
+        delete value;
+    }
+    MyTypeForTesting& operator=(MyTypeForTesting& rhs)
+    {
+        *value = *rhs.value;
+        return *this;
+    }
     int * value;
 };
 
@@ -660,6 +674,12 @@ public:
     virtual SimpleString valueToString(const void* object)
     {
         return StringFrom(*(((MyTypeForTesting*)object)->value));
+    }
+    virtual void* copy(const void* object1, const void* object2)
+    {
+        MyTypeForTesting* ret = (MyTypeForTesting*)object1;
+        *ret = *((MyTypeForTesting*)object2);
+        return (void*)ret;
     }
 };
 
@@ -702,8 +722,9 @@ TEST(MockSupportTest, customObjectOutputParameterShouldSucceed)
     MyTypeForTesting source(1);
     MyTypeForTesting target(2);
     MyTypeForTestingComparator comparator;
-    mock().expectOneCall("function").withOutputParameterReturning("parameterName", &source, sizeof(MyTypeForTesting));
-    mock().actualCall("function").withOutputParameter("parameterName", &target);
+    mock().installComparator("MyTypeForTesting", comparator);
+    mock().expectOneCall("function").withOutputParameterOfTypeReturning("MyTypeForTesting", "parameterName", &source);
+    mock().actualCall("function").withOutputParameterOfType("MyTypeForTesting", "parameterName", &target);
     mock().checkExpectations();
     CHECK_NO_MOCK_FAILURE();
     LONGS_EQUAL(1, *source.value);
--- a/tests/CppUTestExt/MockSupportTest.cpp
+++ b/tests/CppUTestExt/MockSupportTest.cpp
@@ -645,8 +645,9 @@ TEST(MockSupportTest, threeExpectedAndActual)
 class MyTypeForTesting
 {
 public:
-    MyTypeForTesting(int val) : value(val){}
-    int value;
+    MyTypeForTesting (int val) : value(new int(val)) {}
+    ~MyTypeForTesting() { delete value; }
+    int * value;
 };
 
 class MyTypeForTestingComparator : public MockNamedValueComparator
@@ -654,14 +655,13 @@ class MyTypeForTestingComparator : public MockNamedValueComparator
 public:
     virtual bool isEqual(const void* object1, const void* object2)
     {
-        return ((MyTypeForTesting*)object1)->value == ((MyTypeForTesting*)object2)->value;
+        return *((MyTypeForTesting*)object1)->value == *((MyTypeForTesting*)object2)->value;
     }
     virtual SimpleString valueToString(const void* object)
     {
-        return StringFrom(((MyTypeForTesting*)object)->value);
+        return StringFrom(*(((MyTypeForTesting*)object)->value));
     }
 };
-
 
 TEST(MockSupportTest, customObjectParameterFailsWhenNotHavingAComparisonRepository)
 {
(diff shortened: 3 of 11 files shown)


src/CppUTestExt/MockExpectedCall.cpp
```diff
@@ -154,6 +154,15 @@ MockExpectedCall& MockCheckedExpectedCall::withOutputParameterReturning(const Si
     return *this;
 }
 
+MockExpectedCall& MockCheckedExpectedCall::withOutputParameterOfTypeReturning(const SimpleString& type, const SimpleString& name, const void* value)
+{
+    MockNamedValue* newParameter = new MockExpectedFunctionParameter(name);
+    outputParameters_->add(newParameter);
+    newParameter->setObjectPointer(type, value);
+    newParameter->setValue(value);
```
arstrube: I think this line should go.

tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -645,24 +645,44 @@ TEST(MockSupportTest, threeExpectedAndActual)
 class MyTypeForTesting
 {
 public:
-    MyTypeForTesting(int val) : value(val){}
-    int value;
+    MyTypeForTesting () : value(new(int)) {}
```
arstrube: This is now a total overkill for the existing test case....

tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -697,6 +717,21 @@ TEST(MockSupportTest, outputParameterSucceeds)
     CHECK_NO_MOCK_FAILURE();
 }
 
+TEST(MockSupportTest, customObjectOutputParameterSucceeds)
```
arstrube: There are two possible error cases that should be tested:
1. Comparator of wrong type. This does produce an error, but it should really get its own failure class.
2. No comparator found. This should produce an error, but can't, because we need to do memcpy in this case (might have been called via withOutputParameterReturning().
Because the "expected" results wouldn't be right in either case, I didn't implement either test case yet. Might use IGNORE_TEST() though, for now.

arstrube: The additional test cases are no longer  needed when we just add a `copy()` method to the MockNamedValueComparator class.

src/CppUTestExt/MockExpectedCall.cpp
arstrube: It was an error indeed.

arstrube:
Cleaned up the commits to show more clearly what was done:
1. Change MyTypeForTesting so it allocates some memory and uses a pointer
2. Add a test for output param that uses MyTypeForTesting
3. Add withOutputParameterOfType() method interface
4. Add Implementation (along with copy() method for MockNamedValueComparator)
5. Change test case, so it uses the new methods and succeeds

random85:
I tried to compile your branch  and slightly changed my code as follow:
```diff
Index: mocks/SerializedMessageQueueMock.cpp
===================================================================
--- mocks/SerializedMessageQueueMock.cpp	(revision 12641)
+++ mocks/SerializedMessageQueueMock.cpp	(working copy)
@@ -96,11 +96,8 @@
 int8_t SerializedMessageQueueMock::timedReadIpcMessage(data_server::IpcMessage& msg,
         uint32_t timeout)
 {
-    if(m_pIpcMsg != NULL)
-    {
-        msg = *m_pIpcMsg;
-    }
     return mock().actualCall("timedReadIpcMessage").onObject(this).
+            withOutputParameterOfType("IpcMessage", "msg", &msg).
             returnIntValue();
 }
 
Index: tests/DataPoolClientMQueueTest.cpp
===================================================================
--- tests/DataPoolClientMQueueTest.cpp	(revision 12647)
+++ tests/DataPoolClientMQueueTest.cpp	(working copy)
@@ -703,8 +703,8 @@
             withParameterOfType("IpcMessage", "msg", &connectMsg).
             withParameterOfType("IpcPriority", "prio", &prio).
             andReturnValue(true);
-    cltMqueue->addMessageToRead(&hbMsg);
     mock().expectOneCall("timedReadIpcMessage").onObject(cltMqueue).
+            withOutputParameterOfTypeReturning("IpcMessage", "msg", &hbMsg).
             andReturnValue(0);
     mock().expectOneCall("close").onObject(cltMqueue).
             withParameter("unlink", true);
Index: tests/utils/Comparators.h
===================================================================
--- tests/utils/Comparators.h	(revision 12589)
+++ tests/utils/Comparators.h	(working copy)
@@ -216,6 +216,14 @@
         return true;
     }
 
+    virtual void* copy(const void* object1, const void* object2)
+    {
+        data_server::IpcMessage* msg = (data_server::IpcMessage*)object1;
+        *msg = *((data_server::IpcMessage*)object2);
+
+        return (void*)msg;
+    }
+
     virtual SimpleString valueToString(const void* object)
     {
         std::string ret;
```
Unfortunately I have a segmentation fault probably not related to the custom copiers feature. It crashes in a test where I don't event use custom copiers neither any of the _withOutput*_ functions family:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000446123 in StrLen (str=0xed <error: Cannot access memory at address 0xed>)
    at src/CppUTest/SimpleString.cpp:91
91	    do n++; while (*str++);
(gdb) bt 
#0  0x0000000000446123 in StrLen (str=0xed <error: Cannot access memory at address 0xed>)
    at src/CppUTest/SimpleString.cpp:91
#1  SimpleString::copyToNewBuffer (bufferToCopy=0xed <error: Cannot access memory at address 0xed>, 
    bufferSize=bufferSize@entry=0) at src/CppUTest/SimpleString.cpp:404
#2  0x00000000004463d1 in SimpleString::SimpleString (this=0x1374030, other=...) at src/CppUTest/SimpleString.cpp:166
#3  0x0000000000456afe in MockNamedValue (this=0x1374020) at ./include/CppUTestExt/MockNamedValue.h:70
#4  MockCheckedExpectedCall::returnValue (this=0x1373f50) at src/CppUTestExt/MockExpectedCall.cpp:434
#5  0x0000000000435bb7 in TEST_SerializedMessageQueue_timedReadEmptyDoesntFail_Test::testBody (this=0x1373e80)
    at /home/moragu/projects/DataServer-cpp/IFC/apps/data-server/build/../tests/SerializedMessageQueueTest.cpp:246
#6  0x000000000044d7b1 in PlatformSpecificSetJmpImplementation (function=0x44b5a0 <helperDoTestBody(void*)>, 
    data=0x1373e80) at src/Platforms/Gcc/UtestPlatform.cpp:144
#7  0x000000000044ba8e in Utest::run (this=0x1373e80) at src/CppUTest/Utest.cpp:523
#8  0x000000000044d19f in UtestShell::runOneTestInCurrentProcess (
    this=0x69d7a0 <TEST_SerializedMessageQueue_timedReadEmptyDoesntFail_TestShell_instance>, plugin=0x7fff6a9799f0, 
    result=...) at src/CppUTest/Utest.cpp:203
#9  0x000000000044d7b1 in PlatformSpecificSetJmpImplementation (
    function=0x44b5c0 <helperDoRunOneTestInCurrentProcess(void*)>, data=0x7fff6a9798e0)
    at src/Platforms/Gcc/UtestPlatform.cpp:144
#10 0x000000000044b64d in UtestShell::runOneTest (this=<optimized out>, plugin=<optimized out>, result=...)
    at src/CppUTest/Utest.cpp:177
#11 0x000000000044b109 in TestRegistry::runAllTests (this=0x69dc60 <TestRegistry::getCurrentRegistry()::registry>, 
    result=...) at src/CppUTest/TestRegistry.cpp:62
#12 0x00000000004430a8 in CommandLineTestRunner::runAllTests (this=this@entry=0x7fff6a979a60)
    at src/CppUTest/CommandLineTestRunner.cpp:119
#13 0x000000000044318d in CommandLineTestRunner::runAllTestsMain (this=this@entry=0x7fff6a979a60)
    at src/CppUTest/CommandLineTestRunner.cpp:80
#14 0x00000000004432aa in CommandLineTestRunner::RunAllTests (ac=2, av=0x7fff6a979be8)
    at src/CppUTest/CommandLineTestRunner.cpp:50
#15 0x0000000000431f1b in main (ac=2, av=0x7fff6a979be8)
    at /home/moragu/projects/DataServer-cpp/IFC/apps/data-server/build/../tests/main.cpp:35
```
Did I do something wrong? Are you messing with SimpleString as well?

arstrube:
Hi,
No, I can't see that you did something wrong, and No, I am not messing with Simplestring :D.
There might be a problem with MockNamedValue, though that seems unlikely, since all of the existing tests passed on Travis CI.
Have you been using the Head off Github master before, or a specific version like 3.7.1?

arstrube:
It would be helpful to see the code of 
```c++
TEST(SerializedMessageQueue,timedReadEmptyDoesntFail)
```
or, better even, a very simple generalized test, that produces the same effect. SimpleString is used in a lot of places; it seems that here, we are working with a return value (which is of type MockNamedValue, which might have been affected by my changes).

include/CppUTestExt/MockNamedValue.h
```diff
@@ -40,6 +40,7 @@ class MockNamedValueComparator
 
     virtual bool isEqual(const void* object1, const void* object2)=0;
     virtual SimpleString valueToString(const void* object)=0;
+    virtual void* copy(const void*, const void*) { return 0; }
```
jgonzalezdr: Why return anything at all if the returned value is not used? 

`void` return value could be a better choice, or maybe return `bool` and check in `MockCheckedActualCall::finalizeOutputParameters` that the value is not `false`(and throw an exception / error otherwise), such that the default implementation returns `false` and overriden implementatations should return `true`. That way if a user installs a comparator without overriding `copy` and it's used to copy an output parameter, it doesn't fail silently (which can be hard to detect).

jgonzalezdr:
@arstrube Nice implementation. 

BTW, it's a bit disturbing (at least for me) that a "comparator" can be used to "copy" things around, so maybe MockNamedValueComparator could be renamed to something more suitable like MockNamedValueMatcher (subtle nuance here... hehe). What do you think?

basvodde:
Didn't check the implementation yet, but I do suggest not to mix comparators with copiers :) I assume they are rarely used together.

random85:
@arstrube Sorry, I was too enthousiastic and wanted to try on my project. I understand your point and running the example discussed in the bug worked sucessfully and valgrind reported no leak as well!

include/CppUTestExt/MockNamedValue.h
arstrube: Yes, that sounds like a very good idea, especially returning a boolean and checking that the copy() method was indeed overridden. @random85 - what was the original idea behind returning a void *?

arstrube:
@jgonzalezdr I totally agree. I would have liked to call it "MockNamedValueHandler" but I don't think that's an option, because renaming it would break existing user code.

@basvodde I mostly agree about not mixing the two. However, my reasoning is that
1. If I have a custom object, I need to "handle" it, which involves a number of things, like turning it into a sting, comparing it, and copying it.
2. I don't feel strongly about 1. but I do shy away from duplicating all this housekeeping code, both for handling another type of "custom handler" and for the mocking classes: nodes, lists, default lists, etc. in order to handle a different kind of handler. If there's any good idea out there how to handle that side of things, I'd be open to consider it. A common base class perhaps?

@random85, no problem, it was supposed to work! Is the (perhaps unrelated) problem still present? Is it possible for me to reproduce it somehow?

arstrube:
@basvodde - part of my reasoning also was that it is a hassle, for each custom object, to have to install a comparator, a copier, a ... you name it. It seems to me more sensible to install one handler for all of these. But you don't always have to define all methods, in that's you're absolutely right. That's why copy() isn't pure virtual. In any case, it I could see a _simple_ and elegant way to do the housekeeping, I would definitely consider it.

jgonzalezdr:
It might be simpler (from the POV of a user) to just have a differentiated class `MockNamedValueCopier` and a separate registry for copiers, because in fact there is no need at all being forced to implement the `isEqual` method, isn't it?. 

But then of course if you want both a comparator and a copier, you have to put them in different classes, register them twice, etc.

At the end, it's just a matter of which use case is more common: Will comparators and copiers be used together for the same class frequently or rarely? I vote for rarely.

Regarding the renaming of `MockNamedValueComparator` without breaking existing code, you can just rename it, then add a typedef to the old name which can be deprecated in the future.

basvodde:
@arstrube I tend to agree with @jgonzalezdr arguments. I fear they are more often used separately than together, and hence... combining them wouldn't be good. I'll leave it to you, but if you do combine them, then the comparator should be renamed (without a typedef)

arstrube:
I created an alternative PR #675 for the other solution.

random85:
I find the discussion interesting and agree with @basvodde and @jgonzalezdr. As I write more CPP code, I would vote for using copiers and comparator for the same class happens more often. However, when we write C code, we will bump into this situation less often.

This is just a choice of design but my personnal preference would be using the same class and find a more appropriate name.

random85:
@arstrube Sorry for the late answer

**Is it possible for me to reproduce it somehow?**
Unfortunately I haven't any simple test case right now. This happens when I compile my current project with your *copiers* branch, even though I don't use the new code. Unfortunately, I haven't time right now to track the issue. As the simple example I have works, I think I will wait until your change are merged into the master branch and then I will perform further tests if you don't mind...

arstrube:
Personally, I don't have a strong preference. Funnily, just renaming this class would mean breaking more of existing user code. Any idea for a name? MockNamedValueHandler wasn't too well received ;-)

random85:
Yep, *handler* is a very tempting name :tongue: ! My imagination to find class name is also pretty limited ...  what about **MockNamedCustomObject**?

arstrube:
That would make for installCustomObject(name, comparator), installCustomObject(name, copier), removeAllCustomObjects() (thinking of the other PR). That would be a bit odd, I think.

For an all-in-one approach (e.g. this PR) it sounds good to me though.

arstrube:
Closing, as it would mean putting copying and comparing into the same class.

PR #675, opened 2015-06-24
Custom Copiers - Alternative Implementation

arstrube:
Okay -- here goes. This is the version that has separate copiers and comparators, aggregated into a MockNamedValueHandler class.

The nice thing is that existing user code need (almost) not be changed. The only thing that changed is that `mock().removeAllComparators()` became `mock().removeAllHandlers()`.

Also, a custom class can have both a Comparator and a Copier.

All in all, I like this implementation better myself.

BUT -- there's a downside. It means a lot of changes, even though I made sure to change as little as possible.

Suggestions for improvements welcome :)

### DIFF
--- a/examples/AllTests/EventDispatcherTest.cpp
+++ b/examples/AllTests/EventDispatcherTest.cpp
@@ -72,7 +72,7 @@ TEST_GROUP(EventDispatcher)
     void teardown()
     {
         delete dispatcher;
-        mock().removeAllComparators();
+        mock().removeAllHandlers();
     }
 };
 
--- a/examples/AllTests/EventDispatcherTest.cpp
+++ b/examples/AllTests/EventDispatcherTest.cpp
@@ -76,7 +76,6 @@ TEST_GROUP(EventDispatcher)
     }
 };
 
-
 TEST(EventDispatcher, EventWithoutRegistrationsResultsIntoNoCalls)
 {
     dispatcher->dispatchEvent(event, 10);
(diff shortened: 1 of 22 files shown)


basvodde:
Why Handlers and not Copiers?

Handlers has no meaning :)

arstrube:
Needed a box to pack both Handlers and Copiers into ;-).

Copiers are called Copiers, and comparators are called Comparators.

I know, I don't like `removeAllHandlers()` (the only place where the user 'sees' it) myself, but I would have to further dismantle existing code, since that method uses `clear()`, and you would have to delete copiers and comparators separately to be able to have `removeAllComparators` and `removeAllCopiers` instead...

basvodde:
Let me try to check it more next week.

arstrube:
Alright. By then the bugs should be ironed out :)

jgonzalezdr:
My 2 cents: If you need to break the API because of "removeAllHandlers", then it would be better for homogeneity sake to go all the way and have two "installHandler" methods with different signatures rather than "installComparator" and "installCopier".

arstrube:
@jgonzalezdr - I'd prefer not to break the API. But I do like your idea; at least it is a lot more logical from a user's point of view.

arstrube:
@jgonzalezdr It has been done. Breaking the API even more -- but nice and consistent

arstrube:
This has been superseded by #676, thank you @jgonzalezdr.

PR #676, opened 2015-06-25
Custom Copiers - Third contender

jgonzalezdr:
Inspired by the works of @arstrube, I implemented a version that doesn't break existing code, keeping comparators and copiers as separate things:

Added a new class MockNamedValueCopier homologous to
MockNamedValueComparator, but which copies an object of a non-native
data type into another (instead of comparing them).

The comparator repositories have been renamed to "handlers"
repositories, and they can now store both comparators and copiers.

Finally, also added the methods "withOutputParameterOfTypeReturning" to
MockActualCall and "withOutputParameterOfType" to MockActualCall, to
declare output parameters of non-native type, which will try to use
copiers from the repository to copy the expected parameter object into
the actual parameter object.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -28,7 +28,7 @@
 #ifndef D_MockNamedValue_h
 #define D_MockNamedValue_h
 /*
- * MockParameterComparator is an interface that needs to be used when creating Comparators.
+ * MockNamedValueComparator is an interface that needs to be used when creating Comparators.
  * This is needed when comparing values of non-native type.
  */
 
@@ -60,13 +60,41 @@ private:
 };
 
 /*
+ * MockNamedValueCopier is an interface that needs to be used when creating Copiers.
+ * This is needed when copying values of non-native type.
+ */
+
+class MockNamedValueCopier
+{
+public:
+    MockNamedValueCopier() {}
+    virtual ~MockNamedValueCopier() {}
+
+    virtual void copy(void* dst, const void* src)=0;
+};
+
+class MockFunctionCopier : public MockNamedValueCopier
+{
+public:
+    typedef void (*copyFunction)(void*, const void*);
+
+    MockFunctionCopier(copyFunction copier) : copier_(copier) {}
+    virtual ~MockFunctionCopier(){}
+
+    virtual void copy(void* dst, const void* src) _override { return copier_(dst, src); }
+
+private:
+    copyFunction copier_;
+};
+
+/*
  * MockNamedValue is the generic value class used. It encapsulates basic types and can use them "as if one"
  * Also it enables other types by putting object pointers. They can be compared with comparators.
  *
  * Basically this class ties together a Name, a Value, a Type, and a Comparator
  */
 
-class MockNamedValueComparatorRepository;
+class MockNamedValueHandlerRepository;
 class MockNamedValue
 {
 public:
@@ -104,9 +132,11 @@ public:
     virtual const void* getConstPointerValue() const;
     virtual const void* getObjectPointer() const;
     virtual size_t getSize() const;
-    virtual MockNamedValueComparator* getComparator() const;
 
-    static void setDefaultComparatorRepository(MockNamedValueComparatorRepository* repository);
+    virtual MockNamedValueComparator* getComparator() const;
+    virtual MockNamedValueCopier* getCopier() const;
+
+    static void setDefaultHandlerRepository(MockNamedValueHandlerRepository* repository);
 private:
     SimpleString name_;
     SimpleString type_;
@@ -124,7 +154,8 @@ private:
     } value_;
     size_t size_;
     MockNamedValueComparator* comparator_;
-    static MockNamedValueComparatorRepository* defaultRepository_;
+    MockNamedValueCopier* copier_;
+    static MockNamedValueHandlerRepository* defaultRepository_;
 };
 
 class MockNamedValueListNode
@@ -167,18 +198,28 @@ private:
  */
 
 struct MockNamedValueComparatorRepositoryNode;
-class MockNamedValueComparatorRepository
+struct MockNamedValueCopierRepositoryNode;
+class MockNamedValueHandlerRepository
 {
-    MockNamedValueComparatorRepositoryNode* head_;
+private:
+    MockNamedValueComparatorRepositoryNode* comparatorsHead_;
+    MockNamedValueCopierRepositoryNode* copiersHead_;
+
 public:
-    MockNamedValueComparatorRepository();
-    virtual ~MockNamedValueComparatorRepository();
+    MockNamedValueHandlerRepository();
+    virtual ~MockNamedValueHandlerRepository();
+
+    virtual void installHandlers(const MockNamedValueHandlerRepository& repository);
 
     virtual void installComparator(const SimpleString& name, MockNamedValueComparator& comparator);
-    virtual void installComparators(const MockNamedValueComparatorRepository& repository);
     virtual MockNamedValueComparator* getComparatorForType(const SimpleString& name);
 
-    void clear();
+    virtual void installCopier(const SimpleString& name, MockNamedValueCopier& copier);
+    virtual MockNamedValueCopier* getCopierForType(const SimpleString& name);
+
+    void clearComparators();
+    void clearCopiers();
+    void clearAll();
 };
 
 #endif
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -116,6 +116,7 @@ public:
     virtual void setName(const char* name);
 
     virtual bool equals(const MockNamedValue& p) const;
+    virtual bool compatibleForCopying(const MockNamedValue& p) const;
 
     virtual SimpleString toString() const;
 
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -106,8 +106,12 @@ public:
     virtual void setDefaultComparatorRepository();
 
     virtual void installComparator(const SimpleString& typeName, MockNamedValueComparator& comparator);
-    virtual void installComparators(const MockNamedValueComparatorRepository& repository);
+    virtual void installCopier(const SimpleString& typeName, MockNamedValueCopier& copier);
+    virtual void installHandlers(const MockNamedValueHandlerRepository& repository);
+
     virtual void removeAllComparators();
+    virtual void removeAllCopiers();
+    virtual void removeAllHandlers();
 
 protected:
     MockSupport* clone();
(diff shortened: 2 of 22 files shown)


include/CppUTestExt/MockSupport.h
```diff
@@ -106,8 +106,12 @@ class MockSupport
     virtual void setDefaultComparatorRepository();
 
     virtual void installComparator(const SimpleString& typeName, MockNamedValueComparator& comparator);
-    virtual void installComparators(const MockNamedValueComparatorRepository& repository);
+    virtual void installCopier(const SimpleString& typeName, MockNamedValueCopier& copier);
+    virtual void installHandlers(const MockNamedValueHandlerRepository& repository);
+
     virtual void removeAllComparators();
+    virtual void removeAllCopiers();
+    virtual void removeAllHandlers();
```
arstrube: Duh! Of course, keeping around the old methods while providing new ones is the way to go.

src/CppUTestExt/MockFailure.cpp
```diff
@@ -160,10 +160,22 @@ MockUnexpectedInputParameterFailure::MockUnexpectedInputParameterFailure(UtestSh
...
+        message_ += functionName;
+        message_ += "\": ";
+        message_ += parameter.getName();
+    }
+    else {
+        message_ = "Mock Failure: Unexpected parameter type to parameter \"";
+        message_ += parameter.getName();
+        message_ += "\" to function \"";
+        message_ += functionName;
+    }
```
arstrube: It feels to me like this covers two different failures, like:
```c++
class MockUnexpectedOutputParameterNameFailure
class MockUnexpectedOutputParameterTypeFailure
```

src/CppUTestExt/MockExpectedCall.cpp
```diff
@@ -279,7 +293,7 @@ bool MockCheckedExpectedCall::hasInputParameter(const MockNamedValue& parameter)
 bool MockCheckedExpectedCall::hasOutputParameter(const MockNamedValue& parameter)
 {
     MockNamedValue * p = outputParameters_->getValueByName(parameter.getName());
-    return (p) ? true : ignoreOtherParameters_;
+    return (p) ? p->compatibleForCopying(parameter) : ignoreOtherParameters_;
```
arstrube: Somehow I don't like this. Even when it isn't compatible, it's still an output parameter, no?

arstrube: Or maybe I'm overlooking something here?

arstrube:
Hi Jesús, looks good to me. Only I don't quite understand the reasoning behind checking the name / type of the output parameter. Maybe you could explain. Apart from that I think it's a cleaner approach like that.

arstrube:
```c++
void MockCheckedActualCall::checkOutputParameter(const MockNamedValue& outputParameter)
{
    unfulfilledExpectations_.onlyKeepUnfulfilledExpectationsWithOutputParameter(outputParameter);

    if (unfulfilledExpectations_.isEmpty()) {
        MockUnexpectedOutputParameterFailure failure(getTest(), getName(), outputParameter, allExpectations_);
        failTest(failure);
        return;
    }

    unfulfilledExpectations_.outputParameterWasPassed(outputParameter.getName());
    finalizeCallWhenFulfilled();
}
```
If you want to distinguish the two cases: Unexpected output parameter (wrong name), and output parameter of the wrong type, then I guess the name failure is just an ordinary unexpected parameter failure, which would be caught here, but the unexpected type failure should be caught separately wherever you detect the incompatible type.

tests/CppUTestExt/MockExpectedCallTest.cpp
```diff
@@ -413,6 +476,33 @@ TEST(MockExpectedCall, hasNoOutputParameter)
...
+    foo.setObjectPointer("TypeForTestingExpectedFunctionCall", &object);
+    CHECK_FALSE(call->hasOutputParameter(foo));
+}
+
+TEST(MockExpectedCall, hasNoOutputParameterOfTypeDifferentType)
+{
+    TypeForTestingExpectedFunctionCall object(543);
+    call->withOutputParameterOfTypeReturning("TypeForTestingExpectedFunctionCall", "foo", &object);
+    MockNamedValue foo("foo");
+    foo.setObjectPointer("OtherTypeForTestingExpectedFunctionCall", &object);
```
arstrube: How do you suppose this can happen? I mean, in normal execution, the object pointer is based on the information in withOutputPointerReturning(), so how can the type be different? Can you provoke this testing through the MockSuppor interface?

include/CppUTestExt/MockNamedValue.h
```diff
@@ -60,13 +60,41 @@ class MockFunctionComparator : public MockNamedValueComparator
...
+ * This is needed when copying values of non-native type.
+ */
+
+class MockNamedValueCopier
+{
+public:
+    MockNamedValueCopier() {}
+    virtual ~MockNamedValueCopier() {}
+
+    virtual void copy(void* dst, const void* src)=0;
```
arstrube: Why not `cont void* dst`?

src/CppUTestExt/MockExpectedCall.cpp
jgonzalezdr: Compare it with the sibling `hasInputParameter`. The latter checks that there is an expected call with the same name, same parameter type and same parameter value (these last two by using `equals`). In the case of ouput parameters, checking value is not necessary, but checking type is necessary, otherwise we could copy an incompatible type in the top of the other, so we need an extra function that checks that types check.

In the old implementation that wasn't necessary, because the types could only be `void`.

src/CppUTestExt/MockFailure.cpp
arstrube: There is a `message_ += "\": ";`missing before line 178

src/CppUTestExt/MockExpectedCall.cpp
arstrube: Sorry, didn't see your comment. You are right, I already deleted the note :)

arstrube:
I accidentally came across something weird. Given the following tests:
```c++
//--------------------------------------------------------------------------------------------
struct DummyForTesting {
};
class DummyForTestingDummyCopier : public MockNamedValueCopier {
    virtual void copy(void*, const void*) {}
};
//--------------------------------------------------------------------------------------------
TEST_GROUP(Copier) {};

TEST(Copier, NextTest)
{
    mock().checkExpectations();
    mock().clear();
}

TEST(Copier, DummyForTesting_weirdFailureWhenMissingCopier)
{
    DummyForTestingDummyCopier copier;
    mock().installCopier("Other", copier);
    DummyForTesting source, destination;
    mock().expectOneCall("foo").withOutputParameterOfTypeReturning("Dummy", "dummy", &source);
    mock().actualCall("foo").withOutputParameterOfType("Dummy", "dummy", &destination);

    mock().checkExpectations();

    mock().clear();
    mock().removeAllHandlers();
}
```
I get the following 2 (!) failures:
```bash
$ make BugTests
g++ -Wextra -Wall -Werror -g -I/e/cpputest/include -L/e/cpputest/lib BugTests.cpp -lCppUTest -lCppUTe
stExt -o BugTests.exe
./BugTests.exe
...
BugTests.cpp:61: error: Failure in TEST(Copier, DummyForTesting_weirdFailureWhenMissingCopier)
D:/dev/05_CppUTest/cpputest/src/CppUTestExt/MockNamedValue.cpp:193: error:
        expected <const void*>
        but was  <Dummy>
        difference starts at position 0 at: <          Dummy     >
                                                       ^

.
BugTests.cpp:55: error: Failure in TEST(Copier, NextTest)
D:/dev/05_CppUTest/cpputest/src/CppUTestExt/MockActualCall.cpp:285: error:
        Actual call is in progress. Checking expectations. But no unfulfilled expectations. Cannot ha
ppen.

.
Errors (2 failures, 5 tests, 5 ran, 10 checks, 0 ignored, 0 filtered out, 10 ms)

Makefile:21: recipe for target 'BugTests.exe' failed
make: *** [BugTests.exe] Error 2
```
This doesn't seem quite right?

arstrube:
Let me check what happens when I run this past the other two PRs.

arstrube:
My first attempt has the same problem.

arstrube:
So does my second proposal...

arstrube:
We must be jumping out of an actual call where we shouldn't?

include/CppUTestExt/MockNamedValue.h
jgonzalezdr: For several reasons, all related:
1. `void *` represents a pointer to modifiable memory. `const void*` represents a pointer to non-modifiable memory. Of course, the object that I copy from shall be non-modifiable, but the object that I copy into shall be modifiable. Of course, I could declare `dst` as `const void*` then cast to non-const, but that would remove the "hint" in the interface definition that `dst` has to be modifiable and `src` not.
2. To mimic the "memcpy" interface.
3. Because the expected call method gets a `const void *`, and the actual call method gets a `const void*`, so it's pretty straightforward where each parameter finally goes into the call to the copier (so they're not messed the other way).

arstrube: You are right. The original suggestion by @random85 had `const void*`, but that was silently cast away by casting to (Object*) for dst. `void *` it shall be.

jgonzalezdr:
No, I've merged some of the code from your 1st attempt to check that a copier is installed for custom types. Otherwise, when no copier was found the object was tried to be copied just as a buffer, that's why it asserted that it couldn't do it because it couldn't retrieve a `const void*` from the parameter. 

In fact, the error text was quite misguiding, because "expected" didn't refer to the expected call, but rather to the attempt to retrieve the pointer via `getConstVoidPointer`.

arstrube:
Right, I came to the same conclusion.

arstrube:
@basvodde - do we have to try and break up this one into smaller PRs?

jgonzalezdr:
@arstrube  I actually can't see how to break this up, as it just resolves a single problematic and there isn't anything that can be considered optional. If you just remove any change, the whole thing won't be working properly...

arstrube:
@jgonzalezdr - Experience has taught me that @basvodde may not find the time to even look at a PR this size. I do agree with you about much of it being interrelated (which is why at first I tried a 'cheap' solution with few changes). However, I do think it can be improved at least a little. I was thinking along these lines:

PR 1. MockNamedValue.cpp/.h additions plus tests (should be fairly independent).
PR 2. MockFailure.cpp/.h addition (is independent).
PR 3. Interface changes with no tests (they won't be called, so the build should be fine)
PR 4. Implementation plus a few tests. (probably still big; could perhaps split ActualCall vs. ExpectedCall)
PR 5. MockSupport.cpp / MockSupport.h can come later (if the above are tested directly)
PR 6. MockSupport_c.cpp

Obviously, it can only be used after all PR's have been merged.

As for usefulness, I couldn't agree more :)

arstrube:
I'll play around with it when I find the time.

jgonzalezdr:
@arstrube I see the point, no need to push "useful" PRs, but just "digestible" ones 'till the puzzle is complete.. OK, I'll try that this weekend.

basvodde:
Can I close this one now?

jgonzalezdr:
Yes, the merged implementation is working, so this can be closed.

@basvodde @arstrube Sorry for not having actively participated in the subsequent discussions of last week, but my workload was way too big to do anything else besides trying to sleep a little...

basvodde:
Thats open source life :)

PR #678, opened 2015-07-02
Custom Copiers #1: Refactorings in preparation for

arstrube:
addition of copiers to copy non-native types.

This one deals with renaming of existing classes and members appropriately.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -28,7 +28,7 @@
 #ifndef D_MockNamedValue_h
 #define D_MockNamedValue_h
 /*
- * MockParameterComparator is an interface that needs to be used when creating Comparators.
+ * MockNamedValueComparator is an interface that needs to be used when creating Comparators.
  * This is needed when comparing values of non-native type.
  */
 
@@ -66,7 +66,7 @@ private:
  * Basically this class ties together a Name, a Value, a Type, and a Comparator
  */
 
-class MockNamedValueComparatorRepository;
+class MockNamedValueHandlerRepository;
 class MockNamedValue
 {
 public:
@@ -106,7 +106,7 @@ public:
     virtual size_t getSize() const;
     virtual MockNamedValueComparator* getComparator() const;
 
-    static void setDefaultComparatorRepository(MockNamedValueComparatorRepository* repository);
+    static void setDefaultHandlerRepository(MockNamedValueHandlerRepository* repository);
 private:
     SimpleString name_;
     SimpleString type_;
@@ -124,7 +124,7 @@ private:
     } value_;
     size_t size_;
     MockNamedValueComparator* comparator_;
-    static MockNamedValueComparatorRepository* defaultRepository_;
+    static MockNamedValueHandlerRepository* defaultRepository_;
 };
 
 class MockNamedValueListNode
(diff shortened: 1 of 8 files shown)


basvodde:
I'm not ok with Handlers. Why not Comparators and Copiers?

Handler is a word that means that you don't have a good concept.

arstrube:
Why not? Handlers are an aggregation of Comparators and Copiers (and possibly other classes that _handle_ custom objects). 

Are you suggesting we should just duplicate the whole thing, and we should not have any methods (such as RemoveAllHandlers()) that can operate on both?

arstrube:
It would also be helpful if you had some suggestions for names you find more appropriate.

arstrube:
And, please take a look at #680 to see where this is really headed. Using a common name allows us to have one repository, instead o _n_ repositories.

arstrube:
@jgonzalezdr what's your take on having 2 separate repositories, on for Comparators, and another for Copiers?

basvodde:
Re: Handlers. All objects are handlers, they almost always 'handle' other objects. "Handlers" usually means someone couldn't think of a better name.

When Copiers and Comparators are grouped together and the only thing they mean is CopierAndComparator, then renaming them to Handler doesn't make it any better.

How much actual LOCs would be added when we use 2 repositories? If it is a lot, can we generalize the repository (to e.g. void*) to avoid the duplication? I wouldn't want to put Comparators and Copiers together when we use them mostly separately and they actually do not share any code or logic.

arstrube:
>When Copiers and Comparators are grouped together and the only thing they mean is CopierAndComparator, then renaming them to Handler doesn't make it any better.

Maybe not, but at least it's shorter :-P 
`MockNamedValueComparatorAndCopierRepository`,  `setDefaultComparatorAndCopierRepository()`, `removeAllComparatorsAndCopiers()`
I can't get myself to like it.....

>How much actual LOCs would be added when we use 2 repositories?

I think it's a bit difficult to guesstimate.... There are ten methods and seven attributes. Of those, 8 methods and 4 attributes would be identical.

I don't like the void* approach. It is error prone. Other than that, one could derive both repositories from a common base, or have a common base class for Comparators and Copiers. In either case, we would need a good name, and we are not allowed to use "Handler". So if you ask me, the problem is still basically the same: What, in your opinion, is the common base of Comparators and Copiers?

arstrube:
@basvodde how about Functor? e.g. MockNamedValue**Functor**Repository etc.

basvodde:
Let me check a bit later. Not quite sure why we would need a MockNamedValueFunctor repository then...

arstrube:
Sure. And why not use "Functor" with the existing PR.

jgonzalezdr:
From my point of view, I think that it's an overkill to separate in two repositories, it will make the whole thing less maintainable, because each time that the repository is passed around, we should have to pass two repositories... add more parameters to initialization and copying functions and methods... At the end, it will be just boilerplate code, it will be less readable, less maintainable, less robust...

It's true that "Handler" means nothing (and it means everything at the same time), but sometimes when you've got something that can do many different things it's impossible to find a very precise name without paraphrasing.

In the case "Functor" I wouldn't use that name, because comparators have two functions, which is against the paradigm of functors (an object that executes a single function). But I see the point of @arstrube.

The fact is that we have a repository of object that operate on the parameters of calls. No more, no less. A better name might then be just "ParameterOperators". We would have then a "parameterOperatorsRepository_" attribute, a "removeAllParameterOperators" method, etc.. 

If you don't like it, taking in account that the operations that we perform are comparisons and copies, then "ComparatorsAndCopiers" would be the only solution left. I wouldn't like that if the set of different operations that can be performed on parameters could increase ("ComparatorsAndCopiersAndWhaterver", ouch!), but I can't think of something else that can be performed on parameters.

arstrube:
@jgonzalezdr nicely put. I wasn't entirely sure myself because Comparators do have two functions -- strictly speaking, converting a value to a string has nothing to do with comparing. Not any more than copying, actually. My personal preference would be:

1. Current solution, as is.
2. Current solution, named MockNamedValueOperatorRepository
3. Current solution,  named ParameterOperatorRepository
4. Solution based on my first PR (e.g. add a copy() method to MockNamedValueComparator (see above)
...

basvodde:
How about #689 ?

arstrube:
Hmm. It's wordy and it violates the Open-Closed Principle. But if you're happy with it, I'd rather go for that than getting stuck on discussing names. I think the new functionality is much needed.

basvodde:
@arstrube @jgonzalezdr I thought about the Operators as a name and felt it was good. Still, in #689 I chose for ComparatorsAndCopiers because I guess there won't be another one. (guess!) Then ComparatorsAndCopiers is just more clear than Operators. If another one is added, renaming it to Operators would make sense.

arstrube:
@basvodde sure. As for me, I can live with that. As I said above, I'd rather go along with your name than waste more time on discussions :). And yes, we can switch to operators if a third one is ever added.

arstrube:

PR #680, opened 2015-07-02
Custom Copiers #3: MockNamedValue Changes & Additions

arstrube:
in preparation for addition of copiers to copy non-native types.

This one completes the necessary changes and additions to the MockNamedValue class. There is now a list of MockNamedValueCopiers in addition to the list of MockNamedValueComparators and methods to handle both.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -60,6 +60,34 @@ private:
 };
 
 /*
+ * MockNamedValueCopier is an interface that needs to be used when creating Copiers.
+ * This is needed when copying values of non-native type.
+ */
+
+class MockNamedValueCopier
+{
+public:
+    MockNamedValueCopier() {}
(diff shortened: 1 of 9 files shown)


include/CppUTestExt/MockNamedValue.h
```diff
@@ -167,18 +199,28 @@ class MockNamedValueList
...
-class MockNamedValueComparatorRepository
+struct MockNamedValueCopierRepositoryNode;
+class MockNamedValueHandlerRepository
 {
-    MockNamedValueComparatorRepositoryNode* head_;
+private:
+    MockNamedValueComparatorRepositoryNode* comparatorsHead_;
+    MockNamedValueCopierRepositoryNode* copiersHead_;
+
 public:
```
arstrube: @basvodde why not have two different lists like this? Especially seeing that we do not -- as you pointed out -- need comparators and copiers together? Then, when we need them, we just look into the appropriate list to start with.

arstrube:

PR #684, opened 2015-07-02
Custom Copiers #7: Changes for MockSupport_c

arstrube:
in order to support copiers to copy non-native types.

TODO: Don't we need some tests for MockSupport_c?


arstrube:

Issue #686, opened 2015-07-02
ignoreOtherParameters combined with withOutputParameter produces strange behavior


itavero:
Given the following test:
```cpp
TEST(bug, ignore_output_param)
{
        int param = 1;
        int retval = 2;
        mock().expectOneCall("function")
                .withParameter("boo", 5)
                .ignoreOtherParameters();
        mock().actualCall("function")
                .withOutputParameter("parameterName", &param).withParameter("boo", 5);
        CHECK_EQUAL(param, 1);
        CHECK_EQUAL(retval, 2);
        mock().checkExpectations();
}
```

I get the following output:
```shell
 error: Failure in TEST(bug, ignore_output_param)
cpputest-3.7.2/src/CppUTestExt/MockNamedValue.cpp:190: error:
	expected <const void*>
	but was  <int>
	difference starts at position 0 at: <          int       >
	                                               ^
```

Adding `withOutputParameterReturning` causes the test to succeed, however I do not want the mock to modify the data the pointer is referring too. I just want to ignore it, which seems impossible with the built-in feature!

arstrube:
It seems `ignoreOtherParameters()` is not handled correctly with output parameters.

As a "quick fix" workaround, if all you are ever going to do is ignoring it, you could simply omit the output parameter from the `actualCall()`. I know this isn't 'clean', but maybe it would be acceptable.

Another workaround might be to use `withParameter()` on `parameterName` in the test instead. To be able to do this, you need to know the actual address of the pointer, which you do know in the simple example above, but probably not in production code.

No question this needs to be fixed in CppUTest, though.

itavero:
I am currently using `mock().setData( ... )` to inject some extra data into the mock function, so I can decide whether or not to include the output parameter, like so:

```cpp
    MockActualCall& actual_call = mock().actualCall(__FUNCTION__).withPointerParameter("object", (void *) object);

    if (mock().getData("not_use_output").getUnsignedIntValue() == 0)
    {
        actual_call.withOutputParameter("some_out", (void *) out);
    }
```

This is OK for a workaround, but I would rather just use `ignoreOtherParameters`.

arstrube:
Sure. This needs to be fixed.

PR #687, opened 2015-07-02
Fix ignored output parameter

arstrube:
As @itavero pointed out in #686, an attempt is made to copy the object, even though no expectation was specified and the actual call should have been ignored.

A strange failure occurs in the test, because an attempt is made to retrieve a `const void*` value from a nameless default `MockNamedValue`, which holds an `int` value.

So far, this pull request contains only the failing test that proves this behavior -- @ryanplusplus could you perhaps help with fixing this?


arstrube:
Improved fix is now integrated in the Copier PR (#688)

PR #689, opened 2015-07-04
Changing the name of the Comparators repository so we can add copiers to it :)

basvodde:
As the title says. Now we can add NamedValueCopier to the Node and voila!

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -66,7 +66,7 @@ private:
  * Basically this class ties together a Name, a Value, a Type, and a Comparator
  */
 
-class MockNamedValueComparatorRepository;
+class MockNamedValueComparatorsAndCopiersRepository;
 class MockNamedValue
 {
 public:
@@ -106,7 +106,7 @@ public:
     virtual size_t getSize() const;
     virtual MockNamedValueComparator* getComparator() const;
 
-    static void setDefaultComparatorRepository(MockNamedValueComparatorRepository* repository);
+    static void setDefaultComparatorsAndCopiersRepository(MockNamedValueComparatorsAndCopiersRepository* repository);
 private:
     SimpleString name_;
     SimpleString type_;
@@ -124,7 +124,7 @@ private:
     } value_;
     size_t size_;
     MockNamedValueComparator* comparator_;
-    static MockNamedValueComparatorRepository* defaultRepository_;
+    static MockNamedValueComparatorsAndCopiersRepository* defaultRepository_;
 };
 
 class MockNamedValueListNode
@@ -162,20 +162,20 @@ private:
 };
 
 /*
- * MockParameterComparatorRepository is a class which stores comparators which can be used for comparing non-native types
+ * MockParameterComparatorRepository is a class which stores comparators and copiers which can be used for comparing non-native types
  *
  */
 
-struct MockNamedValueComparatorRepositoryNode;
-class MockNamedValueComparatorRepository
+struct MockNamedValueComparatorsAndCopiersRepositoryNode;
+class MockNamedValueComparatorsAndCopiersRepository
 {
-    MockNamedValueComparatorRepositoryNode* head_;
+    MockNamedValueComparatorsAndCopiersRepositoryNode* head_;
 public:
-    MockNamedValueComparatorRepository();
-    virtual ~MockNamedValueComparatorRepository();
+    MockNamedValueComparatorsAndCopiersRepository();
+    virtual ~MockNamedValueComparatorsAndCopiersRepository();
 
     virtual void installComparator(const SimpleString& name, MockNamedValueComparator& comparator);
-    virtual void installComparators(const MockNamedValueComparatorRepository& repository);
+    virtual void installComparatorsAndCopiers(const MockNamedValueComparatorsAndCopiersRepository& repository);
     virtual MockNamedValueComparator* getComparatorForType(const SimpleString& name);
 
     void clear();
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -103,10 +103,10 @@ public:
 
     virtual void setMockFailureStandardReporter(MockFailureReporter* reporter);
     virtual void setActiveReporter(MockFailureReporter* activeReporter);
-    virtual void setDefaultComparatorRepository();
+    virtual void setDefaultComparatorsAndCopiersRepository();
 
     virtual void installComparator(const SimpleString& typeName, MockNamedValueComparator& comparator);
-    virtual void installComparators(const MockNamedValueComparatorRepository& repository);
+    virtual void installComparatorsAndCopiers(const MockNamedValueComparatorsAndCopiersRepository& repository);
     virtual void removeAllComparators();
 
 protected:
@@ -127,7 +127,7 @@ private:
     bool enabled_;
     MockCheckedActualCall *lastActualFunctionCall_;
     MockExpectedCallComposite compositeCalls_;
-    MockNamedValueComparatorRepository comparatorRepository_;
+    MockNamedValueComparatorsAndCopiersRepository comparatorsAndCopiersRepository_;
     MockNamedValueList data_;
 
     bool tracing_;
--- a/include/CppUTestExt/MockSupportPlugin.h
+++ b/include/CppUTestExt/MockSupportPlugin.h
@@ -42,7 +42,7 @@ public:
 
     virtual void installComparator(const SimpleString& name, MockNamedValueComparator& comparator);
 private:
-    MockNamedValueComparatorRepository repository_;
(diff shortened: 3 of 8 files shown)


include/CppUTestExt/MockNamedValue.h
```diff
@@ -66,7 +66,7 @@ class MockFunctionComparator : public MockNamedValueComparator
  * Basically this class ties together a Name, a Value, a Type, and a Comparator
```
arstrube: You'll want the correction that @jgonzalezdr did to the comment on line 31 as well

include/CppUTestExt/MockNamedValue.h
```diff
@@ -162,20 +162,20 @@ class MockNamedValueList
...
 
 /*
- * MockParameterComparatorRepository is a class which stores comparators which can be used for comparing non-native types
+ * MockParameterComparatorRepository is a class which stores comparators and copiers which can be used for comparing non-native types
  *
  */
 
-struct MockNamedValueComparatorRepositoryNode;
-class MockNamedValueComparatorRepository
+struct MockNamedValueComparatorsAndCopiersRepositoryNode;
```
arstrube: My line note got messed up. Please keep the node name as is.

arstrube:
As I explained in my line notes, I would keep the name of `struct MockNamedValueComparatorNode and add a different kind of node for the copier (as shown by @jgonzalezdr and myself).

Other than that, if you really want to go with this name, I'll be happy if you want to merge it and we can take it from there.

basvodde:
@arstrube The current node, we can just add a Copier as a member in the node. That way, we can store the copier and comparator with the same String.

arstrube:
@basvodde I see your point now, but the code for retrieving it will not be as simple as it is now, because now we just look for the node and if there is one, return it. With your approach, we would then also have to check that it contains a copier, or a comparator. Personally, I'd prefer to separate lists. This has already been implemented by @jgonzalezdr and I see no reason to change it. Duplicating the struct isn't a problem because it contains only few attributes and no functionality at all.

basvodde:
@arstrube Searching was already implemented in my pull requests, it was one more word in the if statement. Adding is more problematic as we would prefer to add it to an existing node instead of creating a new one (though both are possible).

I'm ok with seperate repositories too. As long as it won't require a Handler or any other class where the Copier and Comparator are put together :) So far, I still dislike that as you seem to rarely need them together.

How to proceed?

arstrube:
Why make such an issue of this? If you leave the node as is (like I requested), fitting our existing PR onto yours is must a matter of renaming. Which can be a pain in the behind, but at least it's just renaming.

If you don't like that idea, I would prefer if you presented a complete solution of your own choosing. This discussion is really getting old. If your solution passes the tests, it will be good enough for me.

arstrube:
As an aside, which pull requests are you talking about? I have only seen this one?

arstrube:
I guess I'm lazy. I don't like being forced to make changes to the my / @jgonzalezdr's implementation when I can't clearly see the benefit. He and myself, we both presented complete solutions, which made it much easier to weigh pros and cons and settle for one (actually, his). In the same way, if you present a complete solution, we might both see its benefits more easily.

arstrube:
BTW if you put Comparators into the same node Copiers, then when installing, you will have to search existing nodes first. You can't blindly add to head_ anymore. Because there might already be a copier for the comparator you're about to install - and vice versa.

arstrube:
How are you going to initialize copier_& when only installing a comparator, and vice versa?

arstrube:
@basvodde With your design, you'll end up having to have both, always. IOW we might as well add an extra method to MockNamedValueComparator (then to be called MockNamedValueComparatorAndCopier) :-(

basvodde:
Sure you can, but the MockNamedValueComparatorAndCopier would be used by CppUTest users and the Repository not or rarely.

arstrube:
Yeah. :-) But do you really want to have to add a copier each time you add a comparator? They're references not pointers in the node.

arstrube:
You can init an unused ppointer to NULL. But you are gonna end up with an unused reference.

basvodde:
@arstrube As you can see in the pull request, it already changed the & to a *. I think it has little effect as the interface is actually the same.

basvodde:
@arstrube @jgonzalezdr I'm ok with either:
1) The ComparatorsAndCopiers repository with *one* list in it one ComperatorAndCopierNode
2) Two separate repositories, one for Comparators and one for Copiers.

Having two lists in one repository doesn't make too much sense to me. And neither does combining the Comparator and Copier under the same interface as a user is likely to use either a Comparator or a Copier.

arstrube:
@basvodde I think it will be best if you go ahead and implement it the way you prefer.

jgonzalezdr:
@basvodde IMO the ComperatorAndCopierNode solution, as @arstrube said, is not a very good idea, as it only has drawbacks: There's no need to go through a list that contains comparators when we are searching for copiers (and there is no use case where you have to find any or both types of parameter operators), and the implementation will be far less complicated and hard to maintain or extend. If you find "conceptually" wrong to have a repo with both type of operators, then having a ComperatorAndCopierNode is even more awful.

I think that having two separate repos makes the internal interface more complex than a repo with two lists, but if I have to vote between the two I'll go for this one for sure.

arstrube:
My vote is for @jgonzalezdr's pull request, with "Operator" instead of "Handler" and otherwise "as is", that is, with separate lists for Comparators and for Copiers.

I strongly dislike both options 1) and 2) above, for the reasons already stated.

I am prepared to go along with using "ComparatorsAndCopiers" for the repository name, if that meant we could go ahead, but frankly, I don't think its a good idea (I also don't call my toolbox "ScrewdriverAndWrenchBox" :P).

basvodde:
Ok, sorry for my slow replies, I'm suppose to be on vacation. I'll try to do something every now and then.

We got 2 issues, in above threads:

1) Name

I actually *had* Operators before changing it to ComparatorsAndCopiers. Why? Well, I thought about it for quite a bit and figured and the ComparatorsAndCopiers are both operators on a type that we externalize as we avoid using templates. So, Operators make sense. I was uncomfortable with Operators as the name is still so general, though less than Handlers. Thinking it over more, I figured that it is unlikely that we have more than ComparatorsAndCopiers, so I decided that ComparatorsAndCopiers is a better name because it *exactly* describes what it is. Usually I wouldn't like names like these, but in *this* case it seems a much more descriptive name than Operators. Likewise, if your toolbox will *only* contain a Screwdriver and Wrench, then a ScrewdrivenAndWrenchBox would be a more descriptive name than a ToolBox :) Nobody would look for a hammer in a ScrewdriverAndWrench box. For me, the critical point is that I can't image (at this moment) to having other operators than Comparators and Copiers.

2) One or Two lists.

Also, I was thinking about this and figured that one list with both items is actually likely to lead to cleaner code. Especially if we do not attempt to combine comparators and copiers inside the same node, then the code would be trivial. Conceptually it made sense to me to have one "type" and the "operators" (which are only Comparator and Copier) in one node.

basvodde:
Oh, btw. I'm not merging any yet without getting to at least a shared understanding on advantages/disadvantages :) Not need to be agreement, but some discussion on it is good as we'll all learn from that.

Whatever we chose, I'll still be relying on merging the changes to the MockSupport and the usage of the Copier from your previous changes as they seemed good!

arstrube:
@basvodde

1. To keep using the allegory, while I can follow your line of reasoning, I still won't have a ScrewdriverAndWrenchBox, ever <g>. But, on the other hand, Toolbox can be too generic indeed (it could, however, be perfectly adaquate for  a _limited_ set of diverse tools). When it becomes too generic, in the case of the toolbox, I tend to organize tools along the lines of their purpose (much like interfaces should be organized around the needs of specific clients). So there may be a Woodworking toolbox, an electrical toolbox, possibly a Japanes toolbox, etc.
How does this relate to our discussion? Well, we actually do **not** have an **Operator**Repository (or heaven forbid, a **Handler**Repository), which I agree, would say exactly nothing. Rather, we have a **MockNamedValue**OperatorRepository which, to me, narrows down the definition of a rather wide term to exactly what it should be. Operators on MockNamedValues is exactly what we have here.

2. One or Two Lists: No matter how hard you try to explain, I just can't see it. It would really help if you could provide a complete, mergeable implementation of the 'one-list-contender' in its full glory and triviality. True simplicity has a charm that is hard to resist, you know ;-)

arstrube:
Btw I removed my other PRs because they heavily depend on the naming of internal interfaces and thus make no sense anymore, whatever we decide. There is still @jgonzalezdr PR which we can draw on.

basvodde:
@arstrube Written in 10 minutes mergable implementation of the one repository at: #691. It'll need a bit more work on the InstallComparatorAndCopier function, but other than that, I guess this would work.

As said, we *could* make the copier and comparators in the same node, that would be a bit more work (not much). But, I'll need to catch some sleep now. Enjoying Barcelona with 2 little kids, which made me tired.

PR #690, opened 2015-07-05
Fix bug when ignoreOtherParameters() is combined with withOutputParameter()

### LINKED ISSUE #686: ignoreOtherParameters combined with withOutputParameter produces strange behavior

arstrube:
This fixes #686.

### DIFF
--- a/src/CppUTestExt/MockActualCall.cpp
+++ b/src/CppUTestExt/MockActualCall.cpp
@@ -82,9 +82,12 @@ void MockCheckedActualCall::finalizeOutputParameters(MockCheckedExpectedCall* ex
 {
     for (MockOutputParametersListNode* p = outputParameterExpectations_; p; p = p->next_)
     {
-        const void* data = expectedCall->getOutputParameter(*p->name_).getConstPointerValue();
-        size_t size = expectedCall->getOutputParameter(*p->name_).getSize();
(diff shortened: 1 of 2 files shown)


arstrube:
The problem is that, when no expectation was specified, it tried to compare with the default MockNamedValue, which has name_ "" and type_ "int". So you got an error that "const void*" was expected, but it got "int" (not a mock expectation error). Which was strange indeed.

PR #691, opened 2015-07-07
Demo - Stupid repository implementation

basvodde:
This is an example of how you can implement the repository. Not done yet.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -66,7 +66,7 @@ private:
  * Basically this class ties together a Name, a Value, a Type, and a Comparator
  */
 
-class MockNamedValueComparatorRepository;
+class MockNamedValueComparatorsAndCopiersRepository;
 class MockNamedValue
 {
 public:
@@ -106,7 +106,7 @@ public:
     virtual size_t getSize() const;
     virtual MockNamedValueComparator* getComparator() const;
 
-    static void setDefaultComparatorRepository(MockNamedValueComparatorRepository* repository);
+    static void setDefaultComparatorsAndCopiersRepository(MockNamedValueComparatorsAndCopiersRepository* repository);
 private:
     SimpleString name_;
     SimpleString type_;
@@ -124,7 +124,7 @@ private:
     } value_;
     size_t size_;
     MockNamedValueComparator* comparator_;
-    static MockNamedValueComparatorRepository* defaultRepository_;
+    static MockNamedValueComparatorsAndCopiersRepository* defaultRepository_;
 };
 
 class MockNamedValueListNode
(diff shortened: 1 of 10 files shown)


arstrube:
Okay, looking good so far. Could you please use the established order for the arguments in `copy()` though?

arstrube:
Now I'm looking closer, the order of arguments is actually correct. But I find the naming really confusing. To me, `out` is and **output** parameter and `in` is an **input** parameter. I would prefer something like `src` and `dst` or even just `obj1` and `obj2` which at least doesn't lead me to wrong assumptions.

arstrube:
Looking good so far. 

I deleted a bunch of comments I made earlier, because somehow I got it all wrong. Not quite awake yet? I wonder....

include/CppUTestExt/MockNamedValue.h
```diff
@@ -42,6 +42,16 @@ class MockNamedValueComparator
     virtual SimpleString valueToString(const void* object)=0;
 };
 
+class MockNamedValueCopier
+{
+public:
+    MockNamedValueCopier() {}
+    virtual ~MockNamedValueCopier() {}
+
+    virtual void copy(const void* in, void* out)=0;
```
arstrube: Please put the output parameter first. Doing so is a common idiom used in many places, SimpleString among others.

arstrube:
I did a head-to-head comparison on my fork, by pushing a similar "stupid" implementation of the 2-list approach. I concede that the one-list implementation is preferable:
+ It has fewer changes, overall
+ It is simpler
+ It doesn't need ClearComparators() and ClearCopiers(), which I never really liked
+ It can keep clear(), as opposed to having clearAll()

I would like to suggest the following small improvements (for which I can provide PRs)
+ Follow the standard idiom regarding the arguments of copy()
+ Use of _override in the test cases 
+ Add a test case to demonstrate that, when a comparator and a copier are installed for the same type, both will be found
+ Use "MyType" as typename rather than "copier", which is misleading

basvodde:
@arstrube I'm fine with all suggestions. I just wrote the stupid version to show that implementing it was really easy :)

So, considering that the implementation *is* really easy? Do you and @jgonzalezdr want to decide between the three solutions. The choices are:

1) two separate repositories
2) One repository with two lists
3) One repository with one list

What do you prefer?

arstrube:
Cunning :) 

I actually prefer 3) now I've seen it, obiously with the suggested improvements ;-)

Is there still a chance to vote for "Operator" instead of ComparatorsAndCopiers, perhaps? 

The stupid version isn't that stupid, really -- there isn't a whole lot missing to complete the MockNamedValue aspect of things. (can do PR for that).

basvodde:
Heh, it was stupid because it doesn't store the Comparator and the Copier in the same node. That would make it a bit more complicated... not a lot though. Just the inserting would change.

arstrube:
Oh, what a shame. I am really disappointed that you want to make it more complicated. That way, you will lose your advantage over the two-lists competition. Maybe I'll revise my vote, after all.....

It works perfectly fine as is, although perhaps, inelegantly. I added a test case that proves it. They don't have to be in the same node. I thought that was very clever of you once I realized that. Now you call it stupid =(

arstrube:
Having to search the nodes for inserting was one of the major drawbacks I saw in your proposal. You don't want to retrofit something like that now, do you?

arstrube:
My vote now goes to:

4) One repository with one _stupid_ list

Bah =)

arstrube:
So I grafted the remainder of our changes and tests onto the _simple_ (not stupid) list and all works just fine.

The only thing I dislike is that there are still two lists in MockSupport_c; would need to put in a bit more thought on how to get rid of that.

arstrube:
How about the MockSupport interface. Currently, we have: `RemoveAllComparators()`.
Do we just add `RemoveAllCopiers()` (this would require corresponding repository methods)?
Or do we do the otherwise preferable `RemoveAllComparatorsAndCopiers()`and break the interface?

basvodde:
I suggest RemoveAllCompiers.

arstrube:
o.k.

PR #693, opened 2015-07-08
Added support to get mock data via pointer.

bithium:
It can be useful to access the actual MockNamedValue object associated with a mock data and not a
copy, e.g., to use the size attribute that is already present in the MockNamedValue, to store the size of the data that the pointer hold by the MockNamedValue object has.

### DIFF
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -78,6 +78,8 @@ public:
     void setDataObject(const SimpleString& name, const SimpleString& type, void* value);
     MockNamedValue getData(const SimpleString& name);
 
+    MockNamedValue *getDataPointer(const SimpleString& name);
+
     MockSupport* getMockSupportScope(const SimpleString& name);
 
     const char* getTraceOutput();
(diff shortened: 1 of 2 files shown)


basvodde:
I'm not understanding when it would be useful to access the pointer directly. So far, I'm opposing this PR :)

Could you explain what you are trying to achieve with this?

bithium:
Hello,

this came up when I was overwriting the system __write__ function and wanted to store the data received by the function.

My solution was to create a new mock data to store the received data on the write, however I need a way to store how much data the buffer I save has.

Given that the __MockNamedValue__ class already has a field size, I wanted to reuse it for the purpose above.

The problem is that there is no way of accessing the object created after a __MockSupport::setData__ call. The __MockSupport::getData__ returns a copy of the object so calling __MockSupport::setData__ to create the object and then using __MockSupport::getData__ to set the size field does not work.

basvodde:
Ok. I understand, though wouldn't prefer to re-purpose the size and neither to be able to get a pointer to the MockNamedValue from the MockSupport interface.

I've in the past just used a different SetData with the size information.

basvodde:
I'll close this without merging, unless there is a better reason for merging it :) (sorry!)

PR #694, opened 2015-07-09
For Discussion: Refactor MockFunctionComparator / MockCFunctionComparatorNode

arstrube:
I noticed that `MockFunctionComparator is more or less duplicated by `MockCFunctionComparatorNode`. I was wondering whether this could be improved:

Why do we have MockCFunctionComparatorNode? Why do we duplicate the list in MockSupport_c? I came up with two reasons:
1. We need some way to delete the comparators that we create from MockSupport_c. The list keeps track of them, so we can delete them.
2. We need them to be compatible to `MockTypeEqualFunction_c & Co, so we can't use the C++ function types.

To make this (hopefully), I see two options.
1. Remove MockFunctionComparator. It "might" be used be user code, of course. But it is not used by CppUTest.
2. Refactor MockFunctionComparator, so it can be used in MockSupport_c. This is what is attempted in this pull request.

The motivation is to simplify the upcoming changes for Copiers.

### DIFF
--- a/include/CppUTestExt/MockNamedValue.h
+++ b/include/CppUTestExt/MockNamedValue.h
@@ -28,7 +28,7 @@
 #ifndef D_MockNamedValue_h
 #define D_MockNamedValue_h
 /*
- * MockParameterComparator is an interface that needs to be used when creating Comparators.
+ * MockNamedValueComparator is an interface that needs to be used when creating Comparators.
  * This is needed when comparing values of non-native type.
  */
 
@@ -42,12 +42,12 @@ public:
     virtual SimpleString valueToString(const void* object)=0;
 };
 
+extern "C" typedef int (*isEqualFunction)(const void*, const void*);
+extern "C" typedef const char* (*valueToStringFunction)(const void*);
+
 class MockFunctionComparator : public MockNamedValueComparator
 {
 public:
-    typedef bool (*isEqualFunction)(const void*, const void*);
-    typedef SimpleString (*valueToStringFunction)(const void*);
-
     MockFunctionComparator(isEqualFunction equal, valueToStringFunction valToString)
         : equal_(equal), valueToString_(valToString) {}
     virtual ~MockFunctionComparator(){}
--- a/include/CppUTestExt/MockSupport_c.h
+++ b/include/CppUTestExt/MockSupport_c.h
@@ -104,7 +104,7 @@ struct SMockExpectedCall_c
 };
 
 typedef int (*MockTypeEqualFunction_c)(const void* object1, const void* object2);
-typedef char* (*MockTypeValueToStringFunction_c)(const void* object1);
+typedef const char* (*MockTypeValueToStringFunction_c)(const void* object1);
 
 typedef struct SMockSupport_c MockSupport_c;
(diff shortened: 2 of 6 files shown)


basvodde:
Ok. I don't understand the changes yet, but... before that... I noticed at some point the MockParameterComparator was renamed to MockNamedValueComparator? That feels wrong. I must have missed this before? (not checked history yet)

basvodde:
I guess I've done that myself. I'm confused about it as the user doesn't need to know about a NamedValue when using a Comparator, right?

basvodde:
I think I'm now understanding the change and so far not liking it (sorry). I think it is conceptually wrong to override a node from a FunctionComparator. The FunctionComparator is a comfort class so you can use C functions and shouldn't be inherited from. As the duplication removed is minimal, I suggest not to do this.

arstrube:
Okay... not feeling strongly about this. =)
Interesting about the MockParameterComparator. Yes, that's right. However, as long as I know the insides of CppUTest, it's "always" been like that.

basvodde:
Obviously when writing it, at some point, I wanted it to be called MockParameterComparator and then named it differently. Do we want to change that? I guess MockParameterComparator is a much more descriptive name...

arstrube:
According to Blame, it's never been changed. I was wondering about this myself. I thought there might be a reason MockNamedValue is also used on return values, of course. But other then that, I agree that MockParameterComparator makes much more sense, from the user's point of view.

How much would a change like this impact the public user interface?

basvodde:
Yeah, but comparators are not used THAT often. We can also extract the Comparator and Copier to a separate file then.

But should I start with merging the renaming pull requests that I opened, so we can build from there?

arstrube:
> As the duplication removed is minimal, I suggest not to do this.

aye, aye. But I did save 10 lines of code. That will make roughly 20 lines less, once copiers are added =P

basvodde:
Agree on that :) Purely from that perspective, it was good :)

arstrube:
The renaming, and the not-so-stupid one-lister?

basvodde:
I think its easier if you do the simple list with the changes you wanted?

arstrube:
That's fine with me; I have a couple of follow-up PRs, including one containing "the rest" of the implementation ^^.

I'll shed a tear or two saying Goodbye to "MockNamedValueOperatorRepository" though =(

Also, would you want to go through with "MockParameterXXXRepository etc.? This would be a good time to do so, I guess.

basvodde:
Yeah, when we get *one* more operator than comparators and copiers, we'll do OperatorRepository. In the mean time, I expect your toolbox to be renamed to ScrewdriverAndWrenchToolbox as long as these are the only tools you got ;P

The changed from NamedValue to Parameter seems ok.

arstrube:
Oops. That change from NamedValue to Parameter isn't in the PR now. Maybe add it later, before the next step?

Don't worry, I won't ever have a toolbox named anything like that, for the simple reason that I use many tools =). Mostly I need more than two of them at any given time....

basvodde:
Yeah, I figured that change can be later, this was just a start

PR #695, opened 2015-07-09
Copiers one list

arstrube:
That would be this one


arstrube:
TODO: Need to add the new test file to Cmake still! Not doing so means the tests won't be run in Cmake.

PR #696, opened 2015-07-09
Copiers one list 2

arstrube:
For review. This should be quite "digestable"

It does contain the renaming of the comment.... ah well, can rename it back later...


tests/CppUTestExt/MockFailureTest.cpp
```diff
@@ -200,7 +200,7 @@ TEST(MockFailureTest, MockExpectedParameterDidntHappenFailure)
 TEST(MockFailureTest, MockNoWayToCompareCustomTypeFailure)
 {
     MockNoWayToCompareCustomTypeFailure failure(UtestShell::getCurrent(), "myType");
-    STRCMP_EQUAL("MockFailure: No way to compare type <myType>. Please install a ParameterTypeComparator.", failure.getMessage().asCharString());
```
arstrube: I always found this message a bit strange ;-)

basvodde: If I wrote the original code, I must have been very distracted here :)

PR #697, opened 2015-07-09
Copiers one list 3

arstrube:
Here is "the rest". We make have to break it down....

### DIFF
--- a/src/CppUTestExt/MockSupport_c.cpp
+++ b/src/CppUTestExt/MockSupport_c.cpp
@@ -95,6 +95,24 @@ public:
 
 static MockCFunctionComparatorNode* comparatorList_ = NULL;
 
+class MockCFunctionCopierNode : public MockNamedValueCopier
+{
+public:
+    MockCFunctionCopierNode(MockCFunctionCopierNode* next, MockTypeCopyFunction_c copier)
+        : next_(next), copier_(copier) {}
+    virtual ~MockCFunctionCopierNode() {}
+
+    virtual void copy(void* dst, const void* src) _override
+    {
+        copier_(dst, src);
+    }
+
+    MockCFunctionCopierNode* next_;
+    MockTypeCopyFunction_c copier_;
+};
+
+static MockCFunctionCopierNode* copierList_ = NULL;
+
 extern "C" {
 
 MockExpectedCall_c* expectOneCall_c(const char* name);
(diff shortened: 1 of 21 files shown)


src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -72,28 +72,36 @@ static MockExpectedCall* expectedCall = NULL;
 static MockActualCall* actualCall = NULL;
 static MockFailureReporterForInCOnlyCode failureReporterForC;
 
-class MockCFunctionComparatorNode : public MockNamedValueComparator
+class MockCFunctionComparatorAndCopierNode : public MockNamedValueComparator, public MockNamedValueCopier
```
basvodde: Any way we can avoid the use of multiple inheritance here? I think this is the only place in CppUTest and prefer not using it.

tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -909,6 +939,388 @@ TEST(MockSupportTest, customObjectWithFunctionComparatorThatFailsCoversValueToSt
     CHECK_EXPECTED_MOCK_FAILURE_LOCATION(failure, __FILE__, __LINE__);
 }
 
+TEST(MockSupportTest, customTypeOutputParameterSucceeds)
+{
```
basvodde: Please remove this kind of comments from the tests. (all of them).

If you want to seperate the AAA sections, just use an empty line. No need to name them.

basvodde:
Mostly good! Lets review some of the comments quickly and get it merged without splitting it in smaller ones.

src/CppUTestExt/MockSupport_c.cpp
arstrube: I couldn't think of any. I tried eliminating the extra list, but couldn't
Look at it as these nodes implementing both interfaces. They're nice, clean interfaces and the node implements them. If this was Java, no one would think twice about it.
If you can think of a better solution, I'd be interested. Especially if we didn't need this list at all :)

tests/CppUTestExt/MockSupportTest.cpp
arstrube: No problem. I didn't put them there, I just didn't take the trouble to remove them.

arstrube:
I'll remove the comments tomorrow.

arstrube:
It will be tough about the node:
+ I considered composition, but couldnt think of a good way
+ I considered getting rid of the list, but found no good replacement
+ Ironically, I think two lists would work. I like that less than implementing two interfaces, though.

tests/CppUTestExt/MockSupportTest.cpp
basvodde: Ok. Multiple inheritance is a *potential* problem for old or embedded compilers, so would be good if we can look for a different solution here :( Sorry about that.

basvodde: Ref: http://stackoverflow.com/questions/2929543/instance-where-embedded-c-compilers-dont-support-multiple-inheritance

arstrube:
Okay, the answer right at the end clinches it. Even if it is "only" Kernel development. My only option is the two list approach. I'm relying on you to reduce it to one list again :)

basvodde:
Then we'll need a list per class...

arstrube:
That is so. Unless you have a better idea, of course...

On the bright side, it was like that before I refactored it to use two interfaces :-). 

The stupid thing is that a list is needed only for two reasons:
1. Must keep track in order to delete the custom thingies
2. The c functions have a different type to the c++ "convenience" functions.

arstrube:
@basvodde any bright ideas yet? Or a decision on the lesser evil, perhaps?

basvodde:
Done.

PR #698, opened 2015-07-10
Update ArmCC platform-specific Code

arstrube:
Awaiting feedback... from Ivan Zrno

1. Added missing fp interfaces
2. Fixed somethings so they work with Keil on Arm
3. Fixed the order of everything, so its easier to compare to the other platform-specific versions (sorry, makes the PR a little more messy than need be)


arstrube:
Platform-specific timestring is a dummy; this will cause at least one test to fail, if I remember correctly. Maybe we can do better than that!

PR #700, opened 2015-07-14
Plugin to check for IEEE754 floating point exception flags automatically

arstrube:
I propose a plugins/ directory for contributed plugins.

It would be good to have Travis CI checks on this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ option(C++11 "Compile with C++11 support" OFF)
 option(TESTS "Compile and make tests for the code?" ON)
 option(TESTS_DETAILED "Run each test separately instead of grouped?" OFF)
 
+option(PLUGINS "Compile and make plugins?" ON)
+
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "What kind of build this is" FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
@@ -60,6 +62,10 @@ endif (EXTENSIONS)
 if (TESTS)
     add_subdirectory(tests)
 endif (TESTS)
+
+if (PLUGINS)
+    add_subdirectory(plugins/IEEE754ExceptionFlagsPlugin)
+endif (PLUGINS)
 
 set (INCLUDE_INSTALL_DIR "include")
 set (LIB_INSTALL_DIR "lib")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,5 +146,7 @@ Features configured in CppUTest:
     Compile and run self-tests          ${TESTS}
     Run self-tests separately           ${TESTS_DETAILED}
 
+    Compile plugin and run self-tests   ${PLUGINS}
+
 -------------------------------------------------------
 ")
--- a/plugins/IEEE754ExceptionFlagsPlugin/Makefile
+++ b/plugins/IEEE754ExceptionFlagsPlugin/Makefile
@@ -3,7 +3,7 @@ PLUGIN_HOME := .
 AR := ar
 CXX := g++
 CC := gcc
-CFLAGS := -g3 -Wextra -Wall -Werror \
+CFLAGS := -g3 -O0 -Wextra -Wall -Werror \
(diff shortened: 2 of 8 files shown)


arstrube:
Actually, it is also g++-only at this point. No idea how it works with VS or clang (yet).

arstrube:
Moved the plugin to a more appropriate location.

__Like:__
+ It is separate from actual CppUTest, which reflects that only people who really want this will include it in their tests
+ It has its own Makefile

__Dislike:__
+ It is not included in the Travis CI build

What would be needed to run the tests is something like (libCppUTest.a is expected in the usual place):
```bash
pushd ../plugins/IEEE754ExceptionFlagsPlugin
make check clean
popd
```
I wouldn't want to give it a separate target in the build matrix, because there would be such a big overhead for such a tiny test.

There is no point in running the examples, because all tests in the examples are tests that will fail on purpose. This is why they are in a separate demo 'test'.

@basvodde - your thoughts?

basvodde:
Why not just put it in CppUTestExt for now? Especially considering we'll be pulling out the CppUMock, it feels like the right place?

arstrube:
Okay. That would be fine with me. Also, we could run tests on Travis CI and it would be a lot easier for those who want to install it (no separate Makefile and lib).

arstrube:
If we do that, what do we do with the demo tests that fail (and are meant to fail, to demonstrate that no explicit check is necessary to pick up problems)?

There is at least one build that runs the examples/ tests and would fail if they do.

arstrube:
Oh. An even more important issue. The plugin has to be compiled with C++11, because this feature is only available in C++11. It can still pick up faults in C code compiled with other standards, but as far as I recall, CppUTest must compile without C++11 enabled. So actually, we cannt put it into extensions, right?

Unless we compile it conditionally.

arstrube:
So, I would imagene the intire plugin and its tests and demo must be within an `#ifdef` for C++``, and that should settle it, because to my knowledge the examples are not tested with C++11.

It has to be said though that once it has been compiled, C++11 is not a prerequisite.

basvodde:
Can't you change the demo test in a positive test :)

And yeah, I guess it needs to be in a #ifdef C++11 block, which is fine.

arstrube:
@basvodde how to proceed? The plugin requires C++11 to compile. Which is why I think it would not be a good idea to put it with CppU Extensions after all. Currently, I compile it separately with its own Makefile, and then simply link the library.

It has really proven itself useful working with floating point math, pinpointing potential problems.

To put it in extensions, we would need some mechanism to dis-/enable C++11 for individual sources. I don't know whether that is possible, or desirable.

Andne:
Do the demo/example tests need to be run?  Compiling them makes sense to make sure nothing related to the API changed, but what is gained by running them.  Another option would maybe be to wrap them running such that you want a return value greater than or equal to 0.  That would at least ensure that there wasn't a crash in the examples I think, assuming those cause a negative return value.

arstrube:
Hi Andy, the demo / example tests are not run by default. In order to build and run them, one needs to issue the command
```
make demo
```
I wouldn't run them as part of the build process; they are meant for the user to see how things work and perhaps experiment with it.

arstrube:
Oops. didn't mean to close...

The actual tests use a fixture, so they will never return a failure.  I am busy adding wrappers for feenableexcept() and fedisableexcept(). In order to test that, I might add tests that run in a separate process on those systems that support such things. Not sure yet.

Andne:
Ok, I misunderstood then, I thought you were saying that one build does run the demo tests which would be a problem with this plugin since you make the test fail in order to illustrate how it works.

basvodde:
Why not just put it under #if ?

arstrube:
@basvodde - something like "#if c++11 is configured", right? All plugin code and all tests would be wrapped by one #if?

I ran into nasty problems on MacOS.
1. It doesn't have `feenableexcept()` `fedisableexcept()`. There is public domain substitute code to take care of that. Another #if / #else....
2. On my Mac, I can't get the flags to work at all. If I use the public domain feenableexcept(), then it duly crashes. if not, nothing seems to happen. I found that you're supposed to use `#pragma STDC FENV_ACCESS ON`, however, with that, nothing happens either.

Issue #706, opened 2015-07-15
Overloaded functions are not being handled properly


jgonzalezdr:
Take the following test example:

```C++
int foo()
{
	return mock().actualCall("foo").returnIntValue();
}

int foo(int bar)
{
	return mock().actualCall("foo").withParameter("bar", bar).returnIntValue();
}

TEST(Example1, Test1)
{
	mock().expectOneCall("foo").andReturnValue(1);
	mock().expectOneCall("foo").withParameter("bar", 1).andReturnValue(2);

	int result = foo(1);

	CHECK_EQUAL(2, result);
	mock().checkExpectations();
}

TEST(Example1, Test2)
{
	mock().expectOneCall("foo").andReturnValue(1);
	mock().expectOneCall("foo").withParameter("bar", 1).andReturnValue(2);

	int result1 = foo(1);
	int result2 = foo();

	CHECK_EQUAL(1, result2);
	CHECK_EQUAL(2, result1);
	mock().checkExpectations();
}
```

For Test1 one would expect the get an error saying that the expected call to foo without parameters was not fulfilled, and for Test2 there should be no error. But instead, the following errors are thrown:

```
U:\cpputest-examples\Example1\Example1_Test.cpp:51: error: Failure in TEST(Example1, Test2)
	expected <1>
	but was  <2>
	difference starts at position 0 at: <          2         >
.
U:\cpputest-examples\Example1\Example1_Test.cpp:33: error: Failure in TEST(Example1, Test1)
	expected <2>
	but was  <1>
	difference starts at position 0 at: <          1         >
U:\cpputest-examples\Example1\Example1_Test.cpp:20: error: Failure in TEST(Example1, Test1)
	Mock Failure: Expected call did not happen.
	EXPECTED calls that did NOT happen:
		foo -> int bar: <1>
	ACTUAL calls that did happen (in call order):
		foo -> no parameters
```

jgonzalezdr:
With ignored parameters it's not going any better:
```C++
TEST(Example1, Test3)
{
    // Prepare

    mock().expectOneCall("foo").andReturnValue(1);
    mock().expectOneCall("foo::bar").ignoreOtherParameters().andReturnValue(3);

    // Exercise

    int result = foo(55);

    // Verify
    CHECK_EQUAL(3, result);
    mock().checkExpectations();
}
```

Instead of complaining about the unfulfilled call without parameters, that's the error output:

```
[...]

PR #707, opened 2015-07-15
Fixed errors when mocking overloaded functions

jgonzalezdr:
This PR fixes the problem detailed in #706.

The changes basically consist in not resetting the "actual call fulfillment state" of other potentially matching calls with more requisites (i.e. parameters) when the first fulfillment is done. 

Instead, each time a parameter is passed, if an expected call was previously fulfilled then it's "unfulfilled" (because it isn't actually matching the actual call), and other expected calls that could be previously fulfilled are also discarded. That way, as parameters are passed, expected calls with too few parameters are discarded, and the set of "fulfillable" expected calls is narrowed down.

Finally, when checkExpectations() is called on the last actual call, the "actual call fulfillment state" of the remaining unfulfilled expected calls is reset. This way, we've got again in pristine "unfulfilled" state all the unfulfilled calls that where not fulfilled by the last call, and they are ready to be fulfilled by the next call.

### DIFF
--- a/include/CppUTestExt/MockCheckedActualCall.h
+++ b/include/CppUTestExt/MockCheckedActualCall.h
@@ -97,9 +97,10 @@ protected:
     virtual void failTest(const MockFailure& failure);
     virtual void checkInputParameter(const MockNamedValue& actualParameter);
     virtual void checkOutputParameter(const MockNamedValue& outputParameter);
+    virtual void callIsInProgress();
 
     enum ActualCallState {
-        CALL_IN_PROGESS,
+        CALL_IN_PROGRESS,
         CALL_FAILED,
         CALL_SUCCEED
     };
--- a/include/CppUTestExt/MockExpectedCallsList.h
+++ b/include/CppUTestExt/MockExpectedCallsList.h
@@ -42,7 +42,7 @@ public:
     virtual int size() const;
     virtual int amountOfExpectationsFor(const SimpleString& name) const;
     virtual int amountOfUnfulfilledExpectations() const;
-    virtual bool hasUnfullfilledExpectations() const;
+    virtual bool hasUnfulfilledExpectations() const;
     virtual bool hasFulfilledExpectations() const;
     virtual bool hasFulfilledExpectationsWithoutIgnoredParameters() const;
     virtual bool hasUnfulfilledExpectationsBecauseOfMissingParameters() const;
@@ -53,7 +53,7 @@ public:
     virtual void addExpectedCall(MockCheckedExpectedCall* call);
     virtual void addExpectations(const MockExpectedCallsList& list);
     virtual void addExpectationsRelatedTo(const SimpleString& name, const MockExpectedCallsList& list);
-    virtual void addUnfilfilledExpectations(const MockExpectedCallsList& list);
+    virtual void addUnfulfilledExpectations(const MockExpectedCallsList& list);
 
     virtual void onlyKeepExpectationsRelatedTo(const SimpleString& name);
     virtual void onlyKeepExpectationsWithInputParameter(const MockNamedValue& parameter);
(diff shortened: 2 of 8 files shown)


jgonzalezdr:
With the following example test
```C++
int foo()
{
	return mock().actualCall("foo").returnIntValue();
}

int foo(int bar)
{
	return mock().actualCall("foo").withParameter("bar", bar).returnIntValue();
}
int foo(int bar, double baz)
{
    return mock().actualCall("foo").withParameter("bar", bar).withParameter("baz", baz).returnIntValue();
}


TEST(Example1, Test1)
{
	// Prepare

    mock().expectOneCall("foo").andReturnValue(1);
	mock().expectOneCall("foo").withParameter("bar", 1).andReturnValue(2);

    // Exercise

	int result = foo(1);

    // Verify
	CHECK_EQUAL(2, result);
	mock().checkExpectations();
}

TEST(Example1, Test2)
{
	// Prepare

	mock().expectOneCall("foo").andReturnValue(1);
	mock().expectOneCall("foo").withParameter("bar", 1).andReturnValue(2);

    // Exercise

	int result1 = foo(1);
	int result2 = foo();

    // Verify
	CHECK_EQUAL(1, result2);
	CHECK_EQUAL(2, result1);
	mock().checkExpectations();
}

TEST(Example1, Test3)
{
    // Prepare

    mock().expectOneCall("foo").andReturnValue(1);
    mock().expectOneCall("foo").ignoreOtherParameters().andReturnValue(3);

    // Exercise

    int result = foo(55);

    // Verify
    CHECK_EQUAL(3, result);
    mock().checkExpectations();
}

TEST(Example1, Test4)
{
    // Prepare

    mock().expectNCalls(3, "foo").andReturnValue(1);
    mock().expectOneCall("foo").ignoreOtherParameters().andReturnValue(3);

    // Exercise

    int result1 = foo();
    int result2 = foo(55, 1.0);

    // Verify
    CHECK_EQUAL(1, result1);
    CHECK_EQUAL(3, result2);
    mock().checkExpectations();
}
```
now the following errors are thrown, which is what is expected:
```
U:\cpputest-examples\Example1\Example1_Test.cpp:75: error: Failure in TEST(Example1, Test4)
	Mock Failure: Expected call did not happen.
		foo -> no parameters
		foo -> no parameters
		foo -> no parameters
		foo -> all parameters ignored

U:\cpputest-examples\Example1\Example1_Test.cpp:59: error: Failure in TEST(Example1, Test3)
	Mock Failure: Expected call did not happen.
		foo -> no parameters
		foo -> all parameters ignored

U:\cpputest-examples\Example1\Example1_Test.cpp:25: error: Failure in TEST(Example1, Test1)
	Mock Failure: Expected call did not happen.
		foo -> no parameters
		foo -> int bar: <1>
```

arstrube:
Looking good :).

Just one thing - the diff looks somehow funky. There are a lot of "changes" where nothing has changed, really. Like:
```c++
102 -    CALL_IN_PROGESS,    103 +     CALL_IN_PROGRESS,
````
I have no idea what went wrong, so unfortunately I don't know how to fix it, but if you could manage to get rid of those, it would be a lot clearer what actually changed ;-)

jgonzalezdr:
@arstrube You are missing a "R"... originally it was spelled PROG ESS... but it should have been PROG**R**ESS. I also fixed some other minor spelling errors like Ful**l**filment (with 2 Ls) or F**i**lfilment.

@basvodde If you want I can put the spelling fixes in a different PR to be merged so it may be clearer which are the changes in this one.

arstrube:
I guess the others have spelling errors too. For the life of me, I couldn't see them. Must have been tired or something.... sorry. I honestly couldn't see the changes :-(

jgonzalezdr:
Now it's much more cleaned up, I hope the PR can be reviewed more easily.

basvodde:
Ok, seems ok. We need to work on speeding up the mocking calls too though. I still have a IGNORE_TEST in the TestMockSupport.cpp which ought to become a performance test. This has become an issue in some code bases I worked with used the ExpectNCalls with a high number.

jgonzalezdr:
In fact I found this problem while working in another improvement: expected calls that can be "fulfilled" any number of times between a minimum and a maximum, to implement things like expectAtLeastOneCall or expectAnyCalls.

You can peek on my progress in this feature [here](https://github.com/jgonzalezdr/cpputest/tree/expected-calls-range).

As as side effect, this feature will improve at least memory issues and maybe improve speed, because only one instance of each expected call will exists no matter how many "N" calls are expected.

basvodde:
Excellent, that was exactly one thing I was thinking about.

arstrube:
Nice.

PR #711, opened 2015-07-16
Infinity for CL2000

arstrube:
Unfortunately it's mixed up with additional tests for failures, which involve infinity, and Simplestring. Sort-of happened along the way.... I should probably try and separate this from the CL2000 stuff?

This also fixes a failing test on armcc / iar along with the cl2000.


src/CppUTest/TestFailure.cpp
```diff
@@ -164,18 +164,20 @@ EqualsFailure::EqualsFailure(UtestShell* test, const char* fileName, int lineNum
     message_ = createButWasString(expected, actual);
 }
 
-static SimpleString StringFromOrNan(double d)
+static SimpleString StringFromOrNanOrInf(double d)
```
arstrube: Why not just `StringFromDouble(double d)`? After all, that's what it is.

src/Platforms/C2000/UtestPlatform.cpp
```diff
@@ -221,7 +221,7 @@ static int IsNanImplementation(double d)
 
 static int IsInfImplementation(double d)
 {
-    return (d == (1.0d/0.0d));
+    return 0;
```
arstrube: The cl2000 doesn't have NaN or Inf.

tests/CommandLineTestRunnerTest.cpp
```diff
@@ -161,15 +161,22 @@ TEST(CommandLineTestRunner, listTestGroupAndCaseNamesShouldWorkProperly)
...
 struct FakeOutput
 {
     FakeOutput() : SaveFOpen(PlatformSpecificFOpen), SaveFPuts(PlatformSpecificFPuts),
         SaveFClose(PlatformSpecificFClose), SavePutchar(PlatformSpecificPutchar)
     {
-        PlatformSpecificFOpen = fopen_fake;
-        PlatformSpecificFPuts = fputs_fake;
-        PlatformSpecificFClose = fclose_fake;
-        PlatformSpecificPutchar = putchar_fake;
+        PlatformSpecificFOpen = (FOpenFunc)fopen_fake;
```
arstrube: The cl2000 is fussy about assigning "C" functions to "C++" functions

tests/JUnitOutputTest.cpp
```diff
@@ -296,7 +296,7 @@ TEST_GROUP(JUnitOutputTest)
 
     void setup()
     {
-        UT_PTR_SET(PlatformSpecificFOpen, (PlatformSpecificFile(*)(const char*, const char*))mockFOpen);
```
arstrube: Spurious cast; I think it shouldn't have been there to start with. The cl2000 doesn't like it

Issue #709, opened 2015-07-16
Failed float/double cast prevents build on GCC 5.1


offa:
The current snapshot (b8f5ce0) fails to build by a conversion to float from double (`Werror=float-conversion)`.

 The cause is the function `int IsInfImplementation(double)` (`src/Platforms/Gcc/UtestPlatform.cpp:242:12`):

```cpp
static int IsInfImplementation(double d)
{
    return isinf(d); // Casting to d to float solves the issue
}
```
However, i'm not sure if this is cast safe.

--------------

`isinf()` is implemtend like this on GCC 5.1:

```cpp
#  define isinf(x) \
     (sizeof (x) == sizeof (float)					      \
      ? __isinff (x)							      \
      : sizeof (x) == sizeof (double)					      \
      ? __isinf (x) : __isinfl (x))
```

arstrube:
That is strange, for two reasons:

1. It's gcc, which means it already passed the build in various forms and contortions, including Clang (okay, I don't know the exact version of gcc on Travis CI -- I am using 4.8.3).
2. As you have pointed out, isinf() is implemented as a macro, designed to handle both float and double. No cast should actually occur.

That aside, the cast would be a very bad idea, as infinity is probably defined as the largest value that can be represented, which would be very different between float and double indeed.

arstrube:
I suggest you compile with -E to see whether the `isinf()` macro implementation you outlined above is indeed the one that `isinf()` resolves to (somehow I doubt it).

If the compiler lets you do:
```c++
int (*PlatformSpecificIsInf)(double) = isinf;
```
then it is a function implementation, not the one above.

arstrube:
P.S. If you don't need this functionality, as a quick and safe workaround you can always define a dummy that just returns 0 (false) always.

offa:
> That is strange, for two reasons:

For **three** reasons: It was compiling with version from yesterday (or possibly the day before yesterday ...). To be safe i'll check on another system.

Btw. if that matters, i have compiled with C++11 enabled (`C++11=ON`).

> It's gcc, which means it already passed the build in various forms and contortions, including Clang 

Yes, that's strange; btw. CI has *gcc 4.6.3*.

offa:
Tested on another system:

**GCC 5.1.0** fails just as
**Clang 3.6.2** fails also (but with another cause!?)

Both times C++11 enabled.

offa:
[...]

PR #712, opened 2015-07-16
Fix for issue #709 which caused a build-failure on GCC 5.1

offa:
See #709 for more details.


offa:
As discussed in #709, the solution is implemented in #713.

PR #713, opened 2015-07-16
Fix for compilation errors related to isinf()

### LINKED ISSUE #709: Failed float/double cast prevents build on GCC 5.1

jgonzalezdr:
Fixes #709

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -239,7 +239,7 @@ static int IsNanImplementation(double d)
 
 static int IsInfImplementation(double d)
 {
-    return isinf(d);
+    return isinf((float)d);
 }
 
 double (*PlatformSpecificFabs)(double) = fabs;


src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -239,7 +239,7 @@ static int IsNanImplementation(double d)
 
 static int IsInfImplementation(double d)
 {
-    return isinf(d);
+    return isinf((float)d);
```
arstrube: Yeah, lets do this

PR #714, opened 2015-07-16
Fixed spelling errors in function names and constants.

jgonzalezdr:
Added this so if it's merged then PR #707 changes will be more readable.


basvodde:
Kewl

PR #715, opened 2015-07-16
Move printing of Nan and Inf to StringFrom(double)

arstrube:
Since NaN and Infinity are full citizens of the floating point universe, I propose to move printing "Nan - Not a number" and "Inf - Infinity" directly into SimpleString.

Actually, under Gcc, StringFromFormat would print "Nan" and "Inf" in any case, so why not allow this little bit of extra verbosity outside of failures as well (actually I found it confusing that sometimes "Nan" would be printed and other times "Nan - Not a number".

This can also be tested very easily in SimpleString by having the platform-specific functions just return `true` always for those tests.

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -491,7 +491,12 @@ SimpleString HexStringFrom(const void* value)
 
 SimpleString StringFrom(double value, int precision)
 {
-    return StringFromFormat("%.*g", precision, value);
+    if (PlatformSpecificIsNan(value))
+        return "Nan - Not a number";
+    else if (PlatformSpecificIsInf(value))
+        return "Inf - Infinity";
+    else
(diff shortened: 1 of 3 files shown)


arstrube:
Appveyor is not feeling well :-(

arstrube:
@basvodde The Appveyor MinGW builds are still all failing (2015-07-17 14:20) because something goes wrong with 7z. So all PRs that are 'failing' on Appveyor (which is currently ALL PRs) are most probably good.

basvodde:
Nice refactoring.

What do we need to do to get the Appveyor to pass again?

arstrube:
Appveyor has been fixed. I can restart the Appveyor build on my fork (and did for this PR and it passed). For the actual PR, I suppose you being the owner of the repository, you would have to log into Appveyor and restart the build.

Issue #710, opened 2015-07-16
Building tests fails on Clang 3.6.2


offa:
After checking building of current snapshot (b8f5ce0) with clang (3.6.2) for #709; it seems clang fails to build the `UtestTest`: 

```
cpputest/tests/UtestTest.cpp:88:12: error: comparison of
      address of 'this->fixture' not equal to a null pointer is always true
      [-Werror,-Wtautological-pointer-compare]
    CHECK(&fixture != 0);
           ^~~~~~~    ~
```

This applies to C++11 enabled and disabled.

arstrube:
As far as I know, this code has been around for a long time. I have seen this error before, but can't remember where, unfortunately.

offa:
Fixed by #716.

PR #716, opened 2015-07-17
fix a compile error

lbjcom:
When I build in MacOS using following commands:

```
$cmake
$make
```

I get following error message:
> error: comparison of address of 'this->fixture' not equal to a null pointer is always true [-Werror,-Wtautological-pointer-compare]

The fixes the above error.

### DIFF
--- a/tests/UtestTest.cpp
+++ b/tests/UtestTest.cpp
@@ -85,7 +85,7 @@ TEST(UtestShell, PassedCheckEqualWillIncreaseTheAmountOfChecks)
 
 IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture)
 {
-    CHECK(&fixture != 0);
+    CHECK(&fixture != NULL);
 }
 
 TEST(UtestShell, MacrosUsedInSetup)


arstrube:
Yes, this warning was annoying me too. Strange though about the fix, I mean comparing it to NULL should be just as 'tautological', but seemingly the compiler doesn't catch it....

@basvodde what does this test actually prove? Couldn't it be done in a totally non-tautological way somehow?

basvodde:
The test name was pretty self-descriptive ;)

It tests you can access fixture in an IGNORED_TEST.

arstrube:
Yes - and no. What do you expect to happen, if the test were to fail? The fixture is not accessed through a pointer, so it will always be there, and it will always be non-zero. So, what is the problem that this test case guards against? Could we perhaps include that in the test name? I know I keep tripping over this, and so does Clang :). If I'm not entirely mistaken, if there really was a problem, the test wouldn't even compile, right?

basvodde:
Correct, it is one of these tests that won't fail but won't compile if it doesn't work.

offa:
Related: #710 (fixed by this PR).

PR #718, opened 2015-07-18
Check macros with custom messages

jgonzalezdr:
Added the missing complementary macros for all the current assert macros. 

The new ones accept a user message that is printed if the assertion fails (like CHECK_TEXT or CHECK_TRUE_TEXT), and follow the same nomenclature (CHECK_xxx_TEXT).

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -91,115 +91,157 @@
 // Different checking macros
 
 #define CHECK(condition)\
-  CHECK_LOCATION_TRUE(condition, "CHECK", #condition, __FILE__, __LINE__)
+  CHECK_TRUE_LOCATION(condition, "CHECK", #condition, "", __FILE__, __LINE__)
 
 #define CHECK_TEXT(condition, text) \
-  CHECK_LOCATION_TEXT(condition, "CHECK", #condition, text, __FILE__, __LINE__)
+  CHECK_TRUE_LOCATION(condition, "CHECK", #condition, text, __FILE__, __LINE__)
 
 #define CHECK_TRUE(condition)\
-  CHECK_LOCATION_TRUE(condition, "CHECK_TRUE", #condition, __FILE__, __LINE__)
+  CHECK_TRUE_LOCATION(condition, "CHECK_TRUE", #condition, "", __FILE__, __LINE__)
+
+#define CHECK_TRUE_TEXT(condition, text)\
+  CHECK_TRUE_LOCATION(condition, "CHECK_TRUE", #condition, text, __FILE__, __LINE__)
 
 #define CHECK_FALSE(condition)\
-  CHECK_LOCATION_FALSE(condition, "CHECK_FALSE", #condition, __FILE__, __LINE__)
+  CHECK_FALSE_LOCATION(condition, "CHECK_FALSE", #condition, "", __FILE__, __LINE__)
 
-#define CHECK_LOCATION_TEXT(condition, checkString, conditionString, text, file, line) \
-    { UtestShell::getCurrent()->assertTrueText((condition) != 0, checkString, conditionString, text, file, line); }
+#define CHECK_FALSE_TEXT(condition, text)\
+  CHECK_FALSE_LOCATION(condition, "CHECK_FALSE", #condition, text, __FILE__, __LINE__)
 
-#define CHECK_LOCATION_TRUE(condition, checkString, conditionString, file, line)\
-  { UtestShell::getCurrent()->assertTrue((condition) != 0, checkString, conditionString, file, line); }
+#define CHECK_TRUE_LOCATION(condition, checkString, conditionString, text, file, line)\
+  { UtestShell::getCurrent()->assertTrueText((condition) != 0, checkString, conditionString, text, file, line); }
 
-#define CHECK_LOCATION_FALSE(condition, checkString, conditionString, file, line)\
-  { UtestShell::getCurrent()->assertTrue((condition) == 0, checkString, conditionString, file, line); }
+#define CHECK_FALSE_LOCATION(condition, checkString, conditionString, text, file, line)\
+  { UtestShell::getCurrent()->assertTrueText((condition) == 0, checkString, conditionString, text, file, line); }
 
 //This check needs the operator!=(), and a StringFrom(YourType) function
-#define CHECK_EQUAL(expected,actual)\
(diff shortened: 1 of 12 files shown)


jgonzalezdr:
I was wondering why some assert methods in UtestShell have a last parameter that accepts a terminator (with default value), while others don't.

I can't see its purpose, as these methods are called through macros and they never set this argument it in any case, so they could be removed altogether and let the terminator be chosen in the call to failWith. I think that would cleanup the interface a bit.

What do you think?

arstrube:
I think you are right.

arstrube:
Essentially, the macro above is one that should fail, but not terminate the test (as the test has already been destroyed at the point where this macro is executed).

jgonzalezdr:
The reasons for adding custom messages to failures are several:
*  When using CppUTest to perform integration or differential tests, many times multiple cycles of execution + verification are executed in a loop, so it's necessary to known in with iteration the verification failed.
* To document the purpose of the check (which sometimes isn't evident to collaborators) and improve reviews of failures in continuous integration: a message saying "wrong calculation of XXX: expected 5 but was 7" is more meaningful to people not familiarized with that specific module than just "expected 5 but was 7" preceded by an obscure reference to a line of code, specially when changes in a module break something in another module maintained by someone else.

jgonzalezdr:
Regarding the first reason, I'll explain the use case for which I had to implement these additional macros:

A customer wanted us to integrate in a device an algorithm to calculate the state of charge of batteries, so they gave us an implementation in C. But that implementation used an interface with the system (get configuration parameters, read battery voltage from an ADC, store internal state, etc.) that wasn't compatible with our software architecture, so we had to adapt it.

To ensure that we didn't introduce any modifications to the calculations, we've written differential tests that check that both version give the same results under the same stimuli, effectively executing them in a loop that simulates a profile of input conditions. To do that, CppUMock is used to mock the interfaces of the algorithm with the rest of the system.

It has been then necessary when checks failed to know in what iteration of the loop that happened, to be able to go directly there and debug. That allowed us for example to catch a bug in the original implementation related to casting from int to char...

jgonzalezdr:
Regarding the second reason, in all the (well-managed) java projects in which I've been involved, it was mandatory that junit asserts had meaningful messages for failures, and issues were raised in CI if the version without text was used. I've always found this a good programming practice, specially regarding maintenance and collaboration in large teams, and I've always missed that in CppUTest.

arstrube:
Okay, that makes sense.

I suppose once it is there, we can find ways to use it ;-). One thing that comes to mind is some developers I work with who do calculations involving large arrays of floats (typically 64k). They check this by running DOUBLES_EQUAL() in a loop They could then use this feature to determine directly which one of the 64k floats (e.g. element `result [7][0][134]` failed the test.

CppU also includes a feature for running tests in a loop, which will print out which loop you are currently in and it can also be retrieved via `getCurrentRepetition()`. But I guess you know that already ;-).

arstrube:
> Regarding the second reason, in all the (well-managed) java projects in which I've been involved, it was mandatory that junit asserts had meaningful messages for failures, and issues were raised in CI if the version without text was used.

Oh, I see. Kewl. We are still working on making them give meaningful names to the tests ;-)...

arstrube:
Oh and thanks for the nice use cases. That always helps me understand a new feature that I didn't happen to wish for myself :-)

jgonzalezdr:
@arstrube Thanks for you example macro, now I understand that the terminator parameters are useful to write custom checks that don't terminate the test.

So I suppose it would be a good idea to add this parameter to all assert methods.

arstrube:
Actually, I don't know. This happens very, very rarely. Only when you do checks from a plugin, I suppose.

arstrube:
I guess removing the terminator parameter like you did makes more sense at present (its presence didn't seem to help any in my case).

jgonzalezdr:
@arstrube In fact I didn't remove (I think...) any terminator parameter, I left the ones that were already there. I just noticed that some asserts didn't had that parameters, that's why I was asking why they are there, because maybe we can remove them of all the calls, or the other way, add the ones that are missing. 

Maybe @basvodde has a bit more insight on this.

arstrube:
Hmm. Offhand, I would say there are special terminators for:
* Calls to be used from C code
* Mocks
* Terminating without exceptions
But then, @basvodde probably has more insight into this, like you said.

basvodde:
@jgonzalezdr The terminators were introduced to allow to use assertions in C code without throwing an exception. This is because throwing an exception when you are executing C code has undetermined behavior and in some gcc version actually crashed the program. This relates to bug that I think @simshi ones reported, but I can't find it anymore.

The reason why some do have the Terminator and some don't is that the ones that do have C equivalents and the one who don't are probably not used in the C version of the macros.

include/CppUTest/UtestMacros.h
```diff
@@ -91,115 +91,157 @@
...
   { \
-    UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, file, line); \
+    UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, "", file, line); \
   } }
 
 //This check checks for char* string equality using strcmp.
 //This makes up for the fact that CHECK_EQUAL only compares the pointers to char*'s
-#define STRCMP_EQUAL(expected,actual)\
-  STRCMP_EQUAL_LOCATION(expected, actual, __FILE__, __LINE__)
+#define STRCMP_EQUAL(expected, actual)\
```
basvodde: Can you, in the future, try to keep the irrelevant changes (such as this additional space) out of the diffs :)

basvodde:
The PR looks pretty good and I think I'd like to merge it. However, it does look like the added macros weren't ever used, not even in the tests.

Would you mind to try to test-drive your added code to CppUTest ;)

Before merging it, I'd like some tests (afterwards :( ) to the TestUtestMacros so that the macros are actually used within CppUTest itself (in the tests that is)

include/CppUTest/UtestMacros.h
jgonzalezdr: Sorry, I had my diff viewer configured to ignore whitespace changes and didn't notice that... :-(

jgonzalezdr:
I'll add these tests soon so you can merge.

jgonzalezdr:
@basvodde I've just added the missing tests for the new macros, I expect them to be enough.

basvodde:
Thanks!

PR #721, opened 2015-07-20
Added missing mock_c() withMemoryBufferParameter() interfaces

arstrube:
@basvodde -- while this does work as is, I am not really sure whether I added enough, or too much. 

Should this be part of `sMockValue_c`?  If so, one could use it as a return value. Should one be able to do that?

Are the tests sufficient? Or too many? I haven't used these interfaces recently, so I am not too sure.


basvodde:
Seems fine.

PR #725, opened 2015-07-27
Fixed minor formatting error in MockCheckedExpectedCall::callToString()

jgonzalezdr:


jgonzalezdr:
It seems that there are problems again with the MinGW downloads from SourceForge in Appveyor...

PR #726, opened 2015-07-28
Coveralls online code coverage

jgonzalezdr:
I was messing around with [Coveralls](http://coveralls.io) (which if you don't know is an online service that allows tracking code coverage, it's free for open source projects, and really easy to use because you can sign-in with the github account), and I added support to automatically update CppUTest coverage from Travis-CI builds.

It doesn't support branch coverage, but anyway it may be useful to check in a glance PR's proper testing, so I'm PRing this just in case you find it useful.

In case this is merged, remember to enable coverage for the cpputest repo in [Coveralls](http://coveralls.io),

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - BUILDTOOL=autotools
   - BUILDTOOL=cmake
   - BUILDTOOL=cmake-coverage
+#  - BUILDTOOL=autotools_old_compilers
   global:
   - secure: |-
       P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: cpp
 compiler:
-#- clang
+- clang
 - gcc
 rvm:
 - 1.9.3
 env:
   matrix:
-#  - BUILDTOOL=autotools
-#  - BUILDTOOL=cmake
+  - BUILDTOOL=autotools
+  - BUILDTOOL=cmake
   - BUILDTOOL=cmake-coverage
   global:
   - secure: |-
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,22 @@
 language: cpp
 compiler:
-- clang
+#- clang
 - gcc
 rvm:
 - 1.9.3
 env:
   matrix:
-  - BUILDTOOL=autotools
-  - BUILDTOOL=cmake
-  - BUILDTOOL=autotools_old_compilers
+#  - BUILDTOOL=autotools
+#  - BUILDTOOL=cmake
+  - BUILDTOOL=cmake-coverage
   global:
   - secure: |-
       P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
       vA1NkcW49tQW1wQvBlRtdlLNOmUfDP/oiJFXPwNn4dqwOIOEet2P7JO/5hnH
       MNHlZmGu2WpoZREhOFBfsIhK0IP8mloqLDq2XemBdga/LWygrLU=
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -8,13 +8,15 @@ env:
   matrix:
   - BUILDTOOL=autotools
   - BUILDTOOL=cmake
-  - BUILDTOOL=autotools_old_compilers
```
basvodde: Why was this one removed?

arstrube: blame?

arstrube:
I like it. Line coverage is the best that can be done for Cpp sources with the Gnu tool chain. That's why practically all tools based on it (notably Lcov) don't support anything more by default. In the case of plain gcov or lcov you can get branch coverage, but the results are confusing and not very useful for C++.

.travis.yml
jgonzalezdr: Because the travis build script (scripts/travis_ci_build.sh) has nothing to execute when BUILDTOOL=autotools_old_compilers, therefore it's totally useless and a waste of ci resources.

I might be missing something,  but it seems that the script for this case was removed at some point in the past,  and this rule is a leftover,  so I thought it was a good time to remove it.

basvodde: It was the opposite. The script should have still been added :) The idea of the target is to get an old gcc compiler and compile it with that one, or even a bunch of old compilers. The problem is in implementation, which is I guess it would need to be done by downloading the compiler source, compile the compiler and then run CppUTest. I'm afraid travis people wouldn't like that (and haven't contacted them about it yet), so I left the target there as a reminder to continue that work.

Now the reminder is deleted :P

jgonzalezdr: OK, I'll reverse that... O:-)

Regarding the old compilers thing, they can be just targzipped an left in any URL (e.g. as a fake github release), then just downloaded and uncompressed in the working directory.

jgonzalezdr: BTW, I could re-add it but leave it commented to make ci builds faster, isn't it?

arstrube: Yeah. But the old-compilers "build" is the only one that's always really fast :P

basvodde: Commented out is fine. Downloading binaries or binary packages was what I was thinking of indeed. But then I got distracted and started working on something else and didn't come back to it yet.

I figure that if more people report bugs related to old compilers, I'll get motivated again :)

jgonzalezdr:
I forgot to comment that I added back the old compilers target (commented out).

jgonzalezdr:
@basvodde Remember to activate the coveralls account signing in with the cpputest github account, then enable the cpputest repository (Click "Repos" on the left menu, then "Add Repo" button on the right side, then set the cpputest repo to "ON").

image

image

arstrube:
So cool :)

I can see immediately that you still need a test case for that elipsis ('...') in SimpleString ;-)

It actually shows the change in coverage, so that for a given pull, it's very easy to see what changed for better or for worse.

arstrube:
Err. not so cool:Travis says in the Gcov build:
```
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
```
All existing targets pass. What's going on?

jgonzalezdr:
I suppose it's because in the travis account a personal SSL key has not been generated, but it doesn't matter at all because the coveralls server doesn't authenticate the incoming connection.

arstrube:
But it does matter in that it fails the Travis build. Can we make it not fail?

jgonzalezdr:
I say that it doesn't matter because the fail is not due to that error,  but rather to the corresponding repo not existing in coveralls (or user account not being configured). Fixed that last point in #740.

arstrube:
Ahh. That fix will be great.

arstrube:
@basvodde now that we have a separate -- and much more useful -- build for coverage, should'nt we perhaps remove the old gcov target from the autotools mega-build?

basvodde:
I suggest we'll keep it :) I think it is good to have gcov in the build so that the target works.

PR #727, opened 2015-07-28
Length specifier fixed for C++98

offa:
The length specifier `hh` was [introduced with C99](http://www.cplusplus.com/reference/cstdio/fprintf/) and is not supported by C++98.

Therefore compiling CppUTest with C++11 *disabled* fails.

--------

Tested with GCC 5.2 and `cmake -DC++11=OFF`.


offa:
That's strange, while gcc 5.2 (with C++11 off) compiles now, [clang fails](https://travis-ci.org/cpputest/cpputest/jobs/72998042#L413).

The solution could be a cast to `unsigned short` - if that's ok, i'll insert it.

offa:
Cast added in d2d3c8d.

offa:
There seems an error with the last travis build; it runs [successfuly](https://travis-ci.org/offa/cpputest/builds/73005700) on the branch after manual restarting the build.

PR #729, opened 2015-07-28
Fix the const char * thing (again?)

### LINKED ISSUE #728: Clang failure because of const to non-const cast
Building with clang 3.6.2 fails with this error:

```
tests/CppUTestExt/MockSupport_cTestCFile.c:38:20: error: cast from 'const void *' to 'char *' drops const qualifier [-Werror,-Wcast-qual]
    return (char*) object;
                   ^
```
The error is caused by the `MockTypeValueToStringFunction_c` implementation, which requires a return value of `char*`.

```
static char* typeNameValueToString(const void* object)
{
    return (char*) object;
}
```

arstrube:
This fixes #728.

### DIFF
--- a/examples/AllTests/MockDocumentationTest.cpp
+++ b/examples/AllTests/MockDocumentationTest.cpp
@@ -179,9 +179,9 @@ static  int equalMethod(const void* object1, const void* object2)
     return object1 == object2;
 }
 
-static char* toStringMethod(const void*)
+static const char* toStringMethod(const void*)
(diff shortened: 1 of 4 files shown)


offa:
There some missing change of return type: https://travis-ci.org/cpputest/cpputest/jobs/73047436#L5589

arstrube:
Mmh. Ran into this one before, as well :-(. It's the examples....

PR #734, opened 2015-07-29
Fix problem arising from 32-bit vs. 64-bit long

arstrube:
This is not an elegant solution, but the only one that seems to work.

The problem is that, on systems where 32 bit is an `unsigned long`, we need to cast to `unsigned long` whereas on 64 bit systems, we mustn't.


src/CppUTest/MemoryLeakDetector.cpp
```diff
@@ -66,7 +66,7 @@ void SimpleStringBuffer::addMemoryDump(const void* memory, size_t memorySize)
 	size_t p;
 
     while (currentPos < memorySize) {
-        add("    %04lx: ", currentPos);
+        add("    %04hx: ", (unsigned short) currentPos);
```
jgonzalezdr: Assuming that memory dumps will always be under 64k is a bit risky, isn't it?

IMHO a selection if the actual format by means of a "if(sizeof(unsigned long) == 4)" is preferable.

arstrube: We can do that. I was assuming that if a memory dump was that big, it would probably be hard to glean the important information from anyway. I did find elsewhere that a selection based on sizeof() is pretty much the only sensible way to deal with these things (short of using features we are not supposed to use).

arstrube:
@jgonzalezdr -- this is getting really frustrating. The preprocessor can't use sizeof(), and using sizeof() in code leads to code that will never be executed on any given compiler :-( 

What a no-win :p

src/CppUTest/MemoryLeakDetector.cpp
```diff
@@ -66,7 +68,7 @@ void SimpleStringBuffer::addMemoryDump(const void* memory, size_t memorySize)
 	size_t p;
 
     while (currentPos < memorySize) {
-        add("    %04lx: ", currentPos);
+        add("    %04lx: ", MAYBE_16BIT(currentPos));
```
basvodde: Whao. That doesn't look good :)

tests/TestUTestMacro.cpp
```diff
@@ -1222,7 +1225,7 @@ IGNORE_TEST(UnitTestMacros, BITS_EQUAL_TEXTWorksInAnIgnoredTest)
 } // LCOV_EXCL_LINE
 
 #if CPPUTEST_USE_STD_CPP_LIB
-static void _failingTestMethod_NoThrowWithCHECK_THROWS()
+static void failingTestMethod_NoThrowWithCHECK_THROWS_()
```
basvodde: Why the added _ ?

src/CppUTest/MemoryLeakDetector.cpp
arstrube: But it does work on all system :p

tests/TestUTestMacro.cpp
arstrube: Moved it from front to back. Just to trigger errored Travis CI. But then, in many other places, it's a postfix not a prefix, so there is some sense in it :)

src/CppUTest/MemoryLeakDetector.cpp
arstrube: @basvodde  take a look at the commit history. All of the more straightforwad "solutions" were rejected by Clang. For good reasons, even. I got really frustrated, and this is the epitome of portable code :D. So please, feel free to better it, but it does have to work!

(Technically, even this contains dead code, but the compiler doesn't see it)

arstrube: As it turns out, it didn't even work :-( If the `lx` stays, then we need to pass it an unsigned long with the high word masked away, or it will display random data in the high word.

src/CppUTest/MemoryLeakDetector.cpp
```diff
@@ -58,6 +58,8 @@ void SimpleStringBuffer::add(const char* format, ...)
     va_end(arguments);
 }
 
+#define CAST_FOR_16BIT_SYSTEMS(n) (unsigned long)(n)
+
```
arstrube: The macro is for documentation. I could use a comment (yikes!) instead.

jgonzalezdr:
@arstrube I know, each time I get caught in integer size issues,  Kernighan's ears begin to burn...

I wouldn't care putting it in actual code, the compiler's optimization should remove the dead code. It's not as elegant as precompiling it out, but standard C/C++ doesn't give a better solution.

arstrube:
@jgonzalezdr I guess the cast is actually quite elegant in this case. It turns out that %04lx always takes a long value at the given address, so the cast to `long` is essential on a 16 bit system to prevent garbage from showing up. On 32-bit, or 64-bit systems, the cast doesn't do any harm, as long as we are not trying to display 64-bit values under Windows.

To use a sizeof()-based solution, one would have to selectively turn of the dead-code-warning, which probably means \#pragma, and is probably fine, if you need to consider only one compiler, but it could soon turn into a \#pragma nightmare of its own.

arstrube:
This turned out a bit of a mess. The hopefully "good" fixes can be found in #730 and make this PR obsolete.

PR #737, opened 2015-07-31
Simplified MockSupportTest fixture in preparation for splitting

basvodde:
Now it'll be easier to split. Needs to be shrinked more still.

### DIFF
--- a/tests/CppUTestExt/MockSupportTest.cpp
+++ b/tests/CppUTestExt/MockSupportTest.cpp
@@ -435,8 +435,8 @@ TEST(MockSupportTest, expectOneMemBufferParameterAndValueFailsDueToContents)
 
     MockNamedValue parameter("parameter");
     parameter.setMemoryBuffer( memBuffer2, sizeof(memBuffer2) );
-    addFunctionToExpectationsList("foo")->withParameter("parameter", memBuffer1, sizeof(memBuffer1));
-    MockUnexpectedInputParameterFailure expectedFailure(mockFailureTest(), "foo", parameter, *expectationsList);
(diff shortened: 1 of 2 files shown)


tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -32,85 +32,67 @@
...
+  MockExpectedCallsListForTest expectations;
+  MockFailureReporterInstaller failureReporterInstaller;
+
+  void teardown()
+  {
+    mock().checkExpectations();
+    CHECK_NO_MOCK_FAILURE();
+  }
+};
+
```
arstrube: Okay, that's less stuff. Do you propose duplicating this in every TEST_GROUP, or stuffing it into a fixture? Duplicate the fixutre in each file, or use a header?

(Same would wokr with TEST_BASE, of course :P)

basvodde: You mean the 2 lines in teardown? Actually not all tests need these. Only the tests that check for MockFailure need the Reporter...

arstrube: err, no. I mean everything. That includes the member variables. I mean, I am not dead against duplicating them. But we would need them once per test group (some maybe not), right?

basvodde: Yes, you need them ones per testgroup. But not for all.

arstrube: That wouldn't be too bad then. Also save us from creating a header.

arstrube:
Looks good. If you want to merge it, I'll work from there.

arstrube:
> Now it'll be easier to split. Needs to be shrinked more still.

Take it to a shrink, perhaps ~.~ ?

PR #738, opened 2015-07-31
16-Bit: Fix unsigned (long) 16-bit vs 64-bit problem

arstrube:
The macro serves as documentation; I could do a comment instead ;-)


basvodde:
Why not just add the cast? Why the #define?

arstrube:
I thought I'd answered that question in my description above :-/

I am happy to do just the cast, but then I would want a comment, because most people don't think of 16-Bit targets, and they'll just kindof go: Why the hell cast? and want remove it. So if you could live with
```c++
add("    %04lx: ", (unsigned long) currentPos); // cast for 16-bit system
```
I'd be more than happy to change it.

basvodde:
Yeah, I prefer that, especially as we're adding the 16bit in travis in the future.

arstrube:
okay; will do

basvodde:
Can you remove the comment?

The cast isn't just for 16bit systems, right? It is just to properly match the types. It doesn't really matter it is 16 bit or 8bit or perhaps 256bits in the future. Or am I misunderstanding this?

arstrube:
I guess you're right.

arstrube:
Done.

PR #744, opened 2015-08-01
Add Dos/Platform.h and checks for defined(__X86__)

arstrube:
Not sure that one should add Platform.h when it isn't actually used yet -- but there it is.

The defines are there because we can't fork under DOS ;-)

And this Dos/ vs. dos/ thing is killing me -- there should be no dos/UtestPlatform.cpp in this pull! It wasn't in my local diff! Yikes!!!


arstrube:
misfit !

PR #745, opened 2015-08-01
Add Dos/Platform.h and checks for defined(__X86__)

arstrube:
Not sure that one should add Platform.h when it isn't actually used yet -- but there it is.

The defines are there because we can't fork under DOS ;-)


arstrube:
I would like to replace all those individual checks with `ifndef HAVE_FORK`. This works fine with Automake, Visual Studio and generally all external IDE's and build systems. On those systems, HAVE_FORK won't be defined, and one can simply define it, if one needs fork() and the system supports it.

However, I couldn't figure out how to get cmake to correctly set HAVE_FORK. So the cmake-based builds fail. Ideas, anyone? @jgonzalezdr, perhaps?

arstrube:
Will open new PR as soon as I have cmake sorted out.....

PR #746, opened 2015-08-01
Renamings

arstrube:
As agreed upon.


arstrube:
Uhh... sorry. There should have been something about 31 bytes max length in the commit messages.

basvodde:
Thanks!

I'll need the VS update also when splitting the MockSupport. You saw the first splitted PR, right?

arstrube:
Yup, I saw the splitted PR. The basic principle is fine. In the end though, I need to be able to split roughly like I outlined. So right now, there's no point trying to merege it in my watcom branch. 

I think it will work; execpt for the third file -- will remain to be seen whether it can fit with eithr of the other two. I'd like to avoid having seven executables. Six is bad enough ;-)

PR #747, opened 2015-08-02
Remove redundant Platforms/Xxx/Platform.h

arstrube:
Was: Add Dos/Platforms.h.


basvodde:
Do we need a Platform.h?

arstrube:
No, not at this point. But it seems all platforms have one, whethere they need it or not?

basvodde:
So far, we've attempted to avoid depending from platform independent code to platform dependent code (in #include). So, guess the others can safely be removed?

arstrube:
Yes. The only one that's actually used is the C2000 one.

arstrube:
We only need stdint.h `c2000/stdint.h`.

arstrube:
Had to remove the hidden tentacles in Visual Studio 201x projects.

Let's say we are actively discouraging the use of `Platform.h` henceforth :-)

arstrube:
@basvodde anything wrong with this?

basvodde:
No, was just focusing on MockSupport :)

PR #748, opened 2015-08-02
Use HAVE_FORK

arstrube:
After a lot of trial and error, finally I got the check for fork() to work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,7 @@ else()
 endif()
 
 set (CMAKE_REQUIRED_INCLUDES unistd.h)
-check_function_exists(fork HAVE_FORK) 
+check_function_exists(fork HAVE_FORK)
 
 message("
 -------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CppUTestRootDirectory}/cmake/Module
 
 include("${CppUTestRootDirectory}/cmake/Modules/CppUTestConfigurationOptions.cmake")
 include(CTest)
+include(CheckFunctionExists)
 include("${CppUTestRootDirectory}/cmake/Modules/CppUTestBuildTimeDiscoverTests.cmake")
 
 configure_file (
@@ -104,6 +105,9 @@ else()
     " please update cmake to version which provides CMakePackageConfighelpers module"
     " or write generators for CppUTestConfig.cmake by yourself.")
 endif()
+
+set (CMAKE_REQUIRED_INCLUDES unistd.h)
+check_function_exists(fork HAVE_FORK) 
 
 message("
 -------------------------------------------------------
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -53,7 +53,7 @@
 static jmp_buf test_exit_jmp_buf[10];
 static int jmp_buf_index = 0;
 
-#ifdef __MINGW32__
+#ifndef HAVE_FORK
 
 static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, TestPlugin*, TestResult* result)
 {
--- a/tests/UtestPlatformTest.cpp
+++ b/tests/UtestPlatformTest.cpp
@@ -35,7 +35,7 @@ TEST_GROUP(UTestPlatformsTest_PlatformSpecificRunTestInASeperateProcess)
     TestTestingFixture fixture;
 };
 
-#if defined(__MINGW32__) || defined(_MSC_VER) || defined(__IAR_SYSTEMS_ICC__) || defined(__ARMCC_VERSION) || defined(__TMS320C2000__)
+#ifndef HAVE_FORK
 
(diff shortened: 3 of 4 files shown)


arstrube:
... or so I thought, because it work on my Mac. Not so under Ubuntu. Being able to check CMakeError.log would probably make things a whole lot easier...

arstrube:
Great. So we have:
```
Determining if the function fork exists failed with the following output:
Change Dir: /home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp
Run Build Command:/usr/bin/make "cmTryCompileExec/fast"
/usr/bin/make -f CMakeFiles/cmTryCompileExec.dir/build.make CMakeFiles/cmTryCompileExec.dir/build
make[1]: Entering directory `/home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp'
/usr/bin/cmake -E cmake_progress_report /home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp/CMakeFiles 1
Building C object CMakeFiles/cmTryCompileExec.dir/CheckFunctionExists.c.o
/usr/local/clang-3.4/bin/clang   -include ;/home/travis/build/arstrube/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h"  -Weverything -Wall -Wextra -Wshadow -Wswitch-default -Wswitch-enum -Wconversion -Wsign-conversion -Wno-padded -Wno-disabled-macro-expansion -Werror -Wstrict-prototypes  -DCHECK_FUNCTION_EXISTS=fork" -I/home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp/unistd.h    -o CMakeFiles/cmTryCompileExec.dir/CheckFunctionExists.c.o   -c /usr/share/cmake-2.8/Modules/CheckFunctionExists.c
clang: error: argument to '-include' is missing (expected 1 value)
clang: error: no input files
/bin/sh: 1: /home/travis/build/arstrube/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h  -Weverything -Wall -Wextra -Wshadow -Wswitch-default -Wswitch-enum -Wconversion -Wsign-conversion -Wno-padded -Wno-disabled-macro-expansion -Werror -Wstrict-prototypes  -DCHECK_FUNCTION_EXISTS=fork: not found
make[1]: *** [CMakeFiles/cmTryCompileExec.dir/CheckFunctionExists.c.o] Error 127
make[1]: Leaving directory `/home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp'
make: *** [cmTryCompileExec/fast] Error 2
```

basvodde:
Sure this one also works on automake? Or does it now always turn fork off?

arstrube:
Automake does just fine. Which is of course a problem regarding the distributable config.h, because it will have HAVE_FORK set...

basvodde:
As long as the HAVE_FORK isn't used in the header, then it isn't a problem (yet).

arstrube:
Err... you merged this one? It's kind of broken still..... It will lead to HAVE_FORK being set in config.h. I could use the old code with a check for __X86__ added to it.

As for config.h, I was thinking of combining it with CPPUTEST_USE_FORK, in such a way that only if both are defined, running in a separate process will be available. Then we could define CPPUTEST_USE_FORK for all but the final automake test run.

arstrube:
@basvodde 

> As long as the HAVE_FORK isn't used in the header, then it isn't a problem (yet).

I'm not quite sure I understand. Which header? Used where?

arstrube:
The cmake problem with Travis appears to be a bug. Cmake does:
```bash
 -include ;... -I/home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp/unistd.h
```
but it should be:
```bash
 -include /home/travis/build/arstrube/cpputest/cpputest_build/CMakeFiles/CMakeTmp/unistd.h
```

arstrube:
I suspect the version I have on my Mac is newer.

arstrube:
My Mac uses Cmake 3.2.0-c2.
The Travis-CI build uses Cmake 2.8.7.

It doesn't look like there is an easy way to fix it in 2.8.7. So those builds will not recognize the presence of fork() and compile without.

basvodde:
@arstrube Sorry for merging this one too early.

My idea has always been to let the config.h be renamed to CppUTestConfig.h and generate that one. Then in make install, it would install the generated CppUTestConfig.h. So the HAVE_FORK ought to become CPPUTEST_HAVE_FORK or so and should be generated by automake.

I though this was working now as you seemed to manually set HAVE_FORK ? Anyways, how do we proceed?

arstrube:
No problem. Automake sets HAVE_FORK appropriately, so that's why it works.

What this doesn't do is give the user a chance NOT to have fork() when it is available. Is that an actual usecase? If so, we could use CPPUTEST_HAVE_FORK.

basvodde:
It is more than I wouldn't want the code to conflict with another #define. The HAVE_FORK is relatively safe as it doesn't occur in the header (yet). But when code appears in the header, then setting or changing the #define would cause a different header with the binary which would cause CppUTest to crash and it would be hard to debug.

For that reason, I don't want to use the default names (and neither the default name for config.h)

arstrube:
Frankly, I don't completely understand the ins and outs of this. Please feel free to change. My only concern is that after the change, it still be working :)

PR #749, opened 2015-08-02
Extracted another file out of the MockSupportTest. Getting there.

basvodde:
More


arstrube:
Keep going :)

arstrube:
I just merged the refactored files into my watcom branch. I get 585k + 205k executable size, which is a bit suspicious, since it used to be much more:

1. MockSupportTest now already fits into one exe, which is good.
2. I don't like dropping "Support" from the names. It is now very difficult to know which tests relate to MockSupport.cpp. Can you please keep it? It is also difficult to know which tests I must include in the new builds ("anything I think I haven't seen before", which is sooo vague).

3.

arstrube:
My apologies that I didn't see this earlier. It only struck me now that I'm trying to apply your changes.

arstrube:
The good thing is that from now on, I can use origin/master again for my watcom/Dos tests :)

I am not really happy with the namings. As long as we are using "Mock" as a general prefix, this is not helpful. If we didn't, then tests for the mock() "class" would maybe make more sense, but then perhaps MockSupport.cpp should become Mock.cpp? To be honest, I stil find "MockSupport" more appropriate, because the actual class is called that.

basvodde:
But these tests aren't testing MockSupport, they are just testing mocking. Mapping test names to class names is IMHO a bad habit and this is an opportunity to map them to the actual functionality :)

arstrube:
Bad habit, huh? But it helps finding stuff more quickly... ^^

arstrube:
Also, it makes total sense when you do old-fashioned stuff such as unit testing...

basvodde:
Not really, as the MockSupport tests never related to MockSupport but to whole of CpUMock :) So, it would be hard to find the tests, unless you knew you need to look in MockSupport

PR #750, opened 2015-08-03
Cmake and HAVE_FORK working properly at last

arstrube:
Now also for Cmake, when fork() is available, HAVE_FORK is defined and CppUTest is built so that running tests in a separate process will work. 

(The previous pull request did not really work for Cmake; the build would always be without support for running in a separate process, and HAVE_FORK would never be defined).


arstrube:
P.S. There are 5 more tests run when fork() is available, currently 562 as opposed to 557 tests.

PR #754, opened 2015-08-06
Put 16-bit Build Dummy Into Travis

arstrube:
At this point, it's just a dummy stiil.

What it does do
+ Install Dosbox
+ Invoke the appropriate build
+ Run the proper Makefile
+ Run Dosbox & produce dummy test output from within to a log
+ Print the log

What's still missing
+ Watcom compiler install
+ Proper implementation of Makefile
+ Script to run the actual 6 targets in Dosbox
+ Fail the test if "Failure" is found in the log.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,7 @@ builds/*gcc*
 /.deps
 CppUTestExtTests
 CppUTestTests
-Makefile
+/Makefile
 config.h
 config.h.in
 config.status
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,5 @@ before_script:
 - wcl --version
 - export CPPUTEST_BUILD_DIR=$TRAVIS_BUILD_DIR/cpputest_build
 - mkdir -p $CPPUTEST_BUILD_DIR && cd $CPPUTEST_BUILD_DIR
-
 script:
 - "../scripts/travis_ci_build.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,35 @@
 language: cpp
+os:
+ - linux
 compiler:
 - clang
 - gcc
+- wcl
 rvm:
 - 1.9.3
 env:
-  matrix:
   - BUILDTOOL=autotools
+  - BUILDTOOL=autotools_old_compilers
   - BUILDTOOL=cmake
   - BUILDTOOL=cmake-coverage
-#  - BUILDTOOL=autotools_old_compilers
-  global:
-  - secure: |-
+  - BUILDTOOL=make-dos
+matrix:
+    exclude:
+    - env: BUILDTOOL=cmake-coverage
+    - env: BUILDTOOL=autotools_old_compilers
+    - env: BUILDTOOL=make-dos
+    - compiler: wcl
+    include:
+    - env: BUILDTOOL=cmake-coverage
+      compiler: gcc
+    - env: BUILDTOOL=make-dos
+      compiler: wcl
+global:
+- secure: |-
       P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
       vA1NkcW49tQW1wQvBlRtdlLNOmUfDP/oiJFXPwNn4dqwOIOEet2P7JO/5hnH
       MNHlZmGu2WpoZREhOFBfsIhK0IP8mloqLDq2XemBdga/LWygrLU=
-  - secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
+- secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
 before_install:
 - sudo pip install cpp-coveralls
 install:
@@ -27,3 +41,4 @@ before_script:
 - mkdir -p $CPPUTEST_BUILD_DIR && cd $CPPUTEST_BUILD_DIR
 script:
 - "../scripts/travis_ci_build.sh"
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,12 @@ install:
 - sudo apt-get update --fix-missing
 - sudo apt-get install valgrind
 before_script:
+- wget ftp://ftp.openwatcom.org/pub/open-watcom-c-linux-1.9 -O /tmp/watcom.zip
+- mkdir -p watcom && unzip -a -d watcom /tmp/watcom.zip && sudo chmod -R 755 watcom/binl
+- export PATH=$PATH:$PWD/watcom/binl/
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -1,28 +1,47 @@
...
 compiler:
 - clang
 - gcc
+- wcl
 rvm:
 - 1.9.3
 env:
-  matrix:
   - BUILDTOOL=autotools
+  - BUILDTOOL=autotools_old_compilers
```
basvodde: Is this a merge error? Earlier this was commented out....

.gitignore
```diff
@@ -51,7 +51,7 @@ builds/*gcc*
 /.deps
 CppUTestExtTests
 CppUTestTests
-Makefile
+/Makefile
```
basvodde: Howcome?

.travis.yml
```diff
@@ -1,28 +1,47 @@
...
+global:
+- secure: |-
       P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
       vA1NkcW49tQW1wQvBlRtdlLNOmUfDP/oiJFXPwNn4dqwOIOEet2P7JO/5hnH
       MNHlZmGu2WpoZREhOFBfsIhK0IP8mloqLDq2XemBdga/LWygrLU=
-  - secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
+- secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
 before_install:
 - sudo pip install cpp-coveralls
+- sudo apt-get install dosbox -y
```
basvodde: Will this install it for all targets?

.gitignore
basvodde: Ah this is because dos has a Makefile, I understand.

.travis.yml
arstrube: Yes, and that is dumb, but if I install using the install: section, there doesn't seem to be another way :-(. Same goes for Coveralls, etc.

arstrube:
@basvodde Maybe take a look at the actual Travis output to see what's going on. Specifying the build matrix had to be done in a pretty round-about way, if you ask me (excluding most combinations first, then including the two we actually want).

arstrube:
What this PR doesn't show (because I rebased the branch) is that it took me about 20 times or so doing it wrong before I got it right :)

.travis.yml
arstrube: It's now part of env: not part of matrix:

.travis.yml
basvodde: Can you install it in the script instead, so that it doesn't need to be done for each target? I thought coveralls was needed for all. Also, I assume dosbox is bigger

basvodde:
Ok, let me look at how this matrix works then :)

.travis.yml
arstrube: I can try. Dosbox is really small, but Open Watcom weighs in at 80mb. Coveralls is just needed for one particular build, so that could be done via the script.

I'll leave that until after I got things to work properly.

arstrube: Dosbox will not be installed properly if I don't mention it in before_install: . I can move the Open Watcom stuff, though.

.travis.yml
```diff
@@ -1,23 +1,37 @@
...
+global:
+- secure: |-
       P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
       vA1NkcW49tQW1wQvBlRtdlLNOmUfDP/oiJFXPwNn4dqwOIOEet2P7JO/5hnH
       MNHlZmGu2WpoZREhOFBfsIhK0IP8mloqLDq2XemBdga/LWygrLU=
-  - secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
+- secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
 before_install:
-- sudo pip install cpp-coveralls
+- sudo apt-get install dosbox
```
basvodde: The dos-box comments are gone. But could you help me understand why it couldn't be in the script?

arstrube:
I have reverted the attempt to move installation to the script; weird and timeconsuming-to-fix errors occur when I do.

.travis.yml
basvodde: Any reason for not being installed properly? It sounds strange as it would do the same thing I'd assume. Just curious.

arstrube: If you can manage to make it work, be my guest. I have wasted too many hours on this already :P

basvodde: Well, I don't want to waste the same hours, so I'm curious at what went wrong?

arstrube: How about I resurrect a branch-less head of mine and you take a look at that. Essentially, dosbox isn't installed correctly, it won't start with 'dosbox' like it should. As for Watcom, it couldn't find its headers. In both cases, the exact same commands were used for the install.

arstrube: btw moving works fine for Coveralls.

.travis.yml
```diff
@@ -1,28 +1,47 @@
...
+  - BUILDTOOL=autotools_old_compilers
   - BUILDTOOL=cmake
   - BUILDTOOL=cmake-coverage
-#  - BUILDTOOL=autotools_old_compilers
-  global:
-  - secure: |-
+  - BUILDTOOL=make-dos
+matrix:
+    exclude:
+    - env: BUILDTOOL=cmake-coverage
```
basvodde: Why exclude this one?

basvodde: Oh, nevermind, stupid me.

.travis.yml
```diff
@@ -1,28 +1,47 @@
...
+- secure: Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAwNY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
 before_install:
 - sudo pip install cpp-coveralls
+- sudo apt-get install dosbox -y
 install:
 - gem install travis_github_deployer
 - sudo apt-get update --fix-missing
 - sudo apt-get install valgrind
 before_script:
+- wget ftp://ftp.openwatcom.org/pub/open-watcom-c-linux-1.9 -O /tmp/watcom.zip
```
basvodde: Why not move this inside the travis build script? I think you misunderstood this earlier. Now it will be on all targets, right? But if you add it to the travis_ci_build.sh file, then you can specify it for only the Watcom build.

basvodde:
Merging this anyways as I want to play with travis.

arstrube:
Oops. Did you get the latest stuff?

PR #757, opened 2015-08-08
Pulled out some more

basvodde:
And more


arstrube:
This is missing from both CMakeLists and Makefile.am => These tests are currently not run at all!

arstrube:
P.S. I got onto this because coverage dropped in some places where I knew it shouldn't have.

basvodde:
But there is no new file :)

arstrube:
touché :P

PR #758, opened 2015-08-08
First Stab At Building and Running real CPPU1.EXE

arstrube:
Built the first test executable.


arstrube:
Rebased on #761.

basvodde:
This one needs a rebase too ?

arstrube:
Very much so. Many conflicts. Hopefully got it all right.

basvodde:
The diff looks fine

arstrube:
I noticed the log is more verbose now (it echoes commands), which is often desirable, but there are a few places where I'd rather not have it (e.g. printf 'bla' is irritating when repeated).

I suppose I use '@' ? Did you change this on purpose, or did it happen accidentally?

basvodde:
Yeah, it is the x in set -ex. I thought a bit more verbosity would help us.

PR #761, opened 2015-08-09
Install Coveralls, DOSBox and Open Watcom from script

arstrube:
You missed the best :-)

Results of a sloeepless night. It's working well.


basvodde:
I was trying to get coveralls out of sudo because of: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

arstrube:
Oh, sorry. Perhaps we can get into that after finishing this one? 

It seems to me that we can get stuff EITHER out of sudo, OR out of travis.yml. But then, I haven't looked at your pr yet....

basvodde:
It is a bit more complicated. The sudo for dosbox I'll leave, but I'm wondering about the coveralls sudo pip. Still playing with it.

arstrube:
Anything I must change in this pull to be compatible with what you are trying to achieve?

arstrube:
Neither dosbox nor coveralls seems to be on the whitelist, so they would be rejected? Wouldn't we need to raise an issue also to include those?

basvodde:
Coveralls is resolved. So, only dosbox would be needed. Could you rebase the branch?

arstrube:
Those matrix entries spawned a lot of new offspring :-(

basvodde:
What do you mean? I separated out some of the build targets..

arstrube:
I know, I know. But they tend to be triggered sequentially, and make-dos is the last one :-(

basvodde:
It is last because most of the time, that is what you want ;)

PR #763, opened 2015-08-10
Changes for Keil project with a STM32 target

matdfg:
Hi,

I created a new UtestPlatform.cpp file for Keil project targeting a STM32. It would be nice if we can add this file to the project.

In TestFailureNaNTest.cpp I had a compiler issue. "0.0 / 0.0" returned 0 and not as expected NAN. I did contact the armcc compiler developer.

In UtestPlatformTest.cpp the test "DummyFailsWithMessage" failed. Is this test correct for a single process target? I didn't to a lot of debugging and set the test ignored. 

Regards,

Matt


arstrube:
Please compare `UtestPlatform.cpp` for C2000. It has similar NAN handling.

src/Platforms/Keil/UtestPlatform.cpp
```diff
@@ -0,0 +1,196 @@
...
+extern "C" void* (*PlatformSpecificMalloc)(size_t) = malloc;
+extern "C" void* (*PlatformSpecificRealloc) (void*, size_t) = realloc;
+extern "C" void (*PlatformSpecificFree)(void*) = free;
+extern "C" void* (*PlatformSpecificMemCpy)(void* s1, const void* s2, size_t size) = memcpy;
+extern "C" void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
+
+static int IsNanImplementation(double d)
+{
+    return isnan(d);
+}
```
arstrube: Probably better return 0, if the system does not consider 0.0f / 0.0f NAN.

tests/TestFailureNaNTest.cpp
```diff
@@ -37,7 +37,7 @@ const char* failFileName = "fail.cpp";
 
 static double zero = 0.0;
 static double one = 1.0;
-static const double not_a_number = zero / zero;
+static const double not_a_number = NAN;
 static const double infinity = one / zero;
 
 extern "C" {
```
arstrube: Changing the test like this is not acceptable because it introduce platform specific ( `\<math.h\>` ) into CppUtest code. The above change will heal this failure.

tests/UtestPlatformTest.cpp
```diff
@@ -36,8 +36,7 @@ TEST_GROUP(UTestPlatformsTest_PlatformSpecificRunTestInASeperateProcess)
 };
 
 #ifndef HAVE_FORK
-
-TEST(UTestPlatformsTest_PlatformSpecificRunTestInASeperateProcess, DummyFailsWithMessage)
+IGNORE_TEST(UTestPlatformsTest_PlatformSpecificRunTestInASeperateProcess, DummyFailsWithMessage)
 {
     fixture.registry_->setRunTestsInSeperateProcess();
     fixture.runAllTests();
```
arstrube: Please, fix your implementation of `PlatformSpecificRunTestInASeperateProcess()` to use the proper dummies (found in a number of other UtestPlatform.cpp implementations) rather than ignoring this test.

tests/TestFailureNaNTest.cpp
arstrube: If you have done everything correctly, there should be no failing tests.

matdfg:
Arstrube,

Thank for your feedbacks. I will do some rework and push again.

arstrube:
Thanks. Much appreciated!

arstrube:
Although the other 2 functions should never be called (and will never be called by CppUTestTests, unless HAVE_FORK is defined), it might be safer to provide dummies for all three functions:
```c++
static void DummyRunTestInASeperateProcess(UtestShell* shell, TestPlugin* plugin, TestResult* result)
{
    result->addFailure(TestFailure(shell, "-p doesn't work on this platform, as it is lacking fork.\b"));
}

static int DummyPlatformSpecificFork(void)
{
    return 0;
}

static int DummyPlatformSpecificWaitPid(int, int*, int)
{
    return 0;
}

void (*PlatformSpecificRunTestInASeperateProcess)(UtestShell*, TestPlugin*, TestResult*) = DummyRunTestInASeperateProcess;
int (*PlatformSpecificFork)() = DummyPlatformSpecificFork;
int (*PlatformSpecificWaitPid)(int, int*, int) = DummyPlatformSpecificWaitPid;
```
N.B. These don't need `extern "C"`
N.B.2 None of the function pointer assignments should need "extern", unless this is a really cranky compiler.

matdfg:
Arstrube,

Reworking the src\Platforms\Keil\UtestPlatform.cpp solved already the "DummyFailsWithMessage" test.

Actually UnitTests still fails with problems with NAN and INFINITY. If use the MicroLib option from the Keil compiler the debugger gives following values:
```
// file TestFailureNaNTest.cpp line 38 - 41
static double zero = 0.0;
static double one = 1.0;
static const double not_a_number = zero / zero; // = 0
static const double infinity = one / zero; // = 0
´´´
Because not_a_number and infinity has both 0 my UnitTests will fail. Any ideas how I could solve that in a for the community compatible way? 

Here the failing UnitTests:
```
TEST(TestFailureNanAndInf, DoublesEqualThresholdIsInf)
..\..\..\..\cpputest\tests\TestFailureNaNTest.cpp:108: error: Failure in TEST(TestFailureNanAndInf, DoublesEqualThresholdIsInf)
        expected <expected <1>
        but was  <Nan - Not a number> threshold used was <Inf - Infinity>
        Cannot make comparisons with Nan>
        but was  <expected <1>
        but was  <Nan - Not a number> threshold used was <Nan - Not a number>
        Cannot make comparisons with Nan>
        difference starts at position 64 at: <used was <Nan - Not >
                                                        ^

 - 0 ms
TEST(TestFailureNanAndInf, DoublesEqualActualIsInf)
..\..\..\..\cpputest\tests\TestFailureNaNTest.cpp:101: error: Failure in TEST(TestFailureNanAndInf, DoublesEqualActualIsInf)
        expected <expected <1>
        but was  <Inf - Infinity> threshold used was <3>>
        but was  <expected <1>
        but was  <Nan - Not a number> threshold used was <3>
        Cannot make comparisons with Nan>
        difference starts at position 24 at: <but was  <Nan - Not >
                                                        ^

 - 10 ms
TEST(TestFailureNanAndInf, DoublesEqualExpectedIsInf)
..\..\..\..\cpputest\tests\TestFailureNaNTest.cpp:94: error: Failure in TEST(TestFailureNanAndInf, DoublesEqualExpectedIsInf)
        expected <expected <Inf - Infinity>
        but was  <2> threshold used was <3>>
        but was  <expected <Nan - Not a number>
        but was  <2> threshold used was <3>
        Cannot make comparisons with Nan>
        difference starts at position 10 at: <expected <Nan - Not >
                                                        ^

 - 0 ms
```

**Comment by arstrube** [comment]
@matdfg you didn't close this on purpose, right?

I'll take a look at the code. It should pass.

File: src/Platforms/Keil/UtestPlatform.cpp [hunk 4]
```diff
@@ -0,0 +1,218 @@
...
+    void (*PlatformSpecificFlush)() = PlatformSpecificFlushImplementation;
+    void* (*PlatformSpecificMalloc)(size_t) = malloc;
+    void* (*PlatformSpecificRealloc) (void*, size_t) = realloc;
+    void (*PlatformSpecificFree)(void*) = free;
+    void* (*PlatformSpecificMemCpy)(void* s1, const void* s2, size_t size) = memcpy;
+    void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
+
+    static int IsNanImplementation(double d)
+    {
+#       ifdef __MICROLIB
```
arstrube: ... so `__MICROLIB` does NOT handle NAN?

Either way, the version that returns 0 should be enabled with CppUTest

arstrube:
Looking _much_ better, anyway.

You need to define `__MICROLIB` when compiling CppUTest for the tests to work correctly.

src/Platforms/Keil/UtestPlatform.cpp
matdfg: Yes __MICROLIB doesn't handle NAN and INF.

__MICROLIB is a compiler option for smaller and faster code. For my production code I will use MicroLib option. To support both variants I added this precompiler switch.

Because the compiler returns for infinity and not_a_number both 0. The StringFrom function returns in all cases "Nan - Not a number".

´´´
// SimpleString.cpp line 492
SimpleString StringFrom(double value, int precision)
{
    if (PlatformSpecificIsNan(value))
        return "Nan - Not a number";
    else if (PlatformSpecificIsInf(value))
        return "Inf - Infinity";
    else
        return StringFromFormat("%.*g", precision, value);
}
´´´

matdfg:
__MICROLIB is defined. I see that with the debugger.

The proplem is that I have in TestFailureNaNTest.cpp for not_a_number and infinity both 0. Other toolchains maybe make her a difference. In function StringFrom from file SimpleString.cpp I will have always "Nan - Not a number" as return value because of that.

arstrube:
Strange. The C2000 is exactly like that. I also fixed the infinity-problem quite a while ago (we didn't use to consider infinity in the test before then). Are you sure you are working off the current HEAD?

arstrube:
See the PR to your fork. I think that way will actually be better for all. It was only a matter of time until we'd run into this problem.....

matdfg:
I plan to do a sync with the master today.

arstrube:
I actually rebased it on master on my fork, but undid it again, as it showed a gazillion of irrelevant changes in my PR to your fork. At your end, it should sync with no problem.

basvodde:
Is this ready to be merged now?

matdfg:
Yes, would be okay for me.

PR #768, opened 2015-08-11
Add function pointer support.

jerryryle:
Mixing function and data pointers isn't allowed when compiling in pedantic mode. Attempting to use the CppUTest pointer assertions or mock support causes the following compiler errors:
clang: "error: cast between pointer-to-function and pointer-to-object is an extension [-Werror,-Wpedantic]"
gcc: "error: ISO C++ forbids casting between pointer-to-function and pointer-to-object [-Werror=pedantic]"

This patch adds explicit function pointer support. Sorry it's huge. It touches quite a few files.

### DIFF
--- a/include/CppUTest/SimpleString.h
+++ b/include/CppUTest/SimpleString.h
@@ -133,6 +133,7 @@ private:
 
 SimpleString StringFrom(bool value);
 SimpleString StringFrom(const void* value);
+SimpleString StringFrom(void (*value)());
 SimpleString StringFrom(char value);
 SimpleString StringFrom(const char *value);
 SimpleString StringFromOrNull(const char * value);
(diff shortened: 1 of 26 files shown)


jerryryle:
The AppVeyor failure seems to be a timeout. Is there a way to kick it off again without updating the pull request?

arstrube:
We can't restart it on cpputest/cpputest. But you can restart it on your own fork, if you log in.

arstrube:
You could also push some trivial change. But that would suck.

jerryryle:
Hahah. I know. I was trying to avoid the trivial change. : ) I'll try kicking it off on my own fork. I'm fairly certain it'll pass now, but want to be sure.

basvodde:
Interesting. It isn't a problem to cast different types of function pointers? Otherwise we'll keep adding different signatures of function pointer forever :)

jerryryle:
Hah. I know, that was my concern at first too. I think the pedantic warning exists only because data and code aren't interchangeable on some platforms. Compilers seem fine with casting code to code and data to data (different types and signatures) and only the cross-cast causes the warning. We've been using these function pointer features for a while now to allow us to compile in pedantic mode without problems, so I figured I'd see if you'd like to take them.

basvodde:
Ok, makes sense. Thanks!

jerryryle:
You bet! Thanks for taking it.

PR #771, opened 2015-08-15
Different implementation of the Failable Allocator

basvodde:
Proposed. A simpler implementation. @jkohvakk I do think we lack some tests though, but not sure, I just refactored the implementation.


jkohvakk:
I think I like this implementation by @basvodde more than mine :-). I had some trouble with dynamic memory allocation (do not remember what) and went for C-arrays. But this linked list is much better. Also, pushing the logic of single alloc failure to LocationToFailAllocNode makes lots of sense (and because it is in a list now, the implementation does not need to be in the header).

PR #775, opened 2015-08-17
Get rid of some more overridden empty destructors

arstrube:
Found some more unneeded empty overridden destructors while checking Coveralls report.


basvodde:
Uhm, sorry. It is unmergable :)

arstrube:
I resolved the conflicts.

PR #778, opened 2015-08-23
Improve performance of MockExpectedCallsList

mmartalo:
For large number (e.g. 10000) of mocked calls, the simple linked list
was a performance bottleneck.
This patch brings two improvements:
- tracking the last element of the linked list (for O(1) insertion)
- deleting NULL elements as they are marked, as opposed to rescanning
  the list from scratch after marking.
Profiling with callgrind shows that the linked list structure is no
longer the performance bottleneck.

I know that having a huge number of mocked/expected calls is debatable, but it did happen to matter in my use case. Basically, I'm doing checks of the type: perform an operation, and assert that a certain log message is displayed. This means that MockExpectedCallsList has to scale to the number of logs printed by the operation.

As you see from the Callgrind results for this particular test case, addExpectedCall was spending most of its time in self(). Where the vast majority of this time went was into looping through the linked list in order to find the tail, so that it can insert the new element.

Profiling overview before patch:
callgrind results before patch
Profiling before patch, zoom in on addExpectedCall:
callgrind zoom in on addexpectedcall
Profiling overview after patch:
callgrind results after patch

### DIFF
--- a/src/CppUTestExt/MockExpectedCallsList.cpp
+++ b/src/CppUTestExt/MockExpectedCallsList.cpp
@@ -28,8 +28,9 @@
 #include "CppUTest/TestHarness.h"
 #include "CppUTestExt/MockExpectedCallsList.h"
 #include "CppUTestExt/MockCheckedExpectedCall.h"
+#include <cassert>
 
-MockExpectedCallsList::MockExpectedCallsList() : head_(NULL)
+MockExpectedCallsList::MockExpectedCallsList() : head_(NULL), tail_(NULL)
 {
 }
 
@@ -40,6 +41,7 @@ MockExpectedCallsList::~MockExpectedCallsList()
         delete head_;
         head_ = next;
     }
+    tail_ = NULL;
 }
 
 bool MockExpectedCallsList::hasCallsOutOfOrder() const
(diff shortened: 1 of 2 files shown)


src/CppUTestExt/MockExpectedCallsList.cpp
```diff
@@ -140,74 +141,116 @@ void MockExpectedCallsList::addExpectations(const MockExpectedCallsList& list)
         addExpectedCall(p->expectedCall_);
 }
 
+// Tracking the parent is needed because we need to know it in order to remove an element
+// during the iteration.
+// parent must not advance if p has been deleted (we achieve this by setting p=parent at the end
+// of pruneEmptyNodeFromList(), and letting the "parent = p" execute normally in the for loop).
+//
+// You need to close the curly bracket after the processing step.
+#define ITERATE_WITH_PREV_AND_NEXT \
```
mmartalo: I think keeping the iteration algorithm in one place is good, but I'm ok with inlining this everywhere if you find it ugly.

PREV should be changed to PARENT

arstrube:
Great! 

Do you suppose you could do the same without using Std C Lib assert()? The thing is that CppUTest shouldn't depend on it, and the inclusion of assert.h breaks that.

mmartalo:
Oh ok, I didn't know that. I was wondering how assert.h could be missing :)
I removed the asserts. I left them as comments, I think they help reading the code, but I can remove them completely if they seem overkill.

arstrube:
There you go. About the comments / asserts, let's see what @basvodde has to say. Looking forward to this being merged....

mmartalo:
Any news on this? If the PR needs changes, please let me know this week, as I'll be unavailable for a couple of weeks afterwards.

basvodde:
Sorry! I won't have time to merge it this week, but I'll definitively will request some small changes. That said, I'm pretty sure eventually it'll be merged as this was one problem that needed to be solved. So thanks for that and sorry for the slow response during these days.

PR #787, opened 2015-10-25
Fix CppUTest for OSX / Xcode7 / clang-700.x

charlesnicholson:
Add -Wno-reserved-id-macro and -Wno-keyword-macro to clang builds. Fixes CppUTest on Xcode 7+, with clang-700+, where warnings were added about using reserved namespaces like __, and for macros that shadow keywords.

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -189,10 +189,16 @@ ifeq ($(COMPILER_NAME),$(CLANG_STR))
 # -Wno-global-constructors Wno-exit-time-destructors -> Great warnings, but in CppUTest it is impossible to avoid as the automatic test registration depends on the global ctor and dtor
 # -Wno-weak-vtables -> The TEST_GROUP macro declares a class and will automatically inline its methods. Thats ok as they are only in one translation unit. Unfortunately, the warning can't detect that, so it must be disabled.
 # -Wno-old-style-casts -> We only use old style casts by decision
+	CPPUTEST_CXX_WARNINGFLAGS += -Weverything -Wno-disabled-macro-expansion -Wno-padded -Wno-global-constructors -Wno-exit-time-destructors -Wno-weak-vtables -Wno-old-style-cast
+	CPPUTEST_C_WARNINGFLAGS += -Weverything -Wno-padded
+
+# Clang "7" (Xcode 7 command-line tools) introduced new warnings by default that don't exist on previous versions of clang and cause errors when present.
+ifeq ($(findstring clang-7,$(CC_VERSION_OUTPUT)),clang-7)
 # -Wno-reserved-id-macro -> Many CppUTest macros start with __, which is a reserved namespace
 # -Wno-keyword-macro -> CppUTest redefines the 'new' keyword for memory leak tracking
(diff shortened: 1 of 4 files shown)


basvodde:
Thanks. Only CMake still missing, perhaps?

charlesnicholson:
Happy to pitch in.

I thought I added CMake support over in `cmake/Modules/CppUTestWarningFlags.cmake` in this PR, and the builds look green, but if I missed it or am simply oblivious, sorry and please let me know what else I should do.

Additionally, I'm uncomfortable with the output from https://travis-ci.org/cpputest/cpputest/jobs/87264997:

Line 1059 shows warning spam because the flags are added and that version of clang doesn't understand them:
```
clang++ -DHAVE_CONFIG_H -I. -I..   -include ../include/CppUTest/MemoryLeakDetectorMallocMacros.h -I ../include     -include ../include/CppUTest/MemoryLeakDetectorNewMacros.h  -Weverything -Wall -Wextra -Wshadow -Wswitch-default -Wswitch-enum -Wconversion -pedantic -Wsign-conversion -Woverloaded-virtual -Wno-disabled-macro-expansion -Wno-padded -Wno-reserved-id-macro -Wno-keyword-macro -Wno-global-constructors -Wno-exit-time-destructors -Wno-weak-vtables -Wno-old-style-cast   -g -O2 -MT lib_libCppUTest_a-CommandLineArguments.o -MD -MP -MF .deps/lib_libCppUTest_a-CommandLineArguments.Tpo -c -o lib_libCppUTest_a-CommandLineArguments.o `test -f 'src/CppUTest/CommandLineArguments.cpp' || echo '../'`src/CppUTest/CommandLineArguments.cpp
warning: unknown warning option '-Wno-reserved-id-macro'
      [-Wunknown-warning-option]
warning: unknown warning option '-Wno-keyword-macro'; did you mean
      '-Wno-keyword-compat'? [-Wunknown-warning-option]
2 warnings generated.
```

But, line 1010 shows the checks for the flags succeeding:
```
checking whether CC and CXX supports -Wno-reserved-id-macro... yes
checking whether CC and CXX supports -Wno-keyword-macro... yes
```

I suspect the issue might be the behavior of CMake's `CheckC[XX}CompilerFlag` command, which in the docs over at https://cmake.org/cmake/help/v3.0/module/CheckCXXCompilerFlag.html doesn't indicate that it will compile with `-Werror` or require that a compilation emit no warnings. If compilation succeeds with warnings, does CMake think that the option is supported?

charlesnicholson:
Also, for posterity, my addition:

`ifeq ($(findstring clang-7,$(CC_VERSION_OUTPUT)),clang-7)`

to the makefile is pretty janky, since all future versions of clang will likely support these options. I didn't see the makefile depend on external parsing tools (perl / sed / awk / etc), and I figured it was better to get things working right now.

It feels, though, like the right thing to do is check the major version number of Apple Clang and add these flags if it's >= 7.0.0. That requires some fancy parsing and comparison, though, and I wanted to ask a maintainer for their thoughts first.

basvodde:
I'm not an expert in CMake. Usually the CMakefiles are maintained by others, so I guess the changes will come :) I noticed the autotools files were done already. Appreciated!

SaeedYasin:
As predicted by @charlesnicholson now with clang v9, I got the same issue, pull request open here #1164 to make it future proof for all new versions.

PR #789, opened 2015-11-04
Add integration into TeamCity, plus a few trivial commits

devMichaelJones:
The last commit in this series is the bulk of it.  This enables running the tests with '-o teamcity', which outputs the test results in a format that TeamCity understands, enabling it to present eg. graphical information about the tests.


basvodde:
Kewl. Could you try to make smaller pull requests in the future. Especially the combinations of typos and features makes it hard to check and merge them. I'll merge this one anyways, since it looks good (except for the ugly str[0] etc part)

devMichaelJones:
I thought it would suffice to have the typos and features in separate commits.  Why is it harder to check and merge them if they're in one pull request?  If, for example, you didn't like one of the commits, couldn't you just request that I re-submit the pull request without it?  Or submit it with changes?

Thank you for the merge.

Issue #107, opened 2013-05-03
Additional filters


basvodde:
Add -xg and xn to exclude tests based on some character string in
the name or group.

PR #790, opened 2015-11-10
Additional filters to exclude tests based on group and/or name

apbarrero:
Implementation for issue #107.

New command line options:

`-xg <group>`: exclude test groups matching `group` string. 
`-xn <name>`: exclude test names matching `name` string.
`-xsg <group>`: exclude test group strictly matching `group` string.
`-xsn <name>`: exclude test names strictly matching `name` string.

Working use cases are:

* Exclude one test group.
* Exclude one test name.
* Exclude one test name within a group.
* Filter one test name in all but one excluded group.
* Exclude one test name in all but one excluded group.

Since test filter combination is implemented as a set union rather than intersection, combining two or more non-overlapping exclude group options, i.e. `-xg group1 -xg group2` turns out to run the full list of tests including all groups. Same thing happens when combining multiple exclude name options.

In order to have excluding combinations working differently, it'd be required a new command line option to specify whether filtering options should be combined with logical AND or OR operations. IMHO, that'd be too much convoluted.

### DIFF
--- a/include/CppUTest/CommandLineArguments.h
+++ b/include/CppUTest/CommandLineArguments.h
@@ -79,8 +79,10 @@ private:
     void SetRepeatCount(int ac, const char** av, int& index);
     void AddGroupFilter(int ac, const char** av, int& index);
     void AddStrictGroupFilter(int ac, const char** av, int& index);
+    void AddExcludeGroupFilter(int ac, const char** av, int& index);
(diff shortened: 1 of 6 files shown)


basvodde:
Hey, thanks. :) Appreciated.

apbarrero:
@basvodde thank you! Cheers.

PR #791, opened 2015-11-15
add 'subdir-objects' to AM_INIT_AUTOMAKE, quiet warnings about having…

charlesnicholson:
… source / object files in subdirectories.


charlesnicholson:
tanked on travis. will reopen after root-causing and figuring out next steps.

basvodde:
AFAIK, you won't be able to fix it. It was a bug in automake, which is why we eventually ended up with keeping the annoying warning.

Issue #794, opened 2015-11-17
AppVeyor is broken, all builds of master are failing


charlesnicholson:
https://github.com/cpputest/cpputest/pull/793, a non-change to master is triggering spurious AppVeyor failures.

@Andne observed that it looks like a pathing issue, and that appveyor_ci_build.ps1 needs updating. I'm doing that and testing now, will open a PR that fixes it soon.

charlesnicholson:
PR here: https://github.com/cpputest/cpputest/pull/795

PR #795, opened 2015-11-17
Remove 'C:\Program Files\Git\usr\bin' from the path, fix AppVeyor.

charlesnicholson:
Fixes https://github.com/cpputest/cpputest/issues/794


Andne:
That's ok, this script is probably only run on AppVeyor anyways, the extra paths were mostly just to futureproof things or deal with some of the other possible environments on windows.  Really the only time the "Program Files (x86)" paths matter is if 32-bit git is installed on a 64-bit machine.  Currently that doesn't seem to be the case on AppVeyor.

Issue #797, opened 2015-11-24
Multiple mock failures per test case


arstrube:
Is the following test output normal:
```
Errors (27 failures, 15 tests, 15 ran, 73 checks, 0 ignored, 0 filtered out, 15 ms)
```
I would have thought any test case can at most fail once? This behavior is caused by several mock failures per test case. It may have been fixed; I can't remember.

I'm currently using 3.7.1.

basvodde:
Not sure. It might be that the checkExpectations of the Plugin allows for the second failure?

arstrube:
You are right. Not installing the plugin makes multiple failures per test case go away. I still need to reduce this to a simple example. It seems not every mock failure causes this effect.

arstrube:
```c++
#include "CppUTest/CommandLineTestRunner.h"
#include "CppUTest/TestHarness.h"
#include "CppUTestExt/MockSupport.h"
#include "CppUTest/TestRegistry.h"
#include "CppUTestExt/MockSupportPlugin.h"

TEST_GROUP(MultiMockFailTest)
{
    void teardown()
    {
//      mock().checkExpectations(); /* Makes multiple failures go away */
    }
};

TEST(MultiMockFailTest, MyTest)
{
    mock().expectOneCall("NonExistantMock");
	LONGS_EQUAL(1, 2);
}

int main(int ac, char** av)
{
    MockSupportPlugin mockPlugin;
    TestRegistry::getCurrentRegistry()->installPlugin(&mockPlugin);
    return RUN_ALL_TESTS(ac, av);
}
```

arstrube:
Should have been closed automatigically ?!

basvodde:
Thanks for cleaning up the issues!

PR #799, opened 2015-11-25
Fix multiple failures per test case.

arstrube:
Fixes issue #797


arstrube:
@basvodde -- merge this, perhaps?

basvodde:
Done via the other one :)

PR #801, opened 2015-11-26
CMake generates malformed cpputest.pc 

bukulin:
``libdir`` and ``includedir`` variables in cpputest.pc don't take ``exec_prefix`` and ``prefix`` into account.


basvodde:
Thanks

PR #805, opened 2015-12-15
Fix overwriting xml files if groups are mixed in their order #168

tommf:
This is a simple fix for ticket #168. 

The JunitTestOutput instance keeps book about files already written to and will append instead of cut if a file is about to reopened again.

### DIFF
--- a/include/CppUTest/JUnitTestOutput.h
+++ b/include/CppUTest/JUnitTestOutput.h
@@ -28,7 +28,8 @@
 #ifndef D_JUnitTestOutput_h
 #define D_JUnitTestOutput_h
 
-#include <map>
+#include <string>
+#include <list>
 #include "TestOutput.h"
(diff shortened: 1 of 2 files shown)


arstrube:
Hello there,

Could you perhaps do the same thing without using <string>, <map> etc.? CppUTest doesn't use these, in order to minimize implementation dependencies. This is why the class SimpleString exists. In the case of <map> I know it may feel like re-inventing the wheel....

There are several places in CppUTest where "handmade" list of various kinds are used; perhaps you could model your code after any of those. Perhaps we could even invent something like "SimpleMap"...

tommf:
I will do, but it will take some time. Probably not before the beginning of the next year.

tommf:
Don't bother with this fix. It does not work as supposed to. I will need another start for this. Sorry.

PR #806, opened 2015-12-28
Clang 3.7 fixes

offa:
Some fixes for clang (3.7).

**Note:** Tests must be *disabled* to get cpputest compiled with Clang, since they fail to build (#807).


basvodde:
I think I've made all the same changes (plus more) for clang3.7.

Could you check the current trunk and let me know if it works?

offa:
Great!

Build & Test work on latest GCC and Clang, each with and without C++11 enabled (tested on Linux).

PR #814, opened 2016-01-05
Added missing test files to visual studio projects

Andne:
This is some cleanup to the visual studio project files to add some missing test files.  I noticed this on appveyor when adding autotools, VS was running a couple hundred fewer tests than autotools or MinGW.

I haven't been able to update/test the VC6 project, I can run builds but don't have the IDE available to regenerate certain files.  Looks like it broke as part of PR #603, so I think it's been broken for a while.  `CppUTest.dep` is looking for `include\Platforms\VisualCpp\stdint.h` which no longer exists and it doesn't know how to create it.

### DIFF
--- a/tests/AllTests.vcproj
+++ b/tests/AllTests.vcproj
@@ -401,6 +401,29 @@
 				</FileConfiguration>
 			</File>
 			<File
+				RelativePath="CppUTestExt\ExpectedFunctionsListTest.cpp"
+				>
+				<FileConfiguration
+					Name="Release|Win32"
+					>
+					<Tool
+						Name="VCCLCompilerTool"
+						AdditionalIncludeDirectories=""
+						PreprocessorDefinitions=""
+					/>
+				</FileConfiguration>
+				<FileConfiguration
+					Name="Debug|Win32"
+					>
+					<Tool
+						Name="VCCLCompilerTool"
+						AdditionalIncludeDirectories=""
+						PreprocessorDefinitions=""
+						ForcedIncludeFiles=""
+					/>
+				</FileConfiguration>
+			</File>
+			<File
 				RelativePath="CppUTestExt\GMockTest.cpp"
 				>
 				<FileConfiguration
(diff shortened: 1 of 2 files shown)


arstrube:
@Andne - I should be able to do the 2005 / VC6 thing - haven't used it in a while though..... stdint.h needs to be removed from the project, because it was never used (recently anyway).

arstrube:
I found that the split up test cases are already in the VC6 project, however, a number of other tests were missing. Currently I am facing two challenges:
1. The total number of test cases is 764 (VS2010: 799). As far as I can tell, I have added all test files.
2. I get one test failure, which I cannot explain (not present in VS2010):
```
...\MockSupportTest.cpp(186): error: Failure in TEST(MockSupportTestWithFixture, CHECK_NO_MOCK_FAILURE_LOCATION_failed)
..\include\CppUTest/TestTestingFixture.h(112): error:
 actual <.
OK (1 tests, 1 ran, 0 checks, 0 ignored, 0 filtered out, 0 ms)
>
 did not contain  <Unexpected mock failure:>

**Comment by arstrube** [comment]
Okay, I found the culprit for 2. But I can't think of a decent solution. In `MockFailureTest.h` somehow VC6 doesn't inline `mockFailureString()` correctly, which subsequently causes the failure string not to be seen.

One solution is to replace the inline function with
```c++
#define mockFailureString() /* inline function will not work in VC6 */ \
    MockFailureReporterForTest::getReporter()->mockFailureString
```
This is rather ugly, and I am hesitant to propose it, just to take care of one bug in a really old compiler.

Previously, the implementation of `mockFailureString()` was in one of the test sources. This was ugly to begin with, and would break the 16-bit build. 

That leaves introducing `MockFailureReporterForTest.cpp`, which would amount to a lot of work and kind of an overkill for one small function.

Is there perhaps another option that I am not seeing?

basvodde:
Proposed a better way in #822

PR #816, opened 2016-01-06
Fixed a failed test on clang++3.7.

basvodde:
The separate process test returned a different signal. Looking into it, it seems that signal is also valid and thus it shouldn't fail the test. For now, created a test that has an 'or' in it which checks both signals. Not the best choice, but I guess it will do for now?

@arstrube: Might want to check this one. You think this fix is ok?

### DIFF
--- a/include/CppUTest/TestTestingFixture.h
+++ b/include/CppUTest/TestTestingFixture.h
@@ -100,16 +100,19 @@ public:
         return genTest_->hasFailed();
     }
 
-
     void assertPrintContains(const SimpleString& contains)
     {
-        assertPrintContains(output_, contains);
+        assertPrintContains(getOutput(), contains);
     }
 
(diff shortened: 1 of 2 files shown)


arstrube:
I think the fix is fine. We are testing CppUTest and not the compiler, and we want to make sure it crashes properly, which it does; the signal is secondary here. Slightly odd all the same, how the signal was changed all of the sudden...

arstrube:
On second thoughts (listening to my own comment ;-), since we are really not concerned which signal is raised, how about just simply testing for `"signal"` or `"signal "`? in the output?

basvodde:
I thought you had put the signal number on purpose to check that it isn't killed for the wrong reason :)

arstrube:
You may be right there... it's been a while. Either way is fine with me.

basvodde:
Decided to keep it for now, as long as there are just 2 signals :)

PR #818, opened 2016-01-11
Replace includes with excludes in travis.yml

### LINKED ISSUE #809: Travis CI stuck?
I noticed that my last pull request was still pending for Travis CI so I went to see what it was doing.  It looks like a build from a couple weeks ago never finished all of the configurations and is holding up any other builds from completing.  Some of the configurations have completed, but the next configuration isn't starting for some reason.

From what I can tell, builds 2099 and 2100 are both waiting to start a gcc/cmake build with coverage, but for some reason that build isn't starting.  2099 seems to be trying to build a commit that doesn't exist, so maybe 2100 is just waiting for that particular configuration to complete before it starts it's build with that configuration?  Maybe someone just needs to cancel the hung build and everything will catch back up to itself again.

Andne:
This appears to fix #809 for now.  I don't think it's ideal since the excludes are maybe harder to quickly process, but trying to make it use includes only doesn't appear to work right now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,16 +19,25 @@ env:
 matrix:
     exclude:
     - env: BUILD=cmake-coverage
+      compiler: clang
+    - env: BUILD=cmake-coverage
+      compiler: wcl
     - env: BUILD=autotools_old_compilers
     - env: BUILD=make-dos
-    - env: BUILD=test_report
-    - compiler: wcl
-    include:
-    - env: BUILD=cmake-coverage
-      compiler: gcc
-    - env: BUILD=test_report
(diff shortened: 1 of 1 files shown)


Andne:
Looks like you're going to need to cancel a few builds in order for the build for this pull request to run successfully.

Andne:
Hold off on this one, I think I had a different issue messing with the includes (travis build queue was stuck), so I'm going to try that again.

basvodde:
Ok, I'll continue cancelling builds though :)

basvodde:
This one passed now.

basvodde:
Closed

PR #819, opened 2016-01-12
Travis.yml updates to use includes only

Andne:
Here's my solution with just includes, I think it's easier to read than the one with excludes.  Your choice which one you like better though.


basvodde:
Better

arstrube:
Better. Also better than my previous attempt :)

PR #820, opened 2016-01-13
Update READMEs to make InstallScript.sh working

nanakira:
Hi,
I followed README files to install scripts, but it seems it doesn't work correctly.
I think these READMEs should be updated.
Could you please review them?


basvodde:
Looks ok to me. Let me see if @jwgrenning has a comment to it as he uses the scripts more than me.

basvodde:
As @jwgrenning didn't comment, I'll merge it :)

PR #821, opened 2016-01-13
Fix broken VC6 test; add missing files to projects

arstrube:
I don't really like this fix, but I dislike the alternatives even more.

### DIFF
--- a/CppUTest.dsp
+++ b/CppUTest.dsp
@@ -166,6 +166,10 @@ SOURCE=.\SRC\CPPUTEST\SimpleString.cpp
 # End Source File
 # Begin Source File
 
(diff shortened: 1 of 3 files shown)


basvodde:
Hey, sorry. I didn't notice the #define is outside the platform specific files. That is indeed not ok :)

basvodde:
Why not simply put the implementation in a cpp file?

arstrube:
I tried to avoid that, because we'd need a new cpp file (because the DOS build ends up needing it in several executables), and adding a new .cpp file means touching all platforms.....

Perhaps there is another way? A pragma for VC6 that prohibits inlining, for example. Or perhaps even a command line option?

PR #822, opened 2016-01-14
Moved functions in the cpp file

basvodde:
Removed the #define for VC6

### DIFF
--- a/tests/CppUTestExt/MockFailureTest.cpp
+++ b/tests/CppUTestExt/MockFailureTest.cpp
@@ -31,6 +31,17 @@
 #include "CppUTestExt/MockExpectedCallsList.h"
 #include "MockFailureTest.h"
 
+UtestShell* mockFailureTest()
+{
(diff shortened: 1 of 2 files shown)


arstrube:
Yeah... as I said, that will break the Dos build :-/

basvodde:
Why?

arstrube:
Let me hunt for a compiler option. I would like that best, because only one compiler is concerned anyway....

basvodde:
Uhm, I'd prefer to move the inline functions to a cpp file.

basvodde:
Usually inline is only because of my laziness :)

arstrube:
But there's also that there are some platforms that are easy to forget and hard / impossible for us to test. 

Perhaps it would be a good idea to create an issue for everything that involves adding / removing / renaming files, and then do them all at once (like with the splitting of mock tests).

My bad not to have proposed a separate source to begin with.

basvodde:
Yeah, perhaps...

PR #823, opened 2016-01-14
Put the Vc6 fix where it belongs

arstrube:
(actually, it is the method `MockFailureReporterForTest::getReporter()` that must not be inlined)

### DIFF
--- a/tests/AllTests.dsp
+++ b/tests/AllTests.dsp
@@ -42,7 +42,7 @@ RSC=rc.exe
 # PROP Ignore_Export_Lib 0
 # PROP Target_Dir ""
 # ADD BASE CPP /nologo /W3 /GX /O2 /D "WIN32" /D "NDEBUG" /D "_CONSOLE" /D "_MBCS" /YX /FD /c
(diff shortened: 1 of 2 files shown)


basvodde:
Sure, but we'll need to pull out the inline functions sometime anyway. But ok, we'll delay that.

arstrube:
Pulling them out would definitely be even better. But it will mean:
1. Add `MockFailureReporterForTest.cpp`
2. Preferably rename the header to match.
3. Update all platforms / build environments to use the new sources.

Right now, it's okay for them to be inlined with all other compilers we use, so I don't know how important it is to pull them out.

basvodde:
Ok, I'll take it then as it is after it passed the build :)

arstrube:
Okay. I propose to find all the places where TeamCityTestOutput.cpp / .h are still missing, and pull out MockFailureReporterForTest at the same time. The I will test all I can (VC6, Cl2000) and I will have to change any other build configurations blindly.

basvodde:
Made an issue for it ? :)

arstrube:
No, but can do :)

arstrube:
will do

Issue #824, opened 2016-01-14
Add new sources where still missing, and pull out  MockFailureReporterForTest.cpp


arstrube:
1. I am about to check some recently added sources that might still be missing in some platforms / build configurations, and prepare a PR to add them. One likely candidate is `TeamCityTestOutput.cpp/.h`. Please mention any others you might be aware of in this issue.
Potentially the new Mock test sources might not be reflected correctly for some of the lesser used platforms.

2. `MockFailureTest.h` to be split in `MockFailureReporterForTest.h` and `MockFailureReporterForTest.cpp`, and inlined functions to be replaced by normal ones.

arstrube:
The Cl200 has compiler errors and test failures which I propose not to fix at this time.
I suggest we defer the changes to the remaining platforms until someone actually needs them.

PR #825, opened 2016-01-14
Fixes for #824.

arstrube:
Included so far:
- Pull out `MockFailureReporterForTest.cpp`
- Correct all build targets for Travis Ci (cmake, autotools, dos)
- Add missing tests to Dos target
- Incorporate changes for VC6
- Incorporate changes for VS2005 & later

To do:
- Incorporate changes for Cl2000
- Incorporate "blind" changes to other targets

This fixes issue #824.


arstrube:
The Cl2000 compiler is giving me grief, because it will insist on not assigning a "C" function pointer to a "C++" function pointer, therefore it complains about the new function pointer parameter in `MockSupport_c.cpp`

arstrube:
@basvodde I think this is okay to merge now. The only things missing now are the targets that require 'blind' changes, and Cl2000, which requires additional fixes.

Both of these, especially the Cl2000, can be handled in separate PRs.

PR #826, opened 2016-01-14
More fixes for issue #824 related to Cl2000 compiler

arstrube:
Updated the project files with current source files.

The following problems were also found:
- Distinction between C and C++ function pointers causes compiler errors
- sizeof(char)==sizeof(short)==1 cost bits equal test to fail
- Some sources had to be moved around to stay within memory constraints

Solutions to the above problems have been provided, but may not be acceptable yet. There is probably no point in updating the project files without also addressing the compiler errors and test failures.

Fixes issue #824.


include/CppUTest/UtestMacros.h
```diff
@@ -249,8 +249,10 @@
 #define BITS_EQUAL_TEXT(expected, actual, mask, text)\
   BITS_LOCATION(expected, actual, mask, text, __FILE__, __LINE__)
 
+#define ROBUST_SIZEOF(type) (2*sizeof(type)/sizeof(short)) /* For the Cl2000, sizeof(char) == sizeof(short) == 1 */
```
basvodde: Hmmm. You know I don't like compiler specific code in the headers :) If we start that, it won't end eventually. Is there no way we can put this somewhere else or solve it another way?

src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -311,7 +313,7 @@ MockExpectedCall_c* withConstPointerParameters_c(const char* name, const void* v
 
 MockExpectedCall_c* withFunctionPointerParameters_c(const char* name, void (*value)())
 {
-    expectedCall = &expectedCall->withParameter(name, value);
+    expectedCall = &expectedCall->withParameter(name, (Cpp_Func*)value);
```
basvodde: Can you write the cast in here, without the typedef? And remove the compiler specific comment

include/CppUTest/UtestMacros.h
arstrube: I agree. One possibility would be PlatformSpecificSizeOf(). I don't like that particularly, either, just for one platform that is quickly getting outdated....

It would be better, imo, to do a "hack" just for the Cl2000, something like redefining sizeof() in a forced include, perhaps? Need to dig a little deeper to determine whether that is possible.

src/CppUTestExt/MockSupport_c.cpp
arstrube: Unfortunately, no. Of course I can remove the comment. But the only way to cast to "C" or "C++" that I know of is via typedef. As it is not normally a problem, C / C++ doesn't seem to provide another way to cast void (*)() C to void(*)() C++.

arstrube: Again, it feels wrong to change the code here because of a quirk in just one compiler. Sometimes I feel like just dropping support for the Cl2000. Besides, these problems have been around for half a year and nobody has noticed.

I do wonder though about the whole idea of a function pointer parameter. Sure, I don't have to cast a void, void function anymore. But any other type of function with parameters, return value etc. I'd still have to force-cast to a basically incompatible signature, right?

So I was wondering, what is the advantage that this operation has over just casting to a generic pointer in the first place?

basvodde: Does (void ()()) not work as cast?

arstrube: It would still be a cast to cpp - func. But a c - func is passed via the argument.

arstrube:
I find that there are test failures and memory errors in the CppUTestExt part of the cl2000 run, and I cannot take the time to look into them and fix them. I would therefore suggest the following:
- I only fix the project files
- If we can agree on how to do it, I will include fixes for the compiler errors
- I will not fix any failures or memory access / leak problems on the cl2000

arstrube:
@basvodde the Dos build failed due to some timeout - could you restart the jobs, perhaps?

arstrube:
It should then pass.

basvodde:
Ok. Should be ok.

Issue #817, opened 2016-01-11
is adding expect no call check reasonable?


cuitoldfish:
Hi,
I have been used CppUTest in my daily CI for a long time, and the requirements pop out from time to time that I need to check one mock function is not invoked in some case. I would like devote some time to add this functionality, do you guys think it's valuable?

arstrube:
Hi there,

Actually, this check is standard. Perhaps a call to `mock().checkExpectations()` might be missing, if an unexpected call does not cause a failure.

Simply _not_ specifying an expectation is sufficient.

In my own work, I sometimes use:
```
mock().expectNcalls(0, "Somefunction");
```
rather than a comment, to make explicit in my test case that this call is not supposed to occur. This is purely for the benefit of the reader; it is not necessary for an unexpected call to "Somefunction" to fail the test.

arstrube:
Hi there,

Actually, this check is standard. Perhaps a call to `mock().checkExpectations()` might be missing, if an unexpected call does not cause a failure.

Simply _not_ specifying an expectation is sufficient.

In my own work, I sometimes use:
```
mock().expectNcalls(0, "Somefunction");
```
rather than a comment, to make explicit in my test case that this call is not supposed to occur. This is purely for the benefit of the reader; it is not necessary for an unexpected call to "Somefunction" to fail the test.

cuitoldfish:
Hi arstrube,
I think sometimes it's necessary to fail the test if an unexpected call happens.

say, I have a decorator class, and also there is a delegation class inside that decorator.

```.cpp
decorator::functionA(Interface *pInterface)
{
    if (NULL == pInterface)
            return;
    //some decoration here
    .....
    //delegate call to delegation class
    delegator->functionA(pInterface);

    //may be some decoration also here
    .....
}
```
in my decorator class's unit test, I should have one NULL test to ensure that, if one exceptional NULL value is passed as the parameter, then the decorator will not call the delegator's functionA. 

In this case, in the decorator mock class, if I use one *expect no call*,  when someone modify my codes and remove the guard clause for NULL by mistake, the NULL test will fail, this will help me to protect my decorator class always has correct behavior.

if i use expectNcalls with number 0, , NULL test will still pass but functionality of this decorator::functionA is break.

arstrube:
Sure, but my point is that your NULL test _will_ fail, in that case. It should be simple for you to try out, no? E.g. remove the guard clause and see what happens.

arstrube:
[...]

PR #827, opened 2016-01-17
Fix for #817 one balanced implementation for expecting no call functionality

cuitoldfish:
after several implementations try, I chose this as my solution to for the expecting no call functionality. It's the more balanced one, for satisfying the simplicity , sufficient and compliant with existed system design requirement .
Let's see if it's worth to be added.

#Solution Description
add one unexpected list in MockSupport, and add one expectNoCall in MockSupport. When expectNoCall() is invoked, One unexpected call will be appended to the unexpected list.
then in actualCall(), current existed expectation empty judge will changed to judge if both expectation list and unexpectation list are empty. If expectation list is empty and unexpectation list is not empty, actualCall() will go ahead to trigger the unexpected call failure.

### DIFF
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -47,6 +47,7 @@ public:
 
     virtual void strictOrder();
     virtual MockExpectedCall& expectOneCall(const SimpleString& functionName);
+    virtual void expectNoCall(const SimpleString& functionName);
     virtual MockExpectedCall& expectNCalls(int amount, const SimpleString& functionName);
     virtual MockActualCall& actualCall(const SimpleString& functionName);
     virtual bool hasReturnValue();
@@ -127,6 +128,7 @@ private:
     MockFailureReporter *standardReporter_;
     MockFailureReporter defaultReporter_;
     MockExpectedCallsList expectations_;
+    MockExpectedCallsList unExpectations_;
     bool ignoreOtherCalls_;
     bool enabled_;
     MockCheckedActualCall *lastActualFunctionCall_;
@@ -144,6 +146,9 @@ private:
     MockNamedValue* retrieveDataFromStore(const SimpleString& name);
 
     MockSupport* getMockSupport(MockNamedValueListNode* node);
+    
+    bool hasntExpectationWithName(const SimpleString& functionName);
+    bool hasntUnexpectationWithName(const SimpleString& functionName);
 };
 
 #endif
--- a/src/CppUTestExt/MockSupport.cpp
+++ b/src/CppUTestExt/MockSupport.cpp
@@ -118,7 +118,8 @@ void MockSupport::clear()
     tracing_ = false;
     MockActualCallTrace::instance().clear();
 
-    expectations_.deleteAllExpectationsAndClearList();
+    expectations_.deleteAllExpectationsAndClearList();    
+    unExpectations_.deleteAllExpectationsAndClearList();
     compositeCalls_.clear();
     ignoreOtherCalls_ = false;
     enabled_ = true;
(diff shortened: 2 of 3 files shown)


basvodde:
Uhm, if you add it to the exceptationsList with 0 amount, then it should also work and it should be easier, right?

cuitoldfish:
do you mean just directly use `expectNCalls(0, "")`? as we has discussed, this can not work if ignoreOtherCall is specified

cuitoldfish:
I need to get no call check with no break current ignore other call functionality.

this is showed in below test case

```.cpp
    mock().expectNoCall("lazy");
    mock().ignoreOtherCalls();
    mock().actualCall("bar").withParameter("foo", 1);
    mock().actualCall("bar1").withParameter("foo", 1);
    mock().actualCall("bar2").withParameter("foo", 1);
    mock().actualCall("lazy");

    //call to lazy is checked, while ignore other call except "lazy"
```

let the expectation list with 0 amount can not satisfy.

arstrube:
Much as I would like to see this functionality, somehow the "un-expectation-list" doesn't sit quite right with me, even though it obviously does the job......

cuitoldfish:
I would like to see this functionality too, so @arstrube would you like to have a try too?

basvodde:
@cuitoldfish I think your attempt was a good start. I'd suggest to use the expectionList rather than create an un-expectationList. If you add an expectation with 0 to the exceptationList, that ought to work. Should be easier to implement.

cuitoldfish:
what's the meaning of  " with 0 "?

basvodde:
Sorry, you are right that was unclear. We still didn't implement an amount in the expected call yet.

What I mean is to put an expectation in the fulfilled list. But thinking it over, it might become ugly, but I'm not sure. It would be more easy. @arstrube any idea on this?

cuitoldfish:
@basvodde I has tried this possible solution in my local. 

I create one expectation and initiate it as fulfilled, then add it to the expectation list. Then expect no call will work, but....

even ignore the ugly, this solution still conflict with current print meaning.

1. if we add that expectation in the expectation list, then
`MockUnexpectedCallHappenedFailure` will report **"Mock Failure: Unexpected additional (%dth) call to function: "**, what print i suggest is still **"Mock Failure: Unexpected call to function: "**

2. in the case that expect no call is specified, but one 'MockExpectedCallsDidntHappenFailure' is generated, then the prints **"\tEXPECTED calls that did NOT happen:\n" "\n\tACTUAL calls that did happen (in call order):\n"** in 'MockExpectedCallsDidntHappenFailure' will include that expectation initiated as fulfilled

in a word, adding one expectation of no call in the expectation list might change current meaning of fulfilled list and unfulfilled list. and changing them will challenge the current function name design.

that's why i am forced to change to use the unexpected list ;(

basvodde:
Ok, does make sense. Want to hear from @arstrube if he has any idea or opinion on this implementation. If not, I'll merge this.

arstrube:
As I don't have a better idea at this time, let's go with this one.

arstrube:
I guess I would probably try to find a better name for "unExpectation", but it's okay - a bit like the Undead....

basvodde:
Do you @arstrube have any alternative suggestion?

arstrube:
I was thinking of `Exclusions` but not too happy with that, either.

arstrube:
`MustNotbeCalledList` or perhaps `ExpectNoCallsList` etc. But `UnExpectationList` is kind of funkier...

basvodde:
Ok, lets keep it then :) Thanks!

arstrube:
screenshot

@cuitoldfish perhaps you might consider in the future dividing up your commit messages into a short heading, followed by an empty line (important!), followed by the rest of your message. In this way, it wouldn't look so ungainly in a command line listing, and still contain all the information you wanted to put in there:

```
Adding to CPPUTEST_MEM_LEAK_DETECTION_DISABLED to VS6 project 

to get around unresolved externals 1>CppUTest.lib(MemoryLeakWarningPlugin.obj) :
error LNK2001: unresolved external symbol... etc. etc.
```
Thanks!

cuitoldfish:
sorry for that, I will follow your guide in the future! Thanks for the reminder

Issue #829, opened 2016-01-18
strictOrder() not honored when using Mock Name (aka namespace, scope)


matthewpiatt:
Hey Bas / James / others:

In the midst of authoring some tests, I came across an unexpected situation. It seems as though strictOrder() is not honored when a mock name is supplied. (And yes, I am installing the MockPlugin in my main).

Is this expected behavior?

If this is a bug, I'd even be willing to help fix this, if that is of interest. Is there something you can point me to as a place to start?

I'll let my tests do the talking from here:

```
TEST_GROUP(FooTests)
{
};

TEST(FooTests, StrictOrderViolatedWhenNotUsingMockName)
{
	mock().strictOrder();

	mock().expectOneCall("A");
	mock().expectOneCall("B");

	mock().actualCall("B");
	mock().actualCall("A");

	// THIS FAILS AS EXPECTED
}

TEST(FooTests, StrictOrderViolatedWhenUsingMockName)
{
	mock("foo").strictOrder();

	mock("foo").expectOneCall("A");
	mock("foo").expectOneCall("B");

	mock("foo").actualCall("B");
	mock("foo").actualCall("A");

	// THIS DOES NOT FAIL AS EXPECTED
}
```

matthewpiatt:
Update:

Just to clarify, I don't leave failing tests laying around or anything -- I just always start my TDD cycle with a failure :)

So I actually just tried the following (see below), and things now work. Perhaps the issue is with the mock plugin not checking certain expectations for things that use mock names?

I poked around in tests/CppUTestExt/MockSupportTest.cpp and actually saw that CppUTest covers itself for similar behavior. Not sure what I am missing. The [MockPlugin section of the docs] (https://cpputest.github.io/mocking_manual.html#mock_scope) state that using the MockPlugin calls "checkExpectations at the end of every test (on global scope, which goes recursive over all scopes)".

```
TEST_GROUP(FooTests)
{
};

TEST(FooTests, StrictOrderViolatedWhenNotUsingMockName)
{
	mock().strictOrder();

	mock().expectOneCall("A");
	mock().expectOneCall("B");

	mock().actualCall("B");
	mock().actualCall("A");
}

TEST(FooTests, StrictOrderViolatedWhenUsingMockName)
{
	mock("foo").strictOrder();

	mock("foo").expectOneCall("A");
	mock("foo").expectOneCall("B");

	mock("foo").actualCall("B");
	mock("foo").actualCall("A");

[...]

PR #839, opened 2016-01-22
another proposed fix of #829

cuitoldfish:
For current fix for #829 , I'm afraid if lastActualFunctionCall is null then we lost chance to check other scope out of order.
for example:
```.cpp
    mock().strictOrder();
    mock("bla").strictOrder();
    mock().ignoreOtherCalls();

    MockExpectedCallsListForTest expectations;
    expectations.addFunction("foo1", 1)->callWasMade(2);
    expectations.addFunction("foo2", 2)->callWasMade(1);
    MockCallOrderFailure expectedFailure(mockFailureTest(), expectations);

    mock("bla").expectOneCall("foo1");
    mock("foo").expectOneCall("foo1");
    mock("foo").expectOneCall("foo2"); 

    mock("bla").actualCall("foo1");
    mock("foo").actualCall("foo2");
    mock("foo").actualCall("unexpected1");
    mock("foo").actualCall("foo1");
    mock("foo").actualCall("unexpected2");
```
I admit i'm fastidious about the out of order check :)

I have another proposed fix pull request, hopefully you guys can give some look and comments.

### DIFF
--- a/include/CppUTestExt/MockExpectedCallsList.h
+++ b/include/CppUTestExt/MockExpectedCallsList.h
@@ -53,6 +53,8 @@ public:
     virtual void addExpectedCall(MockCheckedExpectedCall* call);
     virtual void addExpectations(const MockExpectedCallsList& list);
     virtual void addExpectationsRelatedTo(const SimpleString& name, const MockExpectedCallsList& list);
+    
+    virtual void onlyKeepOutOfOrderExpectations();
     virtual void addUnfulfilledExpectations(const MockExpectedCallsList& list);
 
     virtual void onlyKeepExpectationsRelatedTo(const SimpleString& name);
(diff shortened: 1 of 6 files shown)


arstrube:
Looks good to me :-)

cuitoldfish:
One thing I am not sure is that, if we get one out of order call failure, say, for these prints
```.cpp
	Mock Failure: Out of order calls.
	
```
then we only print those out of order call in `ACTUAL calls that did happen (in call order):`, or we should print all call that did happen no matter out of order or not? as far as now I think CppUTest only print out of order calls

arstrube:
I think the latter; it makes it easier to see which calls are affected. With all calls listed, that could get a bit more confusing. But I don't know for sure @basvodde what do you think?

basvodde:
Seems ok. I'll merge this now.

Issue #840, opened 2016-01-25
Export missing parts of C++ interface to C


ykaliuta:
I've checked MockSupport_c.h and it looks like some part of feature set is not available for C. In particular I'm missing ignoreOtherCalls() and ignoreOtherParameters(), but expectNoCall or expectNCalls would be nice to have too. Would it be possible to make all the features available?

Thanks!

arstrube:
Hi there, I just discovered the same thing when checking test coverage. However, I don't know yet whether I'll be able to add ALL of the interfaces you mention... I am currently working on `withOutputParameterOfType()`. I am not sure how come the `ignore...` interfaces aren't there or whether there was a problem with `expectNCalls()`. I'll look into that while I am busy with it anyway.

arstrube:
I am busy with PR #846. It's taking longer as expected because I'm also adding some niceties such as being able to write
```c++
mock_c()->returnIntValueOrDefault(5)
mock_c()->intReturnValue();
```

arstrube:
@ykaliuta  All should be taken care of now - close this?

@basvodde Shouldn't this Issue have auto-closed when the associated PR was merged?

ykaliuta:
Thanks a lot! Sorry, I'm travelling and have limited internet access.

PR #846, opened 2016-01-27
 More interfaces for MockSupport_c...

arstrube:
This fixes issue #840.


arstrube:
I guess this PR is growing way too large.....

cuitoldfish:
seems Travis CI is broken?

arstrube:
@cuitoldfish - Yup. Travis isn't well at all. Sort of on-and-off erroring all the time....

arstrube:
That's about it. There's a small bit of 'residue', which either doesn't make sense in C or I'm still unsure about (stuff within /** */ in MockSupport_c.h). Maybe @basvodde could have a look at those.

basvodde:
Can you remove the comments ?

arstrube:
Sure, but only after discussming them with you :-). I'd like your opinion on which ones might end up as interface, and which ones as simply being removed.

There are also some whitespace changes in two unrelated files that krept in. I can split those off for another PR if you prefer.

basvodde:
I suggest to not implement them if we don't need them :)

Would be better if whitespace changes are in other PRs in the future, but lets not add more work now?

arstrube:
Line coverage for this PR should be 100%.

PR #848, opened 2016-01-28
change getOneFulfilledExpectationWithIgnoredParameters to right name

cuitoldfish:
change getOneFulfilledExpectationWithIgnoredParameters to right name which compliant with it's behavior

### DIFF
--- a/include/CppUTestExt/MockExpectedCallsList.h
+++ b/include/CppUTestExt/MockExpectedCallsList.h
@@ -67,7 +67,7 @@ public:
 
     virtual MockCheckedExpectedCall* removeOneFulfilledExpectation();
     virtual MockCheckedExpectedCall* removeOneFulfilledExpectationWithIgnoredParameters();
-    virtual MockCheckedExpectedCall* getOneFulfilledExpectationWithIgnoredParameters();
+    virtual MockCheckedExpectedCall* getOneFulfilledExpectationWithoutIgnoredParameters();
 
     virtual void resetExpectations();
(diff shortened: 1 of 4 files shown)


cuitoldfish:
@basvodde what's your opinion? This name even perplex me when I read those code parts, because it do not describe correctly what it does.

cuitoldfish:
OK maybe I do not describe clearly.
the detail is, there is a discrepancy between this function `MockExpectedCallsList::getOneFulfilledExpectationWithIgnoredParameters`'s behavior and it's name.
```.cpp
MockCheckedExpectedCall* MockExpectedCallsList::getOneFulfilledExpectationWithIgnoredParameters()
{
    for (MockExpectedCallsListNode* p = head_; p; p = p->next_) {
        if (p->expectedCall_->isFulfilledWithoutIgnoredParameters()) {
            return p->expectedCall_;
        }
    }
    return NULL;
}
```

this function should own it's fit name. @basvodde do you think it should be fixed or just omitted?

arstrube:
Personally, I think it should be changed. I hate stuff with names that suggest it does something else than what it actually does. Same is misleading comments ;-)

cuitoldfish:
so?

basvodde:
I'll check it later, I didn't check it in detail yet, sorry!

cuitoldfish:
@basvodde did you get time to check this PR? :)

basvodde:
Hey, can you open it again? Why close it?

PR #851, opened 2016-01-29
Prune install section in .travis.yml; install Dosbox only for make_dos.

arstrube:
This fixes issue #809 - well a related issue which causes all jobs to error when for some reason Dosbox (which is only needed for `make_dos`) cannot be installed.

I tested the github deployer part on my fork and it work (installing correctly, then displaying a message that nothing will be deployed, because it is running on a for - all as expected). After that, I commented out the github deployer part again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,11 @@ matrix:
     - compiler: gcc
       env: BUILD=cmake_gtest
     - compiler: gcc
-      env: BUILD=cmake-coverage
+      env: BUILD=cmake_coverage
     - compiler: gcc
       env: BUILD=test_report
     - compiler: wcl
-      env: BUILD=make-dos
+      env: BUILD=make_dos
 global:
 - os: linux
 - rvm: '1.9.3'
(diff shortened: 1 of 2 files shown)


basvodde:
This causes problems due to the "new hardware" Travis... right?

arstrube:
No, it's not related to the new hardware stuff actually. But it did happen that DosBox could not be installed, and then all jobs would error when only one needed it. So I figured it would be wise to move the "specialty" installs into the build script.

arstrube:
Same goes for Travis Github Deployer.

The update command is issued by the standard apt: block (minuse the --fixmissing, but that doesn't seem to matter anymore)

basvodde:
See: https://docs.travis-ci.com/user/workers/container-based-infrastructure/

arstrube:
In order to not use sudo (with or without my proposed streamlining), we'd have to get them to put DosBox on the whitelist.

Let's say I reduced the sudo-count by 2, if nothing else :-). I think DosBox is the only one left.

basvodde:
Yes, need to get it on the whitelist. Especially as the non-container infra will go away at some point.

arstrube:
At least, while we are still using sudo, only the make_dos job would break, not everything.

.travis.yml
```diff
@@ -34,15 +34,7 @@ global:
 addons:
   apt:
     packages:
-      # - dosbox
```
basvodde: Maybe put this back, as eventually it ought to be here :)

arstrube: Okay, done.

There is a sudo for openwatcom, too, unfortunately, which won't go away unless an apt package were created for openwatcom, or we need to find a way to install the package for the local user. I hear this is possible but somewhat involved.

basvodde: At some point, we'll need to figure it out :) It would speed up the build a lot, I guess.

PR #852, opened 2016-01-30
IEEE754 floating point exception flags Plugin in CppUTestExt

arstrube:
Here's the other one for comparison. It is not quite where it should though.


arstrube:
The failures we are seeing here on travis (e.g. build=cmake) are identical to the ones on my Mac. So it seems to be a more general problem. Never happened under Cygwin though?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL <copyright holder> BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#if defined(__cplusplus) && ((__cplusplus >= 201103L) || (defined(_MSC_VER) && (_MSC_VER >= 1600)))
```
basvodde: Is the MSC_VER needed?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -0,0 +1,75 @@
...
+#define IEEE754_CHECK_CLEAR(flag) { \
+    if(!hasFailed_) { \
+        result_->countCheck(); \
+        if(flag) { \
+            CheckFailure failure(test_, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", #flag); \
+            result_->addFailure(failure); \
+            hasFailed_ = true; \
+        } \
+    } \
+}
```
basvodde: Can you not use a macro here and try to put as much of this code as possible in functions instead?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: I don't know for sure. I got this from somewhere. I guess the next term would just return false, if it wasn't defined?

tests/CppUTestExt/IEEE754PluginTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    void setup(void) override {
+        registry->installPlugin(&ip);
+    }
+};
+
+IGNORE_TEST(IEEE754ExceptionsPlugin2, should_not_fail_in_ignored_test) {
+    set_everything_c();
+}
+
+#endif
```
arstrube: I don't like the fact that nothing is compiled when c++11 is not set. This makes it very hard to distinguish passed tests from tests not run and the functionality being absent. Ideas? E.g. an alternative block of IGNORE_TEST tests.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
basvodde: Could you try without the MSCVER as the cplusplus macro is part of the standard and ought to work :)

tests/CppUTestExt/IEEE754PluginTest.cpp
basvodde: I wouldn't, but its up to you :)

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: I think I got your answer wrong. Yes I believe is needed, because older versions of MS C probably do c++ but don't support the functions we are using.

arstrube: We don't test many versions of MSC. The only way to find out it doesn't work is via user feedback..... which will make for some very frustrated users, potentially...

basvodde: Ok, the lets add this to CppUConfig and not have this #define here like this. It is likely to grow and repeat itself over many places.

arstrube: *ought to*. Apperently MS doesn't support c++11 completely (see comment in https://github.com/msgpack/msgpack-c/pull/283/files) so in our case it will be safer to check.

arstrube: see also here: https://connect.microsoft.com/VisualStudio/feedback/details/763051/a-value-of-predefined-macro-cplusplus-is-still-199711l

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Will have a look at it.

arstrube: Why don't you suggest an alternative in code? I can't think of a good one.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Agreed.

arstrube:
All in all, I am still not convinced this is a good idea. The failures are due to the optimization setting in CppUTest.

If we let the plugin have it's own folder and Makefile, none of this is a problem. Having to somehow marry it with CppUTest's general build settings makes something reletaively simple a huge challenge to setup (and maintain). Not to talk about the two functions not available under Mac OS....

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
Andne: Right, Visual C++ doesn't define the `__cplusplus` macro correctly for supporting newer features, so in order to know if it supports a feature or not you need to check against `_MSC_VER`.

basvodde:
Well, the CppUtestExt is for these kind of things. The mocking has become bigger than originally expected, so that might need to be extracted. I wouldn't want additional directories for things like these, especially not with the low amount that we got.

PR #854, opened 2016-02-04
Dosbox has been white-listed

arstrube:
With this PR, we'd be using the new, container-based infrastructure on Travis CI.

Is that cool, or is that cool?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+sudo: false
 matrix:
     include:
     - compiler: clang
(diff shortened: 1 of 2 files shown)


arstrube:
There is one remaining sudo in the Dos job. It is needed in order to get execute permissions on the watcom/binl directory. Is there perhaps another way to get execute permissions on the Watcom binaries?

arstrube:
Part of the problem with Watcom C is that it's Linux executables (wcc etc.) don't have the x permission set to start with. So one needs to use sudo chmod.

arstrube:
@basvodde - https://travis-ci.org/cpputest/cpputest/builds/107014698. Looks good, no?

basvodde:
That is indeed good!

PR #855, opened 2016-02-04
Message variable name changed to avoid possible naming conflicts

offa:
The test macro `CHECK_THROWS()` uses the variable name `msg` internally. Since this name is also often used within tests, this will cause a build failure. It'll be better so change it so something less usual.

## Example:

```cpp
... 
TEST(ExampleTest, nameConflict)
{
    ExampleMessage msg;
    CHECK_THROWS(std::exception, msg.willThrow());
}
```

This will result in an error `»class SimpleString« has no member named »willThrow«`.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -277,17 +277,17 @@
 #if CPPUTEST_USE_STD_CPP_LIB
 #define CHECK_THROWS(expected, expression) \
     { \
-    SimpleString msg("expected to throw "#expected "\nbut threw nothing"); \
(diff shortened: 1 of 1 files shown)


basvodde:
Did it cause problems? The { } around it should resolve it and not make it a problem...

offa:
> Did it cause problems?

Yes, because within the `{ }` the CppUTest Message `msg` is used instead the one passed as parameter to `CHECK_THROWS()`. On my example above, `willThrow()` is called on `SimpleString` instead.

It's more a naming conflict, not a serious problem.

PR #856, opened 2016-02-05
 IEEE754 Floating Point Exception Flags Plugin

arstrube:
Follow-up PR for #852

### DIFF
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -178,6 +178,7 @@
 
 /* Visual C++ 10.0+ (2010+) supports the override keyword, but doesn't define the C++ version as C++11 */
 #if defined(__cplusplus) && ((__cplusplus >= 201103L) || (defined(_MSC_VER) && (_MSC_VER >= 1600)))
+#define CPPUTEST_COMPILER_FULLY_SUPPORTS_CXX11
 #define _override override
 #else
 #define _override
--- a/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
+++ b/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
@@ -64,7 +64,7 @@ void IEEE754ExceptionsPlugin::enableInexact()
 
 void IEEE754ExceptionsPlugin::ieee754Check(UtestShell& test, TestResult& result, int flag, const char* text)
 {
-    gitresult.countCheck();
+    result.countCheck();
     if(std::fetestexcept(flag)) {
         std::feclearexcept(FE_ALL_EXCEPT);
         CheckFailure failure(&test, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", text);
--- a/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
+++ b/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
@@ -32,7 +32,7 @@
 
 #include <cfenv>
 
-#define IEEE754_CHECK_CLEAR(flag) ieee754Check(flag, #flag)
+#define IEEE754_CHECK_CLEAR(test, result, flag) ieee754Check(test, result, flag, #flag)
 
 void IEEE754ExceptionsPlugin::preTestAction(UtestShell&, TestResult&)
 {
@@ -42,14 +42,12 @@ void IEEE754ExceptionsPlugin::preTestAction(UtestShell&, TestResult&)
 void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
 {
     hasFailed_ = test.hasFailed();
-    test_ = &test;
-    result_ = &result;
-    IEEE754_CHECK_CLEAR(FE_DIVBYZERO);
-    IEEE754_CHECK_CLEAR(FE_OVERFLOW);
-    IEEE754_CHECK_CLEAR(FE_UNDERFLOW);
-    IEEE754_CHECK_CLEAR(FE_INVALID);
+    IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO);
+    IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW);
+    IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW);
+    IEEE754_CHECK_CLEAR(test, result, FE_INVALID);
     if (inexactEnabled_) {
-        IEEE754_CHECK_CLEAR(FE_INEXACT);
+        IEEE754_CHECK_CLEAR(test, result, FE_INEXACT);
     }
 }
 
@@ -63,13 +61,13 @@ void IEEE754ExceptionsPlugin::enableInexact()
     inexactEnabled_ = true;
 }
 
-void IEEE754ExceptionsPlugin::ieee754Check(int flag, const char* text)
+void IEEE754ExceptionsPlugin::ieee754Check(UtestShell& test, TestResult& result, int flag, const char* text)
 {
     if(!hasFailed_) {
-        result_->countCheck();
+        result.countCheck();
         if(std::fetestexcept(flag)) {
-            CheckFailure failure(test_, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", text);
-            result_->addFailure(failure);
+            CheckFailure failure(&test, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", text);
+            result.addFailure(failure);
             hasFailed_ = true;
         }
     }
--- a/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
+++ b/src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
@@ -41,13 +41,14 @@ void IEEE754ExceptionsPlugin::preTestAction(UtestShell&, TestResult&)
 
 void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
 {
-    hasFailed_ = test.hasFailed();
-    IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO);
-    IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW);
-    IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW);
-    IEEE754_CHECK_CLEAR(test, result, FE_INVALID);
-    if (inexactEnabled_) {
-        IEEE754_CHECK_CLEAR(test, result, FE_INEXACT);
+    if(!test.hasFailed()) {
+        IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO);
+        IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW);
+        IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW);
+        IEEE754_CHECK_CLEAR(test, result, FE_INVALID);
+        if (inexactEnabled_) {
+            IEEE754_CHECK_CLEAR(test, result, FE_INEXACT);
+        }
     }
 }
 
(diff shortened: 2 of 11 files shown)


src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL <copyright holder> BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#if defined(__cplusplus) && ((__cplusplus >= 201103L) || (defined(_MSC_VER) && (_MSC_VER >= 1600)))
```
basvodde: I thought we were going to move this to the CppUConfig.h ?

arstrube: 1. Agreed
2. Nothing new here, just a rebase
3. You mean config.h / autoconf? Not really sure how to do it properly

basvodde: For now, just in CppUConfig.h. Later it would need to be done better. But if all are there, it is a lot easier.

Also, the build seems to fail...

I'll re-do the macro after the build passes :)

tests/CppUTestExt/IEEE754PluginTest_c.c
```diff
@@ -0,0 +1,69 @@
...
+    (void) f;
+}
+
+void set_inexact_c(void) {
+/* feraiseexcept(FE_INEXACT); */ /* Clang ignores -frounding-math */
+    float f = 10.0f;
+    f /= 3.0f;
+    (void) f;
+}
+
```
arstrube: @basvodde I am have trouble making the above work with compiler optimizations (-O2) on. The only reliable alternative is setting the flags directly, which kind of defeats the purpose (which is, among other things, to test that the flags will be set doing ordinary operations in ordinary C code).

arstrube: This is the reason for test failures.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Umm.... where is CppUConfig.h ???

arstrube: Oookay. CppUTestConfig.h

tests/CppUTestExt/IEEE754PluginTest_c.c
basvodde: Uhm, interesting suggestion on stack overflow is to use volatile.

http://stackoverflow.com/questions/2219829/how-to-prevent-gcc-optimizing-some-statements-in-c

Might want to try that!

arstrube: I did. It didn't work simply making the local variables volatile, but making f both global and volatile did work (at least on my computer). Saves me from having to do (void)f as well, so I think it's good now.

arstrube:
@basvodde I'd like to squash this into one commit once Travis passes.

basvodde:
Oki, fine. Just saw when.

basvodde:
I see it is done already :)

include/CppUTest/CppUTestConfig.h
```diff
@@ -178,6 +178,7 @@
 
 /* Visual C++ 10.0+ (2010+) supports the override keyword, but doesn't define the C++ version as C++11 */
 #if defined(__cplusplus) && ((__cplusplus >= 201103L) || (defined(_MSC_VER) && (_MSC_VER >= 1600)))
+#define COMPILER_FULLY_SUPPORTS_CXX11
```
basvodde: Excellent. Next time, perhaps have a CPPUTEST prefix still :)

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#define IEEE754_CHECK_CLEAR(flag) { \
+    if(!hasFailed_) { \
+        result_->countCheck(); \
+        if(flag) { \
+            CheckFailure failure(test_, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", #flag); \
+            result_->addFailure(failure); \
+            hasFailed_ = true; \
+        } \
+    } \
```
basvodde: Won't merge it yet, will still propose a way to get rid of this macro.

include/CppUTest/CppUTestConfig.h
arstrube: Oki

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Please do. I could only come up with ugly alternatives :-P

arstrube:
Regarding `|=` we would get this interesting construct:
```c++
void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
{
    bool hasFailed = test.hasFailed();
    hasFailed |= !hasFailed ? IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO) : false;
    hasFailed |= !hasFailed ? IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW) : false;
    hasFailed |= !hasFailed ? IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW) : false;
    hasFailed |= !hasFailed ? IEEE754_CHECK_CLEAR(test, result, FE_INVALID) : false;
    if (inexactEnabled_) {
        hasFailed |= !hasFailed ? IEEE754_CHECK_CLEAR(test, result, FE_INEXACT) : false;
    }
}
```
Or, we could do the stunningly strange:
```c++
void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
{
    if(!test.hasFailed())
        if(IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO))
            if(IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW))
                if(IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW))
                    if(IEEE754_CHECK_CLEAR(test, result, FE_INVALID))
                       if (inexactEnabled_)
                           IEEE754_CHECK_CLEAR(test, result, FE_INEXACT);
}
```
Both work, and so would adding another parameter, but, honestly, I prefer the attribute.
BTW the most elegant way would be to use `failWith()` on the test and to check test.hasFailed() inside the function call. Unfortunately, this leads to side-effects that are bad for testing (will fail in current test, rather than in TestTestingFixture). If anyone knows how to get around that last problem (short of using fork()) that would be nice.
```

**Comment by arstrube** [comment]
The least ugly being in i/o parameter, thus:
```c++
void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
{
    bool failed = test.hasFailed();
    IEEE754_CHECK_CLEAR(&failed, test, result, FE_DIVBYZERO);
    IEEE754_CHECK_CLEAR(&failed, test, result, FE_OVERFLOW);
    IEEE754_CHECK_CLEAR(&failed, test, result, FE_UNDERFLOW);
    IEEE754_CHECK_CLEAR(&failed, test, result, FE_INVALID);
    if (inexactEnabled_)
        IEEE754_CHECK_CLEAR(&failed, test, result, FE_INEXACT);
}
```
but honestly, I am in favor of keeping the attribute.

basvodde:
Can you change failed as a return value. I don't think you need to pass it as an IO variable. Then you can move the if outside and just make it one if or multiple via ? :

arstrube:
Umm - did you see my last commit?
```c++
void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
{
    if(!test.hasFailed()) {
        IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO);
        IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW);
        IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW);
        IEEE754_CHECK_CLEAR(test, result, FE_INVALID);
        if (inexactEnabled_) {
            IEEE754_CHECK_CLEAR(test, result, FE_INEXACT);
        }
    }
}
```
and:
```c++
void IEEE754ExceptionsPlugin::ieee754Check(UtestShell& test, TestResult& result, int flag, const char* text)
{
    result.countCheck();
    if(std::fetestexcept(flag)) {
        std::feclearexcept(FE_ALL_EXCEPT);
        CheckFailure failure(&test, __FILE__, __LINE__, "IEEE754_CHECK_CLEAR", text);
        result.addFailure(failure);
    }
}
```
I just clear all flags, once any has been detected as set.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+void IEEE754ExceptionsPlugin::preTestAction(UtestShell&, TestResult&)
+{
+    std::feclearexcept(FE_ALL_EXCEPT);
+}
+
+void IEEE754ExceptionsPlugin::postTestAction(UtestShell& test, TestResult& result)
+{
+    if(!test.hasFailed()) {
+        IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO);
```
basvodde: This doesn't work, now the test can fail multiple times. You need to return a value and use || and put the calls in the if, then it works fine.

arstrube:
>Can you change failed as a return value. I don't think you need to pass it as an IO variable. Then you can move the if outside and just make it one if or multiple via ? : 

I've already done both the "if"-thing and the "? :"-thing (see above) and I think the code looks pretty hideous either way. Did you have something else in mind?

I like commit 49e5de3 better than any of those options.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: The test doesn't fail multiple times. There's a test case to prove that, too.

basvodde:
```
if(!test.hasFailed() || 
  ! IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO) ||
  ! IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW) || 
  ! IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW) ||
  ! IEEE754_CHECK_CLEAR(test, result, FE_INVALID) ||
  ! (inexactEnabled_ && IEEE754_CHECK_CLEAR(test, result, FE_INEXACT)) {
   ...
```

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
basvodde: Why wouldn't it fail with multiple failures?

arstrube:
I like the "...". Because there is nothing to go there. The complete code will then be:
```c++
if(!test.hasFailed() || 
    ! IEEE754_CHECK_CLEAR(test, result, FE_DIVBYZERO) ||
    ! IEEE754_CHECK_CLEAR(test, result, FE_OVERFLOW) || 
    ! IEEE754_CHECK_CLEAR(test, result, FE_UNDERFLOW) ||
    ! IEEE754_CHECK_CLEAR(test, result, FE_INVALID) ||
    ! (inexactEnabled_ && IEEE754_CHECK_CLEAR(test, result, FE_INEXACT)) {
        ;
}
``` :P

**Comment by basvodde** [comment]
When the if can be sort-of removed :)

File: src/CppUTestExt/IEEE754ExceptionsPlugin.cpp [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+void IEEE754ExceptionsPlugin::enableInexact()
+{
+    inexactEnabled_ = true;
+}
+
+void IEEE754ExceptionsPlugin::ieee754Check(UtestShell& test, TestResult& result, int flag, const char* text)
+{
+    result.countCheck();
+    if(std::fetestexcept(flag)) {
+        std::feclearexcept(FE_ALL_EXCEPT);
```
arstrube: Because of this line. Once any flag is caught, all flags are cleared :)

arstrube:
> When the if can be sort-of removed :)

Would removing the if be syntactically correct? Computing a bool and not using it for anything?

basvodde:
It would be correct. But you are right, the clearing of the flag solves it!

PR #858, opened 2016-02-05
Correct use of ordinal numbers in MockUnexpectedCallHappenedFailure

hbock:
Previously, all such failures displayed a failure message containing "Unexpected additional (Nth) call", even for values such as 2, 3, etc.

This patch correctly differentiates based on English ordinal numbering rules, adding tests for all the special cases I could think of.  If I should break it into multiple tests, let me know!

(sorry, this must be the most nit-picky pull request of all time...)

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -653,6 +653,26 @@ SimpleString StringFromMaskedBits(unsigned long value, unsigned long mask, size_
     return result;
 }
 
+SimpleString StringFromOrdinalNumber(unsigned int number)
+{
+    unsigned int onesDigit = number % 10;
+
+    const char* suffix;
+    if (number >= 11 && number <= 13) {
+        suffix = "th";
+    } else if (3 == onesDigit) {
+        suffix = "rd";
+    } else if (2 == onesDigit) {
+        suffix = "nd";
+    } else if (1 == onesDigit) {
+        suffix = "st";
+    } else {
+        suffix = "th";
+    }
+
+    return StringFromFormat("%u%s", number, suffix);
+}
+
 SimpleStringCollection::SimpleStringCollection()
 {
     collection_ = 0;
--- a/src/CppUTestExt/MockFailure.cpp
+++ b/src/CppUTestExt/MockFailure.cpp
@@ -103,11 +103,12 @@ MockExpectedCallsDidntHappenFailure::MockExpectedCallsDidntHappenFailure(UtestSh
 
 MockUnexpectedCallHappenedFailure::MockUnexpectedCallHappenedFailure(UtestShell* test, const SimpleString& name, const MockExpectedCallsList& expectations) : MockFailure(test)
 {
-    int amountOfExpectations = expectations.amountOfExpectationsFor(name);
-    if (amountOfExpectations)
-        message_ = StringFromFormat("Mock Failure: Unexpected additional (%dth) call to function: ", amountOfExpectations+1);
-    else
+    if (expectations.amountOfExpectationsFor(name)) {
+        SimpleString ordinalNumber = StringFromOrdinalNumber((unsigned)(expectations.amountOfExpectationsFor(name) + 1));
+        message_ = StringFromFormat("Mock Failure: Unexpected additional (%s) call to function: ", ordinalNumber.asCharString());
+    } else {
         message_ = "Mock Failure: Unexpected call to function: ";
+    }
     message_ += name;
     message_ += "\n";
     addExpectationsAndCallHistory(expectations);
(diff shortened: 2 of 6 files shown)


basvodde:
Funny. Could you extract the code out of there though and put it in either SimpleString, perhaps better, create a OrdinalNumberStringFrom method in SimpleString :)

hbock:
That's a much better idea! :) I added a new free function StringFromOrdinalNumber() to match the pattern of other utilities.  The tests have been ported over to SimpleStringTests, and I left a few of the original tests in MockFailureTests to "prove" the new helper is being used.

src/CppUTestExt/MockFailure.cpp
```diff
@@ -104,10 +104,15 @@ MockExpectedCallsDidntHappenFailure::MockExpectedCallsDidntHappenFailure(UtestSh
 MockUnexpectedCallHappenedFailure::MockUnexpectedCallHappenedFailure(UtestShell* test, const SimpleString& name, const MockExpectedCallsList& expectations) : MockFailure(test)
 {
     int amountOfExpectations = expectations.amountOfExpectationsFor(name);
-    if (amountOfExpectations)
```
cuitoldfish: hi, would it be better if replacing temp variables `amountOfExpectations` and `actualCallCount`with directly function call ``expectations.amountOfExpectationsFor``?  The direct function is already clear and those temp variables is not indispensable.

hbock: I'm not opposed to removing them, but I think it slightly improves readability and doesn't hurt generated code.

If you're suggesting calling amountOfExpectationsFor() twice, I suppose it won't be an issue, since it's a const method.  The compiler will elide the second call, correct?

cuitoldfish: I mean just for improving readability :)

cuitoldfish:
I just think here algorithem is not complex enough to deserve the explanation variables, in this case direct mean clear. But it doesnt have smell currently, we can refactying when the smell get stronger.

src/CppUTest/SimpleString.cpp
```diff
@@ -653,6 +653,30 @@ SimpleString StringFromMaskedBits(unsigned long value, unsigned long mask, size_
...
+{
+    unsigned int onesDigit = number % 10;
+
+    const char* suffix;
+    if (number >= 11 && number <= 13) {
+        // all ordinal numbers in the 10s use 'th'.
+        suffix = "th";
+
+    // all other ordinals follow the same pattern based on the
+    // ones digit: 1 = "1st", 2 = "2nd", 3 = "3rd", rest = "4th"
```
basvodde: Could you remove the comments? They are pretty obvious and keeping them there is no use. In general, in CppUTest no comments are preferred over useless ones.

src/CppUTestExt/MockFailure.cpp
basvodde: I'd agree, I'd prefer not adding useless local variables, that doesn't really improve readibility.

src/CppUTestExt/MockFailure.cpp
```diff
@@ -104,10 +104,15 @@ MockExpectedCallsDidntHappenFailure::MockExpectedCallsDidntHappenFailure(UtestSh
...
-        message_ = StringFromFormat("Mock Failure: Unexpected additional (%dth) call to function: ", amountOfExpectations+1);
-    else
+    if (amountOfExpectations) {
+        // This count should never be negative.
```
basvodde: Please remove such comments.

src/CppUTestExt/MockFailure.cpp
```diff
@@ -103,11 +103,12 @@ MockExpectedCallsDidntHappenFailure::MockExpectedCallsDidntHappenFailure(UtestSh
 
 MockUnexpectedCallHappenedFailure::MockUnexpectedCallHappenedFailure(UtestShell* test, const SimpleString& name, const MockExpectedCallsList& expectations) : MockFailure(test)
 {
-    int amountOfExpectations = expectations.amountOfExpectationsFor(name);
-    if (amountOfExpectations)
-        message_ = StringFromFormat("Mock Failure: Unexpected additional (%dth) call to function: ", amountOfExpectations+1);
-    else
+    if (expectations.amountOfExpectationsFor(name)) {
+        SimpleString ordinalNumber = StringFromOrdinalNumber((unsigned)(expectations.amountOfExpectationsFor(name) + 1));
```
hbock: I personally think this is much less readable than adding one local variable, but to each their own :)

hbock:
I should've asked this before - until now I've been rebasing with fixup/squash on my PR branch to keep it to one commit.  In the future, would you prefer to keep the fixup commits based on comments?  I don't like littering other projects' repositories with redo commits, but I could see you wanting the review history too.

basvodde:
If you want to fixup, go ahead. I personally don't care much, I prefer the code to be good more than the history :P

basvodde:
Let me know if you want to clean up the history. Also, could you remove the unneeded comments still from the test. The rest looks ok.

basvodde:
Thanks

PR #862, opened 2016-02-10
Get rid of C++11 requirement for IEEE754Plugin

arstrube:
This one still faces at least two challenges:

1. Building without Standard C libraries fails (naturally) -- @basvodde what would be the cleanest way to deal with that? PlatformSpecifics function, perhaps?

2. Open Watcom does support `fenv.h`. However, I couldn't convince it to link the fe... functions... yet.

### DIFF
--- a/tests/CppUTestExt/IEEE754PluginTest.cpp
+++ b/tests/CppUTestExt/IEEE754PluginTest.cpp
@@ -27,14 +27,13 @@
 
 /* The IEEE754ExceptionFlags plugin by definition requires C++11 */
 
-#ifdef CPPUTEST_COMPILER_FULLY_SUPPORTS_CXX11
-
 #include "CppUTest/CommandLineTestRunner.h"
 #include "CppUTest/TestHarness.h"
 #include "CppUTest/TestRegistry.h"
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTestExt/IEEE754ExceptionsPlugin.h"
-#include <fenv.h>
+
+#if CPPUTEST_USE_STD_C_LIB
 
 extern "C" { 
     #include "IEEE754PluginTest_c.h"
--- a/tests/CppUTestExt/IEEE754PluginTest.cpp
+++ b/tests/CppUTestExt/IEEE754PluginTest.cpp
@@ -25,15 +25,13 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/* The IEEE754ExceptionFlags plugin by definition requires C++11 */
-
 #include "CppUTest/CommandLineTestRunner.h"
 #include "CppUTest/TestHarness.h"
 #include "CppUTest/TestRegistry.h"
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTestExt/IEEE754ExceptionsPlugin.h"
 
-#if CPPUTEST_USE_STD_C_LIB
+#if CPPUTEST_FENV_IS_WORKING_PROPERLY
 
 extern "C" { 
     #include "IEEE754PluginTest_c.h"
@@ -65,13 +63,19 @@ TEST(FE__with_Plugin, should_fail____when__FE_UNDERFLOW__is_set) {
     fixture.assertPrintContains("IEEE754_CHECK_CLEAR(FE_UNDERFLOW) failed");
 }
 
-TEST(FE__with_Plugin, should_fail____when__FE_INVALID____is_set) {
+#ifndef __MINGW64__
+#define NOT_MINGW64_TEST TEST
+#else
+#define NOT_MINGW64_TEST IGNORE_TEST
+#endif
+
+NOT_MINGW64_TEST(FE__with_Plugin, should_fail____when__FE_INVALID____is_set) {
     fixture.setTestFunction(set_invalid_c);
     fixture.runAllTests();
     fixture.assertPrintContains("IEEE754_CHECK_CLEAR(FE_INVALID) failed");
 }
 
-TEST(FE__with_Plugin, should_fail____when__FE_INEXACT____is_set_and_enabled) {
+NOT_MINGW64_TEST(FE__with_Plugin, should_fail____when__FE_INEXACT____is_set_and_enabled) {
     ieee754Plugin.enableInexact();
     fixture.setTestFunction(set_inexact_c);
     fixture.runAllTests();
--- a/tests/CppUTestExt/IEEE754PluginTest.cpp
+++ b/tests/CppUTestExt/IEEE754PluginTest.cpp
@@ -75,7 +75,7 @@ NOT_MINGW64_TEST(FE__with_Plugin, should_fail____when__FE_INVALID____is_set) {
     fixture.assertPrintContains("IEEE754_CHECK_CLEAR(FE_INVALID) failed");
 }
 
-NOT_MINGW64_TEST(FE__with_Plugin, should_fail____when__FE_INEXACT____is_set_and_enabled) {
+TEST(FE__with_Plugin, should_fail____when__FE_INEXACT____is_set_and_enabled) {
     ieee754Plugin.enableInexact();
     fixture.setTestFunction(set_inexact_c);
     fixture.runAllTests();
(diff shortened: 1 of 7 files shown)


tests/CppUTestExt/IEEE754PluginTest.cpp
```diff
@@ -27,59 +27,67 @@
...
 #include "CppUTest/TestRegistry.h"
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTestExt/IEEE754ExceptionsPlugin.h"
-#include <cfenv>
 
 extern "C" { 
     #include "IEEE754PluginTest_c.h"
 }
 
+#if CPPUTEST_USE_STD_C_LIB
```
basvodde: Why not just putting all tests in the #if. Thats how we do this in other places. No need to ignore tests for functionality that doesn't exists :)

arstrube: You have a point, of course. Wrapping all the tests in one #if does make things simpler.

However, doing so makes it increasingly difficult to know at a glance whether certain tests were run or not. Always using IGNORE_TEST has two advantages:

1. The total number of tests for different configurations should always be the same. This can be verified at a glance.

2. The presence / absence of certain ignored tests is also easy to verify.

It is possible (and actually happened to me just now) that a #define that you thought to be 1 actually was 0 (because of wrong placement, before the CppUTest headers). So it can happen that the functionality is included, but the tests are not run. Without the ignored tests, most likely, you will never know. Fortunately, I had the opposite problem and got a linker error ;-).

arstrube:
What's your take on having `PlatformSpecificFEClearExcept()` and friends though? Really things like `fenv.h` shouldn't be included in CppUTest code directly, no? Do we make exceptions for plugins?

basvodde:
We would need to create a PlatformSpecificExtFunctions as we wouldn't want to move the dependency to the core CppUTest code when it isn't used.

tests/CppUTestExt/IEEE754PluginTest.cpp
basvodde: I'd prefer not to have more of the IGNORE_TEST for this reason and keep it simpler. I guess there are better ways to discover code under #if that isn't used. Must be tools for that that we can integrate in the build. (as it just requires the C preprocessor)

arstrube: I guess it's not such a big deal. My bigger problem is that the Watcom linker doesn't seem to find the functions :-(

arstrube:
Good point. And right now, probably pointless.....

arstrube:
Two remaining issues:

1. I can't get MinGW 64 to compute anything as "inexact". Ideas?
2. Weird failures in the Dos build -perhaps the flags are not reset properly?

arstrube:
After much experimenting and the latest squash, the code is now clean. It took me a while to figure out that Open Watcom optimized away the call to feclearexcept(), because its return value wasn't used...

There are two problems still:

1. MinGW 64 bit will not raise FE_INEXACT upon 10.0f / 3.0f (or the like).

2. On the Dos platform, any kind of floating point error, no matter which, will always raise FE_OVERFLOW

@basvodde - any suggestions on how to handle this? Ifdef / IGNORE_TEST would be a possible solution for the first issue. The second one could only be addressed by running an entirely different (set of) tests, or not running any tests. Or....?

BTW these problems didn't occur with the previous pull, because the configuration was such that the Plugin wouldn't run in those configurations.

arstrube:
Resolved to not run any tests on MsDos, and to IGNORE_TEST() the inexact test on MinGW64.
This should do for now; perhaps someone knows a way of influencing the FP behavior of MinGW64, though?

arstrube:
@Andne -- I need some help singling out the "Environment: PlatformToolset=MinGW, Platform=x64" job. I tried:
```c++
#if defined(__MINGW32__) && defined(__WIN64__)
```
but it doesn't work, presumably because the platform is 64bit, not the target. Any idea?

Andne:
In Appveyor the platform is always 64-bit, but the target switches.  'Platform=x64' is supposed to mean that the target being compiled for is x64.  The variable name comes from what Visual C++ needs defined for its build scripts.

Some quick hunting around implies that `__WIN64__` should exist, otherwise you could also try `__MINGW64__` since it looks like you're only trying to detect MinGW 64-bit and not Visual C++ 64-bit.  Beyond that I'd have to spend some time looking at it myself to figure it out.  Looking at your code, it seems that you've already tried that though, so I'm not sure what's happening here.

arstrube:
Yes that's right, so far only MinGW 64 bit is giving me grief. But then I haven't tackled Visual C++ yet....

Andne:
Um, check the order of your `#ifdef` block, I think you have `TEST` and `IGNORE_TEST` in the wrong branches.  I dumped the preprocessor information for MinGW 64bit and found `__MINGW32__, __MINGW64__, __WIN64__, __WIN32__` and several other similar variants all defined.

arstrube:
Yes, you are right. My bad. I had the #define logic reversed, so __MINGW64__ compiled precisely the test it shouldnt compile...

But, strangely, after I fixed it, the test still gets compiled:
```c++
#ifndef __MINGW64__
#define NOT_MINGW64_TEST TEST
#else
#define NOT_MINGW64_TEST IGNORE_TEST
#endif
```
According to my understanding, it shouldn't have compiled anymore. But apparently it did? (arstrube/cpputest/tmp)

Andne:
I get a failure on an earlier test - `should_fail___when__FE_INVALID___is_set` - even if I replace `NOT_MINGW64_TEST` on the `FE_INEXACT` test, I don't get a failure from it.

I also discovered that when using MinGW Makefiles from CMake, `make test` doesn't actually build any code changes.

Andne:
@arstrube I've been chasing around why `sqrt(-1.0f)` doesn't trigger an `FE_INVALID` exception and have no idea why it's not working properly.  However, I did find that `feraiseexcept()` works.  Is there a specific reason that you didn't use that?

Also, I've learned why you have `FE_INEXACT` disabled by default - it seems to pretty much always be set, even when I think no floating-point code has executed since the last clear.

I ran this through Visual Studio as well, 2010 doesn't have `fenv.h`, 2015 does, I'll have to whittle it down to somewhere in the middle at some point I suppose.  Guess I should add a few newer versions to the appveyor build sometime.

Andne:
I checked, Visual C++ 64-bit doesn't have the same issue with `FE_INVALID` that MinGW does.  Currently we don't have any 64-bit Visual Studio builds running.  Do we want to change that?

arstrube:
Ideally both should pass :-)

arstrube:
I guess it was late at night - I didn't pick up that the failure had changed ;-). My immediate response would be to disable both for mingw64. I am beginning to think that it is a compiler bug and nothing to do with running 64 bit....

As to why I am not using `feraiseexcept()` -- actually, I did do that at some point and it makes the tests so nice and simple (no C file or header, either)... The point here is to make sure the flags gets set and handled properly even when the code under test is written in C and doesn't "know" anything about `fenv.h`. Using feraiseexcept() doesn't demonstrate that. It oculd be argued that the plugin is still tested properly without, but I prefer having it that way.

An interesting side effect in the examples was that an "inexact" flag was raised in some other example code (interestingly, an innocent DOUBLES_EQUAL macro), because I had enabled it in the main().

Andne:
Based on my testing, `FE_INEXACT` does work, it's only `FE_INVALID` that has issues.

Maybe we should have both sets of tests - the ones that use `feraiseerror()` to test that the flags all get checked and the ones that use the c file to show the more practical application/usage.

My only other idea is that the optimizer isn't using the FPU in that function and so the flags work differently as a result.  Any idea how to trigger the exception without calling a library function?  That maybe negates part of your goal though if we can't show the more practical uses of the plugin.

Andne:
I think you said another time you didn't do this because it doesn't exist on OSX, but maybe `feenableexcept(FE_INVALID)` needs to be called for MinGW 64-bit?

arstrube:
@Andne 

1. So I could make FE_INEXACT a normal test - barking up the wrong tree the whole time!
2. I would be open to two sets of tests, but the "practical" one would always have some issues.
3. It would be nice to trigger FE_INVALID without library function - I just don't know how. It would be even nicer if this made the mingw problem go away.
4. feenableexcept() causes a system trap. That doesn't help us here.

Finally, and above all, lots of thanks for your help and ideas. Sometimes it takes someone else to see the obvious :)

arstrube:
@Andne I would like to add wrappers for signalling upon fp error, but ran into a small problem with Os X and a bigger problem with Watcom (would need 2.0, which is inofficial and I don't know that we can get stuff from Sourceforge on Travis CI).

Causing a trap is useful for debugging the source of a problem, because it will stop at the exact point where the problem occured. This is not useful in normal running of tests, when we want the tests all to run and finish.

arstrube:
@basvodde after fighting all the compiler dragons & platform monsters, I think this bit is ready now for you to inspect :). I was thinking of removing the CHECK again from feclearexcept() - what's your take? If I leave them in, I think there's one place that still needs one. So, either way, one more small change. The question is which one.

arstrube:
@Andne regarding enabling / disabling floating point signal traps, you might find this interesting: https://github.com/arstrube/cpputest/tree/new.

It's not a pull request, because it has some issues I don't know how to solve.

PR #866, opened 2016-02-13
Make LONGS_EQUAL(_TEXT)() and UNSIGNED_LONGS_EQUAL(_TEXT)() expression safe

hbock:
This change ensures expected/actual are put in parenthesis before calling (UNSIGNED_)LONGS_EQUAL_LOCATION().  This makes these macros expression-safe, equivalent to their counterparts (CHECK et al, POINTERS_EQUAL, etc.)

This prevents unwanted truncation and silences PC-Lint if the user writes a check like:
LONGS_EQUAL(c, a + b);

FWIW if the user uses the GCC flag -Wconversion=error, the above would cause a compilation error if a would lose precision when cast to (long) per the expansion:

assertLongsEqual(c, (long)a + b, ...)

### DIFF
--- a/tests/TestUTestMacro.cpp
+++ b/tests/TestUTestMacro.cpp
@@ -421,6 +421,22 @@ static int _countingMethod()
     return countInCountingMethod++;
 }
 
+TEST(UnitTestMacros, LONGS_EQUAL_macroExpressionSafety)
+{
+    LONGS_EQUAL(1, 0.4 + 0.7);
+    LONGS_EQUAL(0.4 + 0.7, 1);
+    LONGS_EQUAL_TEXT(1, 0.4 + 0.7, "-Wconversion=great");
+    LONGS_EQUAL_TEXT(0.4 + 0.7, 1, "-Wconversion=great");
+}
+
+TEST(UnitTestMacros, UNSIGNED_LONGS_EQUAL_macroExpressionSafety)
+{
+    UNSIGNED_LONGS_EQUAL(1, 0.4 + 0.7);
+    UNSIGNED_LONGS_EQUAL(0.4 + 0.7, 1);
+    UNSIGNED_LONGS_EQUAL_TEXT(1, 0.4 + 0.7, "-Wconversion=great");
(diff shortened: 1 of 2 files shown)


tests/TestUTestMacro.cpp
```diff
@@ -421,6 +421,28 @@ static int _countingMethod()
     return countInCountingMethod++;
 }
 
+TEST(UnitTestMacros, LONGS_EQUAL_macroExpressionSafety)
+{
+    // If the expansion of arguments to LONGS_EQUAL() are not parenthesized,
```
basvodde: Could you remove the comments, and instead ensure that the test case name describes the case so that no additional comments are needed?

hbock:
Sure. What should they be named, I'm not sure what would make it clearer.

basvodde:
If you think it is clear enough then the cast can be removed. Or else add something like SoThatForExampleACastAppliesToWholeExpressionAndNotTheFirstArgument :)

hbock:
Hahaha, I think I'd rather have the comment than that name :) I removed the comment and left the original name; it should be clear enough as-is.

basvodde:
@arstrube Can you have a look at this failure. It seems to be because of the FE_Plugin where the output is expected to be run in 0 ms and it ran in 1 ms :) Could you remove the test-run from the check?

@hbock Sorry, let me have a look at the weird error before merging this :)

arstrube:
I have noticed this. There are probably more places where "0 ms" is part of an expected string. For some reason, this one actually does take the occasional "1 ms". It would probably be a good idea to ignore the time everywhere? Or should I just fix this test?

P.S. Usually when you restart the job, it passes.

basvodde:
Yup, time ought to be removed :) Don't want tests to fail on a slow computer.

Ps. That is why I didn't restart it :P

arstrube:
Great candidate for -r100, with or without random ;-)

arstrube:
Oops. Ended up in the other PR unintentiaonally....

Issue #863, opened 2016-02-11
Return default value when mock disabled or ignored


grantramsay:
Hi, I was following the [example on website](http://cpputest.github.io/mocking_manual.html#other_mock_support) about `returnIntValueOrDefault` but I can't get it to work.
e.g.

    TEST(none, returns_default_when_mock_disabled)
    {
        mock().disable();
        LONGS_EQUAL(5, mock().actualCall("function").returnIntValueOrDefault(5));
    }

> error: Failure in TEST(none, returns_default_when_mock_disabled)
expected <5 0x5>
but was  <0 0x0>

I get the same thing with `mock().ignoreOtherCalls();`.
Currently using version 32dd4276545dd947e166b6bc9ae646f1e2895973

Am I using this correctly?

arstrube:
I think this is expected behavior. We should probably add test cases for this. The following should work when mocks are disabled:
```c++
TEST(none, returns_default_when_mock_disabled)
{
    mock().disable();
    LONGS_EQUAL(5, mock().returnIntValueOrDefault(5));
}
```

arstrube:
The reason is that `mock().disable()` will cause instances of `MockCheckedActualCall` to be replaced by instances of `MockIgnoredActualCall`, which only has dummy methods.

I think there should probably be at least one test case somewhere to demonstrate this.

arstrube:
TEST(none, mock_returns_default_when_mock_disabled_but_actualCall_doesnt)
{
    mock().disable();
    LONGS_EQUAL(5, mock().returnIntValueOrDefault(5));
    LONGS_EQUAL(0, mock().actualCall("function").returnIntValueOrDefault(5));
}

grantramsay:
So is that expected behavior?

The [example on website](http://cpputest.github.io/mocking_manual.html#other_mock_support) says:
> int function () {
    return mock().actualCall("function").returnIntValueOrDefault(5);
}
If the function is ignored or the framework is disabled, a 5 will be returned, otherwise the expected return value will be returned (according to the expectations set on the test case).

Return value logic like this would seem nice to me:

    if mock_enabled && expected_value_set
        return expected_value
    else if default_value_set
        return default_value
    else
        return 0 // or NULL etc.

arstrube:
Hmm. Clearly, the documentation and the real behavior are different. Let's hear out @basvodde which one should be fixed. Fixing the behavior won't be as easy as the above code suggests.

basvodde:
[...]

PR #867, opened 2016-02-15
Fix default return value of ignored actual calls

arstrube:
Fix for issue  #863


arstrube:
Sorry this is messed up. Will fix later.

Issue #869, opened 2016-02-15
Appveyor / mingw problem


arstrube:
Appveyor now loads mingw with Gcc 5.3.0 instead of 4.8.3.

A "new" familiar, strange compiler error is showing up. I don't have 5.3.0 so it's a bit difficult to know what is going on.

Andne:
@arstrube How do you want to treat this, do you want to fix the compiler error or do you want to force a different version of GCC?

arstrube:
Fixing seems impossible - some compiler always barks at something. The cleanest way would be via \<cmath\> but can't be done because requires std c++ lib. I actually consider this a bug in GCC, there's supposed to be a macro which determines whether it's being passed a float or a double and compile the correct code.

I am busy experimenting with a suitable #pragma (got it wrong several times, but now probably correct. This way, we could actually use GCC 5.3.0.

Forcing the old version would probably be the easiest (or at least a version that doesn't exhibit this annoying problem).

An unrelated point is that since the change, unzip spits out about 1MB of useless messages, making it very hard to scroll to the lines that are actually interesting. Perhaps this could be suppressed.

Maybe by tomorrow, the pragma will work. Let's perhaps also hear out @basvodde on this. It seems pretty much a choice between pragma (if that works) or forcing the old version.

(This is one of those annoying things where it is notoriously difficult to find information or be heard on StackOverflow and the like, because most people getting these error get them for a good reason, not like here. The errors are also strange. double is converted to float, not the other way round, and how come the compiler is whining about int totally escapes me. It did that some time ago, and at that time, casting to float for some reason made it shut up (also rather a hack, really, because isnan() should take a double without complaining).

arstrube:
P.s. apologies for my rambling, and your help is much appreciated!

Andne:
The install has been spitting out a huge log for a while now, looks like chocolately has changed that behavior recently but there's not a new version to use yet.  I can probably suppress the output from the command in general, but I'm not sure I want to lose the normal information or any error messages.

I've force GCC to 4.8.5 and it appears to build correctly now, I'll have a pull request soon, not sure if you want to fix the code or just use the older GCC version.

basvodde:
Fix the code so it works with both versions, please ;)

arstrube:
@basvodde it seems the #pragma is now doing its magic right with 5.3.0

[...]

PR #870, opened 2016-02-16
Force GCC 4.8.5 on appveyor

Andne:
This is one solution to #869. It forces GCC to install version 4.8.5 which doesn't have any errors during the build.

@arstrube Your choice as to if you want this solution or to come up with a code change that works properly.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,7 +14,7 @@ environment:
   - PlatformToolset: v100
 
 install:
-- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw $( if ($env:Platform -eq 'Win32') { Write-Output -forcex86 } ) }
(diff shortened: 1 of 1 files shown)


basvodde:
Can we look at getting to work propoerly with the new gcc? :)

Andne:
Ok, I'll see if I can make both run to make sure we don't break anything.

arstrube:
@Andne not doing this anymore? I guess the failure was due to my fix not being merged yet. Couldn't get myself to scroll down to check :P ...

Andne:
It sounded like @basvodde wanted to fix the error.  I did make it run both versions of the compiler, so I can reopen/resubmit it if somone wants that, but the original purpose didn't seem to apply anymore.

If we want to run both versions, I'll rebase it when I get a chance so that it runs correctly.

arstrube:
I guess both verisons can't do any harm, except for the runs taking longer :)

arstrube:
Oh yes, he wanted me to fix the error first... not that it is a real fix, but at least it stops the compiler from its silly barking :D

PR #871, opened 2016-02-16
Fix isnan() / isinf() GCC bug problem

arstrube:


arstrube:
Hi @bukulin, the way to go would be to extend the warning. Is there a way to distinguish 4.9.x from 4.8.x?

You are right about std::isnan(), which is defined as overloaded functions, rather than a (now broken) macro. However, this code must be able to compile without using the standard C++ library

offa:
> s there a way to distinguish 4.9.x from 4.8.x?

Yes: 

- **`__GNUC__`** Major Version (eg. 4.x, 5.x)
- **`___GNUC_MINOR__`** Minor Version (eg. x.8, x.9, x.3)

arstrube:
Then we can change the #if to suppress the warning as of Gnu C 4.9.

offa:
You are toooo fast :-P (see my updated comment).

arstrube:
no, I'm too lazay. didn't feel like actually coding it ;-)

offa:
I'm not sure, but isn't there a macro like `GNUC_MIN(<major>, <minor>)`?

arstrube:
That would be nice :-)

offa:
**Got it (features.h):**

```
/* Convenience macros to test the versions of glibc and gcc.
   Use them like this:
   #if __GNUC_PREREQ (2,8)
   ... code requiring gcc 2.8 or later ...
   #endif
   Note - they won't work for gcc1 or glibc1, since the _MINOR macros
   were not defined then.  */
#if defined __GNUC__ && defined __GNUC_MINOR__
# define __GNUC_PREREQ(maj, min) \
	((__GNUC__ << 16) + __GNUC_MINOR__ >= ((maj) << 16) + (min))
#else
# define __GNUC_PREREQ(maj, min) 0
#endif
```

So instead of my example above: `__GNUC_PREREQ(4, 9)` -- handles even the case, that `__GNUC__` isn't defined (= no GCC).

offa:
Keeps working on GCC 5.3:

```c
#if __GNUC_PREREQ(4, 9)
#pragma GCC diagnostic ignored "-Wfloat-conversion"
#endif
```

But let's wait for @bukulin or at least someone who has 4.9.

offa:
If someone wants to test, i have pushed an updated version.

arstrube:
The warning was introduced in Gcc 4.9. However, there is something off about this behavior -- there should be no warning for isnan().

offa:
The only thing about this, i have seen in the [changelog for 4.9](https://gcc.gnu.org/gcc-4.9/changes.html):

> **C**
> ISO C11 support is now at a similar level of completeness to ISO C99 support: substantially complete modulo bugs, extended identifiers (supported except for corner cases when -fextended-identifiers is used), floating-point issues (mainly but not entirely relating to optional C99 features from Annexes F and G) and the optional Annexes K (Bounds-checking interfaces) and L (Analyzability).

arstrube:
Suppressing the warning is the correct action. The bug in Gcc is that they don't suppress this warning in the first place. Because, the macro that actually works correctly, always contains a call that _would_ truncate a double to float, except that that call is never executed on a double, only ever on a float.

arstrube:
I have 4.9.3 at work. If you explicitly set -Werror -Wfloat-conversion, and nothing happens, then the test is passed. On versions that don't have the warning yet, like 4.8.x, the compiler barks about an unknwon commandline switch. Later versions all have the same problem. They would have to put the pragma right inside the macro in order to fix it. Perhaps we should push the previous pragma state and pop it afterwards?

bukulin:
Sorry for the extremely late reply. It works very well.

arstrube:
Actually, I'd suggest we go back to the most compatible option:
```c++
#if defined(__GNUC__) && ( __GNUC__ >= 5 || ( __GNUC__ == 4 && GNUC_MINOR__ > 8 ) )
...
```
Because, the simplification is resulting in unnecessary complications ;-)

Issue #872, opened 2016-02-16
Linux isnan() / isinf() GCC bug problem


offa:
PR #871 fixes the issue for MinGW only, GCC on linux need a fix too. I've made branch to address this; but some builds are failing. 

**Branch:** https://github.com/offa/cpputest/tree/gcc_fix

Btw. maybe the build tool is the better place to add this flag.

offa:
After a test for GCC Major >= 5 all builds are ok now; PR follows.

offa:
Closed via #873; tested on latest GCC and Clang.

PR #873, opened 2016-02-16
Gcc fix for #872

offa:
See #872.

Also tested with current GCC / Clang on linux.

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -235,7 +235,7 @@ void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
 /* MinGw GCC 5.3.0 isnan / isinf macros are not working properly, causing a conversion
  * warning / error
  */
-#ifdef __MINGW32__
+#if defined(__MINGW32__) || defined(__GNUC__)
 #pragma GCC diagnostic ignored "-Wfloat-conversion"
 #endif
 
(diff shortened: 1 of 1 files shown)


src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -235,7 +235,7 @@ void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
 /* MinGw GCC 5.3.0 isnan / isinf macros are not working properly, causing a conversion
  * warning / error
  */
-#ifdef __MINGW32__
+#if defined(__MINGW32__) || (defined(__GNUC__) && __GNUC__ >= 5)
```
basvodde: Goes the GNU #define not also cover MINGW?

arstrube: I think if you look at the Gnu C version, then you can (and actually should) leave out the check for mingw. I was wondering about this; it stands to reason that Gnu C (mis-)behavior would be the same across platforms.

Andne: Based on the failure in PR #874, it appears that we definitely want to only check for the GCC version.  MinGW/GCC 4 is breaking there because it doesn't know the #pragma that this wraps.

arstrube: Agreed.

offa:
MinGW check removed, let's wait for CI.

PR #874, opened 2016-02-17
Run both MinGW 4.8.5 and MinGW 5.3.0 on appveyor

Andne:
I've rebased my previous pull request so that it shouldn't fail now.  This version runs both MinGW 4.8.5 and MinGW 5.3.0 on appveyor.  Hopefully this will help us see future failures like `isnan()` one that showed up recently.  Unfortunately, each MinGW configuration seems to take 10-15 minutes to run, so it does add 20-30 minutes to the build time.  Overall it looks like it's 45-70 minutes for appveyor to run, just based on the last couple builds I've seen on this branch.

Also, I've fixed the output from the unzip during the MinGW install command, it gets dumped to null for now.  I'm not the happiest about it since it also dumps the normal install command output, but until chocolatey gets updated to handle this itself, that's how it's going to be.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,13 +8,21 @@ environment:
   - PlatformToolset: Cygwin
   - PlatformToolset: MinGW
     Platform: Win32
+    Version: 4.8.5
   - PlatformToolset: MinGW
     Platform: x64
+    Version: 4.8.5
+  - PlatformToolset: MinGW
+    Platform: Win32
+    Version: 5.3.0
+  - PlatformToolset: MinGW
+    Platform: x64
+    Version: 5.3.0
   - PlatformToolset: v90
   - PlatformToolset: v100
 
 install:
-- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw $( if ($env:Platform -eq 'Win32') { Write-Output -forcex86 } ) }
+- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw --version $env:Version $( if ($env:Platform -eq 'Win32') { Write-Output -forcex86 } ) }
 - ps: if ($env:PlatformToolset -eq 'Cygwin') { (New-Object System.Net.WebClient).DownloadFile('https://cygwin.com/setup-x86.exe', "${env:TEMP}\cygwin-setup.exe") ; . "${env:TEMP}\cygwin-setup.exe" --quiet-mode --packages autoconf automake make gcc-core libtool gcc-g++ }
 
 build_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ environment:
   - PlatformToolset: v100
 
 install:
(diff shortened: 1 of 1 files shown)


basvodde:
Failed :(

Andne:
@arstrube It looks like your fix for MinGW 5 broke MinGW 4

arstrube:
@Andne we need @offa's fix for my fix first, PR #873. It would be nice to remove the now unnecessary condition with MinGW first, though.

basvodde:
How to restart the appveyor?

arstrube:
Log in, select the build, click "Re-Build Commit"

arstrube:
I can do it for my fork; presumably, for cpputest/cpputest you can do it :-). You should be able to have it log you  in via Github.

arstrube:
@basvodde still failed?

arstrube:
I guess @Andne needs to rebase this for it to pass.

appveyor.yml
```diff
@@ -8,13 +8,21 @@ environment:
...
+  - PlatformToolset: MinGW
+    Platform: x64
+    Version: 5.3.0
   - PlatformToolset: v90
   - PlatformToolset: v100
 
 install:
-- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw $( if ($env:Platform -eq 'Win32') { Write-Output -forcex86 } ) }
+- ps: if ($env:PlatformToolset -eq 'MinGW') { choco install mingw --version $env:Version $( if ($env:Platform -eq 'Win32') { Write-Output -forcex86 } ) | Out-Null }
 - ps: if ($env:PlatformToolset -eq 'Cygwin') { (New-Object System.Net.WebClient).DownloadFile('https://cygwin.com/setup-x86.exe', "${env:TEMP}\cygwin-setup.exe") ; . "${env:TEMP}\cygwin-setup.exe" --quiet-mode --packages autoconf automake make gcc-core libtool gcc-g++ }
```
arstrube: @Andne I just found out that Cygwin is now pre-installed. You can remove the  - ps: line for Cygwin.

Andne:
Ok, rebased, it should succeed now.

appveyor.yml
Andne: I think this line acts like a no-op, I knew cygwin was installed, at one point was trying to run x64 cygwin instead of x86, but that never worked right.

Andne:
Failed to install MinGW 5.3.0 on 32bit.  No idea what the failure was since the output was all redirected to null ;)

arstrube:
Just re-run it. Appveyor does that sometimes. Usually knowing the "cause" doesn't help much (e.g. couldn't manage to download).

Andne:
I think appveyor is having network issues, the build under my account for that branch failed on a checkout step for a different configuration because it couldn't connect to github.com

arstrube:
BTW I applied your patch to my local appveyor.yml and it literally saved me hours of waiting, scrolling, and waiting again for the "A script is running slow" dialog so I could kill it and get on with life :-). Thanks!

Andne:
Re-ran my branch and it was successful.  The changes work, just needs to be re-run.

PR #876, opened 2016-02-17
Add signalling to ieee754 plugin

arstrube:
Some of this is debatable.

Also, the test is failing on Appveyor. Apparently a signal is not raised; the test just fails normally. I don't know about MinGW, but I am really puzzled by the Cygwin failure, because I used Cygwin in preparation of this PR a lot, and the same test is successful on my machine.

@Andne any idea what might be wrong?


arstrube:
Also surprised to see that MinGW Gcc doesn't seem to have `feenableexcept()` /  `fedisableexcept()`. I guess it's pointless to ask why - other people seem to take the approach I took for Os X, so it looks like I will need to expand the `#ifdef` to include `__MINGW32__`

arstrube:
The clang failure is strange. It complains about `CPPUTEST_FENV_LACKS_SIGNALLING` not being defined. But the only case in which this can be so is, when `CPPUTEST_HAVE_FENV` is not defined. If `CPPUTEST_HAVE_FENV` were not defined, however, the entire block -- including the line that clang complains about -- should not have compiled.

basvodde:
The code, as it looks now, is not integratable. I won't integrate copy-pasted-ugly code.

arstrube:
I suppose you are referring to the OSX stuff... So, what do you suggest instead? Options would be:

1. Omit functionality for OSX and use stubs, same as for MinGW and Watcom
2. Try to un-copy-paste-un-uglify the code. That wouldn't change the fact that it isn't my code, and you still might not want it in there anyway
3. Suggestions?

As it stands, I am more concerned about Cygwin / Gcc not working properly as it should. The test fails, instead of signalling.

arstrube:
What is worse, I wasn't able to reproduce it.

basvodde:
3. Use the code to rewrite it in a sane and clearer way

arstrube:
It's low-level bit-masking stuff. There's nothing "sane" about it. :P

basvodde:
#if 0 a piece of commented out code :)

arstrube:
I also only works with Osx / Clang. The same code will, for example, not work with Mingw / Gcc because it's low-level implementation stuff that's not guaranteed to be the same.

arstrube:
the #if 0 is purely my fault, because it's a static function that _might_ be useful and hard to find again, but the compiler will complain if it isn't used :-((

arstrube:
of course, I could

- find some way to use it even though it isn't needed
- provide a wrapper for it (that nobody will want to use)
- or finally, throw it out entirely :D 

(the #if 0 one)

arstrube:
A more serious problem I currently see with this code is that it is not tested on either Travis or Appveyor.

arstrube:
@basvodde will de-uglify & rebase later tonight.

This failed test in Cygwin is really weird. I have tried various versions of Gcc, different configurations via ../configure, even let the check_all target run until it fails because of no Gtest -- I absolutely cannot get the test to fail. It crashes in a separate process as it should. Yet, on Appveyor, there is a failure:
```
../include/CppUTest/TestTestingFixture.h:115: actual [
Generic:1: error: Failure in TEST(Generic, Generic)
 Failed in separate process

.
Errors (1 failures, 1 tests, 0 ran, 0 checks, 0 ignored, 0 filtered out, 22 ms)

]
 did not contain  [Failed in separate process - killed by signal 8]
```

examples/AllTests/FEDemoTest.cpp
```diff
@@ -54,10 +54,19 @@ TEST_GROUP(FE_Demo) {
...
+#define CRASHING_TEST IGNORE_TEST
 #endif
 
+CRASHING_TEST(FE_Demo, should_crash_when__feenableexcept_was_called) {
+    f = 1.0f;
+    IEEE754ExceptionsPlugin::enableSignal();
+    CHECK((f /= 0.0f) >= std::numeric_limits<float>::infinity() );
+    IEEE754ExceptionsPlugin::disableSignal();
+}
+
```
basvodde: If we have the ability to run an IGNORED_TEST than all this can be deleted, right? Adding CRASHING_TEST just made it worst ;)

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -34,6 +34,53 @@ extern "C" {
     #include <fenv.h>
 }
 
+#if CPPUTEST_FENV_LACKS_SIGNALLING
+
+/*
+ * Missing implementations for Apple, MinGW and Watcom C.
```
basvodde: Please :)

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -34,6 +34,53 @@ extern "C" {
...
+{
+    fenv_t fenv;
+    unsigned short old_excepts;
+    unsigned short new_excepts = (unsigned short)excepts & FE_ALL_EXCEPT;
+
+    if (fegetenv (&fenv)) return -1;
+
+    old_excepts = fenv.CONTROL_WORD & FE_ALL_EXCEPT;
+    /* Mingw GCC promotes ~<unsigned short> to <int> and then complains about
+     * loss of information when casting from <int> to <unsigned short>. Yikes.
```
basvodde: Huh?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -34,6 +34,53 @@ extern "C" {
...
+#define CONTROL_WORD __control_word
+#elif defined(__WATCOMC__)
+#define CONTROL_WORD control_word
+#endif
+
+int feenableexcept (int excepts)
+{
+    fenv_t fenv;
+    unsigned short old_excepts;
+    unsigned short new_excepts = (unsigned short)excepts & FE_ALL_EXCEPT;
```
basvodde: Can we move to C++ style ?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -34,6 +34,53 @@ extern "C" {
...
+    fenv.CONTROL_WORD &= (unsigned short)~new_excepts;
+
+    return (fesetenv(&fenv) ? -1 : old_excepts);
+}
+
+int fedisableexcept (int excepts)
+{
+    fenv_t fenv;
+    unsigned short old_excepts;
+    unsigned short new_excepts = (unsigned short)excepts & FE_ALL_EXCEPT;
```
basvodde: Likewise

examples/AllTests/FEDemoTest.cpp
arstrube: In this instance, absolutely yes.

There are other cases (like #ifdef HAVE_FORK) where I would still want the test run if the OS supports forking, for instance.

arstrube: The intention here is the the user can opt to run this test to see what happens, but we don't want it run otherwise. So it fits the bill perfectly.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Exactly. You want to submit the bug? I find their system too cumbersome. Minimal example:
```
#pragma GCC diagnostic error "-Wconversion"
int main(int, char**) {
    unsigned short u = 0x000f;
    u = ~u;
    return (int) u;
}

/* Compiler output (MinGW 4.8.1, 4.9.x, 5.3.0):

OneTimeThrowAwayTests.cpp: In function 'int main(int, char**)':
OneTimeThrowAwayTests.cpp:5:7: error: conversion to 'short unsigned int' from 'int' may alter its value [-Werror=conversion]
     u = ~u;
       ^
*/
```
e.g. no cast should be necessary in the line below.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: But this is really a conversion from int to unsigned short, so the Werror will be justified

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Not sure what you mean?

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: Actually, to make it perfectly clear, the comment would need a forth line:

1. The compiler seems to promote ~unsigned short to int (why? beats me)
2. It then complains about _implicitly_ casting that int back to unsigned short
3. The explicit cast will make it shut up.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
arstrube: @basvodde maybe I am blinded by reading too much C code, but I am really at loss what should make this line "C++ style"?

arstrube:
Now that MinGW passes, I find the Cygwin failure or the more puzzling. If they both failed, I would have assumed that on Appveyor for some reason, floating point signals were disabled. But, apparently, they do work.

src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
basvodde: Not declare all variables at the beginning :)

arstrube: Oh. Okay. Must indeed have dealt with too much C code recently. Even had a C compiler bark at me for not doing so :-)

arstrube:
Okay, guys. I am giving up on the Cygwin / Appveyor thing.

I suppose that means this PR is impossible to integrate right now. I feel it can and must pass the Cygwin test (do #ifdef'ing it away is NOT an option), but I have no idea how to make it happen.

basvodde:
I'll leave it open for your inspiration :)

arstrube:
May inspiration kiss me soon :)

arstrube:
I think this isn't needed anymore. Setting a watch based on a function that inspects a flag (#906) is more useful in most cases.

Issue #877, opened 2016-02-17
CMake-generated Xcode builds fail


charlesnicholson:
These repro steps use the Xcode command-line tools, but I've confirmed that it happens in the GUI. This also happens if CppUTest is pulled in via a CMake ExternalProject_Add() command from another CMake project, which is how I found it.

```
git clone https://github.com/cpputest/cpputest.git
cd cpputest
cmake -G Xcode
xcodebuild
```

CppUTest, CppUTestExt, CppUTestTests all appear to build correctly, but the build fails during test discovery:

```
PhaseScriptExecution CMake\ PostBuild\ Rules tests/CppUTest.build/Debug/CppUTestTests.build/Script-44B36442352E44DA89F0468A.sh
    cd /Users/cnicholson/src/cpputest
    /bin/sh -c /Users/cnicholson/src/cpputest/tests/CppUTest.build/Debug/CppUTestTests.build/Script-44B36442352E44DA89F0468A.sh
echo "Discovering Tests in CppUTestTests"
Discovering Tests in CppUTestTests
cd /Users/cnicholson/src/cpputest/tests && /usr/local/Cellar/cmake/3.4.3/bin/cmake -DTESTS_DETAILED:BOOL=OFF -DEXECUTABLE=CppUTestTests -P /Users/cnicholson/src/cpputest/cmake/Scripts/CppUTestBuildTimeDiscoverTests.cmake
CMake Error at /Users/cnicholson/src/cpputest/cmake/Scripts/CppUTestBuildTimeDiscoverTests.cmake:42 (message):
  Executable "/Users/cnicholson/src/cpputest/tests/CppUTestTests -lg" failed
  with output:

  No such file or directory

  Please check that the excutable was added.
Call Stack (most recent call first):
  /Users/cnicholson/src/cpputest/cmake/Scripts/CppUTestBuildTimeDiscoverTests.cmake:75 (buildtime_discover_tests)
  


make: *** [CppUTestTests_buildpart_0] Error 1

** BUILD FAILED **


The following build commands failed:
	PhaseScriptExecution CMake\ PostBuild\ Rules tests/CppUTest.build/Debug/CppUTestTests.build/Script-44B36442352E44DA89F0468A.sh
(1 failure)
```

The issue appears to be that CMake is looking for
`.../cpputest/tests/CppUTestTests`
but the executable actually lives at
`.../cpputest/tests/[Debug|Release]/CppUTestTests`

[...]

PR #878, opened 2016-02-17
Fix cmake xcode

charlesnicholson:
By default, Xcode (and VS?) build artifacts to a path that includes the configuration name. In Xcode's case, this was "Debug". CMake was expecting its self-test executables CppUTestTests and CppUTestExtTests to be in specific paths that did not include the configuration paths.

This PR forces Xcode to put the test executables in the same location that the other CMake generators (make / ninja) put them. This is also where the non-cmake autoconf builds put them.

Fixes https://github.com/cpputest/cpputest/issues/877

cc @basvodde


charlesnicholson:
@basvodde ping on this? not urgent, of course, just curious what your take is.

basvodde:
Merged, thanks. Sorry for delay

PR #880, opened 2016-02-18
Dormant test

PaulBussmann:
Currently the CppUTests are all either "opt-out" (can be filtered to be not run with command line) or they are disabled ("IGNORE_TEST").  

I am looking for a way to have CppUTests "opt-in", means that they are included in the testing executable, but only run if I explicitely chose them.

So I created the macro DORMANT_TEST that declares an "opt-in" test dormant until it is awaked using strict name filtering using e.g. "-sg <group-name> -sn <test-name>" . Like IGNORE_TEST it is not executed by default, so I DORMANT_TEST is an apoption of IGNORE_TEST, with key change of DormantUtestShell::shouldRun and DormantUtestShell::willRun .
**--> Please pull my changes to cpputest master.**

About IGNORE_TEST:
First I thought this could be done using IGNORE_TEST, but it looks like there is no way to run an IGNORE_TEST compiled into a testing executable even not using command line option "IGNORE_TEST" (only sets up filtering for the named test, but cannot execute it -> why sets it up filtering though?) or a custom derived runner class. 

The current behavior of "IGNORE_TEST" is useless to me, I can easily surround a test with "#if 0 ... #endif" if don't like it to be included in the testing executable.
IMHO it would be better to be used for tests that are executed, but not counted as failures if they fail (if all TEST(...)s pass, the IGNORE_TEST()s won't make the test run FAIL, but I can see from the verbose logs if the test was successful or not).

### DIFF
--- a/include/CppUTest/Utest.h
+++ b/include/CppUTest/Utest.h
@@ -95,7 +95,7 @@ public:
     virtual UtestShell *getNext() const;
     virtual int countTests();
 
-    bool shouldRun(const TestFilter* groupFilters, const TestFilter* nameFilters) const;
+	virtual bool shouldRun(const TestFilter* groupFilters, const TestFilter* nameFilters);
     const SimpleString getName() const;
     const SimpleString getGroup() const;
     virtual SimpleString getFormattedName() const;
@@ -152,9 +152,10 @@ protected:
 
     virtual SimpleString getMacroName() const;
     TestResult *getTestResult();
+	bool match(const char* target, const TestFilter* filters) const;
+	const char *group_;
+	const char *name_;
 private:
-    const char *group_;
-    const char *name_;
     const char *file_;
     int lineNumber_;
     UtestShell *next_;
@@ -163,7 +164,6 @@ private:
 
     void setTestResult(TestResult* result);
     void setCurrentTest(UtestShell* test);
-    bool match(const char* target, const TestFilter* filters) const;
 
     static UtestShell* currentTest_;
     static TestResult* testResult_;
@@ -229,6 +229,28 @@ private:
     IgnoredUtestShell(const IgnoredUtestShell&);
     IgnoredUtestShell& operator=(const IgnoredUtestShell&);
 
+};
+
+//////////////////// DormantTest
+
+class DormantUtestShell : public IgnoredUtestShell
+{
+public:
+	DormantUtestShell();
+	virtual ~DormantUtestShell();
+	explicit DormantUtestShell(const char* groupName, const char* testName,
+		const char* fileName, int lineNumber);
+	virtual bool willRun() const _override;
+	bool match(const char* target, const TestFilter* filters, const bool strict) const;
+	virtual bool shouldRun(const TestFilter* groupFilters, const TestFilter* nameFilters) _override;
+protected:  virtual SimpleString getMacroName() const _override;
+			virtual void runOneTest(TestPlugin* plugin, TestResult& result) _override;
+
+private:
+
+	DormantUtestShell(const DormantUtestShell&);
+	DormantUtestShell& operator=(const DormantUtestShell&);
+	bool willRun_;
 };
 
 //////////////////// TestInstaller
--- a/include/CppUTest/TestFilter.h
+++ b/include/CppUTest/TestFilter.h
@@ -42,6 +42,7 @@ public:
     TestFilter* getNext() const;
 
     bool match(const SimpleString& name) const;
+	bool isStrictMatching() const;
 
     void strictMatching();
     void invertMatching();
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -76,12 +76,26 @@
   \
   class IGNORE##testGroup##_##testName##_Test : public TEST_GROUP_##CppUTestGroup##testGroup \
 { public: IGNORE##testGroup##_##testName##_Test () : TEST_GROUP_##CppUTestGroup##testGroup () {} \
-  public: void testBodyThatNeverRuns (); }; \
+  public: void testBodyThatNeverRuns(); }; \
   class IGNORE##testGroup##_##testName##_TestShell : public IgnoredUtestShell { \
       virtual Utest* createTest() _override { return new IGNORE##testGroup##_##testName##_Test; } \
   } IGNORE##testGroup##_##testName##_TestShell_instance; \
    static TestInstaller TEST_##testGroup##testName##_Installer(IGNORE##testGroup##_##testName##_TestShell_instance, #testGroup, #testName, __FILE__,__LINE__); \
-    void IGNORE##testGroup##_##testName##_Test::testBodyThatNeverRuns ()
+    void IGNORE##testGroup##_##testName##_Test::testBodyThatNeverRuns()
+
+#define DORMANT_TEST(testGroup, testName)\
+  /* External declarations for strict compilers */ \
+  class DORMANT_##testGroup##_##testName##_TestShell; \
+  extern DORMANT_##testGroup##_##testName##_TestShell DORMANT_##testGroup##_##testName##_TestShell_instance; \
+  \
+  class DORMANT_##testGroup##_##testName##_Test : public TEST_GROUP_##CppUTestGroup##testGroup \
+{ public: DORMANT_##testGroup##_##testName##_Test () : TEST_GROUP_##CppUTestGroup##testGroup () {} \
+  public: void testBody(); }; \
+  class DORMANT_##testGroup##_##testName##_TestShell : public DormantUtestShell { \
+      virtual Utest* createTest() _override { return new DORMANT_##testGroup##_##testName##_Test; } \
+  } DORMANT_##testGroup##_##testName##_TestShell_instance; \
+   static TestInstaller TEST_##testGroup##testName##_Installer(DORMANT_##testGroup##_##testName##_TestShell_instance, #testGroup, #testName, __FILE__,__LINE__); \
+	void DORMANT_##testGroup##_##testName##_Test::testBody()
 
 #define IMPORT_TEST_GROUP(testGroup) \
   extern int externTestGroup##testGroup;\
--- a/src/CppUTest/TestFilter.cpp
+++ b/src/CppUTest/TestFilter.cpp
@@ -63,6 +63,11 @@ void TestFilter::invertMatching()
     invertMatching_ = true;
 }
 
+bool TestFilter::isStrictMatching() const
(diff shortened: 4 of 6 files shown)


arstrube:
I like this. I have a test in the demo executable (examples) that demonstrates an (intentional) crash upon floating point error. Right now, a user has to comment in a #define in order to run it, and of course it mustn't run by default. An opt-in test would be the most elegant way to handle this situation, also.

arstrube:
IGNORE_TEST has its own merits though. One of the being that the test is still run past the compiler, which is not the case if you #if 0 them.

include/CppUTest/Utest.h
```diff
@@ -152,9 +152,10 @@ class UtestShell
 
     virtual SimpleString getMacroName() const;
     TestResult *getTestResult();
+	bool match(const char* target, const TestFilter* filters) const;
+	const char *group_;
+	const char *name_;
```
basvodde: Moved to public ?

basvodde:
Can we add this behavior to IGNORE_TEST instead ? It seems useful :) At least, with the name explicitly.

arstrube:
I would prefer the name OPT_IN_TEST to DORMANT_TEST

@basvodde - not quite sure what you mean with "add this behavior to IGNORE_TEST" ?

basvodde:
@martiert You want to have a look at this?

basvodde:
The functionality of running an ignored test if you specifically call it via -sn.

arstrube:
Ah. Sounds good to me.

PaulBussmann:
In fact a IGNORE_TEST should IMHO be run if called with -sn && -sg . Which would be the perfect solution for me! 

This is independent of my proposal -> are you awaiting more from me? I’d like to trash it and awaiting your fix to being able to call IGNORE_TEST with -sn && -sg.

 

 

Von: Bas Vodde [mailto:notifications@github.com] 
Gesendet: Donnerstag, 18. Februar 2016 09:40
An: cpputest/cpputest
Cc: PaulBussmann
Betreff: Re: [cpputest] Dormant test (#880)

 

The functionality of running an ignored test if you specifically call it via -sn.

basvodde:
Why is it independent? If IGNORE_TEST can be called with -sn, then DORMANT_TEST is not needed anymore, right?

PaulBussmann:
Yes, if I would be able to call IGNORE_TEST with -sn && -sg (only -sn is insufficient IMHO), then I have my OPT-IN and don’t need DORMANT_TEST anymore.

 

 

 

Von: Bas Vodde [mailto:notifications@github.com] 
Gesendet: Donnerstag, 18. Februar 2016 09:54
An: cpputest/cpputest
Cc: PaulBussmann
Betreff: Re: [cpputest] Dormant test (#880)

 

Why is it independent? If IGNORE_TEST can be called with -sn, then DORMANT_TEST is not needed anymore, right?

basvodde:
Why is -sn insufficient?

arstrube:
Tests with exactly the same name can live in different groups, no?

PaulBussmann:
Example (not really good, but demonstrates my needs):

 

IGNORE_TEST(PPC, ValgrindReportsNoPossibleLeaks)

IGNORE_TEST(PPC, ValgrindReportsNoDefinitiveLeaks)

IGNORE_TEST(PPC, Valgrind)

TEST(x86, ValgrindReportsNoPossibleLeaks)

TEST(x86, ValgrindReportsNoDefinitiveLeaks)

TEST(x86, Valgrind)

TEST(x64, ValgrindReportsNoPossibleLeaks)

TEST(x64, ValgrindReportsNoDefinitiveLeaks)

TEST(x64, Valgrind)

 

I want to run all tests that are named exactly “Valgrind”. I want to filter TESTs, to speed up my test run). 

 

The problem is that “-sn”, “-sg” are currently used only for filtering, not for activation, therefore the user’s intention must be figured out (-> does the user want to filter or activate tests?).

 

Giving only “-sn” without “-sg” should not run the “IGNORE_TEST”, because in the example above I don’t want to activate IGNORE_TESTs from PPC which are currently having issues, I want to filter the others to speed up my tests. 

 

Therefore I suggest to only activate IGNORE_TESTs with -sn && -sg. Giving both there is no doubt that activation instead of filtering was my intention.

 

 

Von: arstrube [mailto:notifications@github.com] 
Gesendet: Donnerstag, 18. Februar 2016 10:04
An: cpputest/cpputest
Cc: PaulBussmann
Betreff: Re: [cpputest] Dormant test (#880)

 

Tests with exactly the same name can live in different groups, no?

basvodde:
The problem is that I can assume people explicitly stating a test, they want to run it. But for a group, I don't think we can assume that.

This makes me wonder whether it is not better to add a parameter for "run ignored tests" ?

arstrube:
Why not use the existing syntax:
```
$ ./CppUTestExtTests.exe -v "IGNORE_TEST(FE__with_Plugin, should_crash___when__feenableexcept_was_called)"
```
This test will run with "IGNORE_TEST" or "TEST" on the command line. Currently, it will just be ignored. The new feature would actually run it.

arstrube:
It is the same as -sg -sn but more explicit.

PaulBussmann:
> $ ./CppUTestExtTests.exe -v "IGNORE_TEST(FE__with_Plugin, should_crash___when__feenableexcept_was_called)"
> This test will run with "IGNORE_TEST" or "TEST" on the command line. Currently, it will just be ignored. The new feature would actually run it.
 
Perfect! It was in fact the first thing I tried when I saw this this command line option to get the ignored test running.

basvodde:
How about just adding a command line option to run ignored tests? That way it can be combined with all the other filtering and selection methods ?

cuitoldfish:
IGNORE_TEST is supposed to be ignored, why not let user to explicitly rename IGNORE_TEST to TEST to run them?

PaulBussmann:
> IGNORE_TEST is supposed to be ignored, why not let user to explicitly rename IGNORE_TEST to TEST to run them?

@Liu Nick: thanks for your answer and sorry for not previously explaining my motivation:

 

> I am looking for a way to have CppUTests "opt-in", means that they are included in the testing executable, but only run if I explicitely chose them.

This is because building the testing executable is part of a build process taking some hours overall. After this, tests are executed in my CppUTest-base product test binary which also contains functional tests and takes up to hours of complete execution. 

After the build is done, it is quite inconvenient to re-build the testing executable for a released product with changed sources. Unfortunately this isn’t easy to handle. Although I am about to improve the build infrastructure to accomplish this, right now, it is simply not possible.

 

It would be really, really comfortable, to not have to re-compile the testing executable which would be possible if CppUTest provides “OPT-IN” tests.

 

> IGNORE_TEST is supposed to be ignored

I agree that IGNORE_TEST is not an OPT-IN test (although it could be and I’d be very happy having this! : ) ), so I suggested to have an additional macro on initial pull request. 

Finally after this discussion, I think it would be very, very smart to have IGNORE_TEST as OPT-IN test, so CppUTest accepts that that I change my mind if I want to run a test or not after compilation … : )

arstrube:
@PaulBussmann I would like to have this feature, too.

cuitoldfish:
Hi,
If the reason is that re-compile take too long time, then I can not concur with this. 
Try to think, if compilation takes long time which might as long as hours, then how do you get real time feed back from unit test? 
One distinct character of good unit test is the short build-and-execute cycle time.
I raise this because i'm also suffer from long build-and execute cycle time in my unit test environment, and I'm finding methods to try to accelerate it.

PaulBussmann:
For our unit tests, we get the response in real-time within seconds and this is not the motivation for my question. 

My use case is that I have another CppUTest based testing executable as artefact from product release building process. This process takes hours and as already stated I am currently not able to easily change sources and recompile just in order to run a dormant test. Changing and recompile from a sealed build is furthermore not good in respect of source code versioning from QA…

 

Therefor I agree with you that for development unit tests from within Visual Studio it’s not needed. 

But in the test harness where you don’t even have the compiler matching the target’s OS, etc. it is simply not possible to recompile. Therefor I just kindly ask you to add this feature. 

 

If my programming as appreciated, please let me know.

arstrube:
I would like to have this, both for running automated tests as part of a build process, and for ad-hoc testing. Also, this would make things much better for CppUTests own testing. 

__Other Use Cases__

1. Some tests can be successful only under certain conditions, or on specific platforms and are therefore implemented as IGNORE_TEST, possibly with #ifdef. We could now do away with the #ifdef and run the tests as opt-in tests, where appropriate.

2. Some tests inCppUTest examples are to demonstrate certain failures or even a (intended) crash. A user ought to be able to run these easily as opt-in tests, but they should be ignored during regular testing.

Personally, I have always disliked the fact that I had to change / recompile the code in order to run an ignored test (something I can just as easily do with #if, which can be commented in / out easily, whereas to change TEST to IGNORE_TEST and vice versa, I have to edit the actual code).

The only time when IGNORE_TEST in its current form is really useful is when I want to run a test by the compiler, but it can't possibly run (because a feature is not implemented) and probably never will. IGNORE_TEST tends to be, well, ignored forever..... ;-)

For me, the current implementation of IGNORE_TEST is useful only if I intend to ignore a test "permanently" (like: I know I should test this, but I can't and I won't).

arstrube:
@basvodde I think a command line option, which can be combined with any filtering, would be nice.

basvodde:
Ok, so when we move to a command line option, then I won't merge the DORMANT tests as a command line option that runs ignored tests will do the same. However, a lot of the code in this PR can be re-used. Can we change it to that?

basvodde:
Now the -ri option is implemented. Does that make this pull request obsolute?

PaulBussmann:
Yes, thank you very much! :))) 

So it'll be part of V3.7.3, right?
Btw. how can I prepare my code for transitioning to V3.7.3? 
I noticed round about 100 compiler errors in my test program due to parameter changes of TestRunner and CHECK... macros since V3.7.2.

arstrube:
Wow. Did we change all that :-/ ? Most changes I recall shouldn't have broken any tests. Would you care to elaborate (perhaps an other issue, or Google Groups)?

Is it possible that perhaps you are using mismatched headers?

PaulBussmann:
Thank you very much for your support! 

I'll summarize the conflicts of my code with V3.7.2 API against current sources on "master" and open another issue if needed. IMHO this pull request can be closed.

cuitoldfish:
hi, I also encountered the compatible issue when I upgrade my CppUTest Framework from an older version to a newer one, including compile error, crash error.

basvodde:
Interesting. Especially crashes. Would love to know where they come from.

Issue #879, opened 2016-02-18
Mock failures with repeat count > 1 do not cause failure


hbock:
Hi,

We noticed in internal testing that with the MockSupportPlugin, any mock failure (e.g., mock().actualCall() without an expectOneCall()) that occurs in multiple runs in the same process - i.e., -r2 or higher, is only reported on the first run:

```
hbock@whiskey-dev ~/development/scratch/cpputest-mock-repeat-test $ ./main -r4
Test run 1 of 4

testgroup.cpp:8: error: Failure in TEST(MyGroup, First)
	Mock Failure: Unexpected call to function: Foo
	EXPECTED calls that did NOT happen:
		<none>
	ACTUAL calls that did happen (in call order):
		<none>

.
Errors (1 failures, 1 tests, 1 ran, 0 checks, 0 ignored, 0 filtered out, 0 ms)

Test run 2 of 4
.
OK (1 tests, 1 ran, 0 checks, 0 ignored, 0 filtered out, 0 ms)

Test run 3 of 4
.
OK (1 tests, 1 ran, 0 checks, 0 ignored, 0 filtered out, 0 ms)

...
```
It seems like the failure should be reported on all test runs.  Running through the debugger, I notice that the failure is detected, but we never get to actually terminate the test or report the failure because UTestShell::hasFailed() returns true.  See the logic in MockFailure.cpp:

https://github.com/cpputest/cpputest/blob/2e32dd0aed2cc4429455e70f148a1865842ba0a6/src/CppUTestExt/MockFailure.cpp#L59-L60

If I remove the conditional on line 59, the failure is reported as expected.  It doesn't seem to affect the outcome of any CppUTestExt unit tests.   I'm not sure if the check is not valid, or if UTestShell's failure state needs to be reset between repeats, or something else.  

I'm happy to fix it if I can get some guidance on what the correct operation of MockFailure is.  :)

hbock:
Attached is a small repeatable test group that exhibits the issue.  Run on the command line with -r2.

[cpputest-mock-repeat-test.zip](https://github.com/cpputest/cpputest/files/135525/cpputest-mock-repeat-test.zip)

cuitoldfish:
Hi hbock,
I think the overall guideline of TDD is that try best to match one test failure to only one test, and only fix one test failure at one time. 
So CppUTest only report the first failure and also only report once to make you focus on fixing that sole failure, then next.   I have met several similar case in other usage. and, 

[...]

PR #881, opened 2016-02-19
Fix for Issue #879

arstrube:
This fixes the bug that issue #879 describes

### DIFF
--- a/tests/CppUTestExt/MockSupportTest.cpp
+++ b/tests/CppUTestExt/MockSupportTest.cpp
@@ -199,7 +199,7 @@ static void crashMethod()
     cpputestHasCrashed = true;
 }
 
-static void crashOnFailureTestFunction_(void)
+static void unexpectedCallTestFunction_(void)
 {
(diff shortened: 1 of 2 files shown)


tests/CppUTestExt/MockSupportTest.cpp
```diff
@@ -233,3 +233,19 @@ TEST(MockSupportTestWithFixture, ShouldNotCrashOnFailureAfterCrashMethodWasReset
...
+    fixture.setTestFunction(unexpectedTestFunction_);
+
+    fixture.runAllTests();
+    fixture.assertPrintContains("Unexpected call to function: unexpected");
+
+    delete fixture.result_;
+    delete fixture.output_;
+    fixture.output_ = new StringBufferTestOutput();
+    fixture.result_ = new TestResult(*fixture.output_);
+
```
arstrube: The reason I am not using a "fresh" fixture is that I want the registry to remain the same, as is the case when a real repeat is done.

PR #885, opened 2016-02-22
DOUBLES_EQUAL - debug output precision

mmidor:
Precision of double number comparison diagnostic message
is adjused dynamically to always show the difference point.

### DIFF
--- a/examples/nbproject/project.xml
+++ b/examples/nbproject/project.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://www.netbeans.org/ns/project/1">
+    <type>org.netbeans.modules.cnd.makeproject</type>
+    <configuration>
+        <data xmlns="http://www.netbeans.org/ns/make-project/1">
+            <name>CppUTest</name>
+            <c-extensions>c</c-extensions>
+            <cpp-extensions>cpp</cpp-extensions>
+            <header-extensions>h</header-extensions>
+            <sourceEncoding>UTF-8</sourceEncoding>
+            <make-dep-projects/>
+            <sourceRootList>
+                <sourceRootElem>src/CppUTest</sourceRootElem>
+                <sourceRootElem>src/CppUTestExt</sourceRootElem>
+                <sourceRootElem>src/Platforms</sourceRootElem>
+                <sourceRootElem>include</sourceRootElem>
+                <sourceRootElem>examples/AllTests</sourceRootElem>
+                <sourceRootElem>../include/CppUTest</sourceRootElem>
+                <sourceRootElem>../include/CppUTestExt</sourceRootElem>
+                <sourceRootElem>../include/Platforms</sourceRootElem>
+                <sourceRootElem>AllTests</sourceRootElem>
+                <sourceRootElem>ApplicationLib</sourceRootElem>
+                <sourceRootElem>../src/CppUTest</sourceRootElem>
+                <sourceRootElem>../src/CppUTestExt</sourceRootElem>
+                <sourceRootElem>../src/Platforms</sourceRootElem>
+                <sourceRootElem>../tests</sourceRootElem>
+            </sourceRootList>
+            <confList>
+                <confElem>
+                    <name>Debug</name>
+                    <type>1</type>
+                </confElem>
+            </confList>
+            <formatting>
+                <project-formatting-style>false</project-formatting-style>
+            </formatting>
+        </data>
+    </configuration>
+</project>
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -189,13 +189,14 @@ DoublesEqualFailure::DoublesEqualFailure(UtestShell* test, const char* fileName,
 : TestFailure(test, fileName, lineNumber)
 {
     message_ = createUserText(text);
-    
+
     bool nums_valid = !((PlatformSpecificIsNan(expected)) || (PlatformSpecificIsInf(expected)) ||
                         (PlatformSpecificIsNan(actual)) || (PlatformSpecificIsInf(actual)));
+    int prec = 3;
     SimpleString sexp = StringFrom(expected, prec);
     SimpleString sact = StringFrom(actual, prec);
-    const int prec_max = sizeof(double) >= 8 ? 17 : 7;
-    int prec = 1;
+    int dsize = sizeof(double);
+    const int prec_max = dsize == 8 ? 17 : 7;
 
     if (nums_valid)
     {
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -189,23 +189,24 @@ DoublesEqualFailure::DoublesEqualFailure(UtestShell* test, const char* fileName,
 : TestFailure(test, fileName, lineNumber)
 {
     message_ = createUserText(text);
+    
     bool nums_valid = !((PlatformSpecificIsNan(expected)) || (PlatformSpecificIsInf(expected)) ||
                         (PlatformSpecificIsNan(actual)) || (PlatformSpecificIsInf(actual)));
-    
-    int prec = 7;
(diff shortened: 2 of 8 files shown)


arstrube:
Hmm. Interesting failures....

arstrube:
upping the initial value `int prec = 3;` makes all tests pass again; but it kind of makes you wonder....

arstrube:
Again, I don't think there's any fault with your code; it's just that `"%.*g"` doesn't always do what you'd expect it to do.

arstrube:
Clang on Linux picks up that sizeof(double) will never be larger than 8. Maybe reduce this to 4.

mmidor:
I agree about the double size (I as thinking about larger one, but it must be declared as long double).
And, as I see, I made a mess by displacing `int prec`

mmidor:
BTW on my PC the cygwin build fails:

/cpputest/src/Platforms/Gcc/UtestPlatform.cpp: In function ‘int IsNanImplementation(double)’:
/cpputest/src/Platforms/Gcc/UtestPlatform.cpp:244:12: error: conversion to ‘float’ from ‘double’ may alter its value [-Werror=float-conversion]
     return isnan(d);
            ^
/cpputest/src/Platforms/Gcc/UtestPlatform.cpp: In function ‘int IsInfImplementation(double)’:
/cpputest/src/Platforms/Gcc/UtestPlatform.cpp:249:12: error: conversion to ‘float’ from ‘double’ may alter its value [-Werror=float-conversion]
     return isinf(d);

- - -

$ g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-cygwin/4.9.2/lto-wrapper.exe
Target: x86_64-pc-cygwin
Configured with: /cygdrive/i/szsz/tmpp/gcc/gcc-4.9.2-3.x86_64/src/gcc-4.9.2/configure --srcdir=/cygdrive/i/szsz/tmpp/gcc/gcc-4.9.2-3.x86_64/src/gcc-4.9.2 
Thread model: posix
gcc version 4.9.2 (GCC)

arstrube:
The isnan() issue is supposed to be fixed? (What happens is that newer versions of GCC exhibit a bug; the correct size function should be selected by a macro -- and was in the past -- but isn't anymore.

arstrube:
here: #871. Not a "good" fix, but should work for now.

arstrube:
> I agree about the double size (I as thinking about larger one, but it must be declared as long double).
And, as I see, I made a mess by displacing int prec

Clang keeps barking, because any size is known at compile time, so the 'else' will never 'execute'. I guess you need to trick it ;-)

arstrube:
Perhaps put `sizeof(double)` in a temporary variable first (perhaps make it volatile for good measure) that should shut it up.......

arstrube:
This seems to work well:

https://github.com/arstrube/cpputest/commit/f6ad79a11b2a6039c7c8730e872f8571a033a56c

mmidor:
I was thinking about template, but if this solution works, then let it be.
Other think I worry about is that that the number of precision digits may vary depending on compiler/platform/version. You know - in C there are only a few constant things and library implementation is not one of them.
So, maybe increment the precision until the returned strings will differ at the last position, and as long as the returned string is longer or different than the previous one. Because, as you said, the `%g*` is tricky ;)

arstrube:
I'll leave that to you -- was just playing around a little and it's your code to begin with :-)

mmidor:
Malloc returned NULL ?
Weird - how to rerun the test ?

arstrube:
The poor thing ran out of memory :-(.

arstrube:
Which is strange, because I ran it not too long ago, with no problems.

arstrube:
Really weird. Nothing else was integrated since I last ran it on my fp branch. You can link your account with Appveyor, then when you push, the test runs on your fork & you can restart it.

examples/nbproject/project.xml
```diff
@@ -0,0 +1,39 @@
...
+                    <name>Debug</name>
+                    <type>1</type>
+                </confElem>
+            </confList>
+            <formatting>
+                <project-formatting-style>false</project-formatting-style>
+            </formatting>
+        </data>
+    </configuration>
+</project>
```
arstrube: A lot of stuff seems to have sneaked into your PR that you probably didn't mean to be in there :-)

mmidor: It's not an accident, if I want to work on this project I need IDE.
You are using VS, by favorite is Netbeans and these three files are not a big deal I think.

arstrube: That wouldn't be for me to decide ;-). Keep in mind that you might be expected to support this if it is there, e.g. if someone else adds some new files, etc.

For the record, I am using Code::Blocks as an IDE and I keep my project files out of the repository (one level above, where I version them in a different repository. I also use automake a lot; saves me surprises when I push :)

basvodde: Usually it would be better to keep them separate. Larger pull requests take me longer to check and thus tend to get delayed. Smaller pull requests are easier and will be integrated faster.

src/CppUTest/TestFailure.cpp
```diff
@@ -190,7 +190,25 @@ DoublesEqualFailure::DoublesEqualFailure(UtestShell* test, const char* fileName,
...
+    {
+        while ((prec < prec_max) && (sexp == sact))
+        {
+            prec++;
+            sexp = StringFrom(expected, prec);
+            sact = StringFrom(actual, prec);
+        }
+    }
+    message_ += createButWasString(sexp, sact);
+
```
basvodde: This code definitively doesn't belong in TestFailure, but probably in String. It doesn't seem to have anything to do with test failures :) Also it has a lot of duplication that can be removed by extracting the functionality out to a separate String methods.

examples/nbproject/project.xml
mmidor: I will remove the NB project files when you approve the changes of course ;)

src/CppUTest/TestFailure.cpp
mmidor: I will try to refactor this next week.

mmidor: I'm looking at these few lines and I can not figure out how this code may be transferred into the SimpleString class. 
DoublesEqualFailure class is supposed to create verbose diagnostic output - and that's what it's doing! 
String is creating formatted double value output when asked, but it has nothing to do with formatting the output, according to how the other String looks like. It's this Failure class job to specify the format precision to achieve valuable message. Am I wrong?

Checking the double values validity is ugly, yet necessary.
Now all I can propose it this:

```C++
    const bool nums_valid = !((PlatformSpecificIsNan(expected)) || (PlatformSpecificIsInf(expected)) ||
                              (PlatformSpecificIsNan(actual)) || (PlatformSpecificIsInf(actual)));
    int prec = 3;
    const int dsize = sizeof(double);
    const int prec_max = dsize == 8 ? 17 : 7;
    SimpleString sexp, sact;

    if (nums_valid)
    {
        while ((prec < prec_max) && (sexp == sact))
        {
            sexp = StringFrom(expected, prec);
            sact = StringFrom(actual, prec);
            prec++;
        }
    }
    else
    {
        sexp = StringFrom(expected);
        sact = StringFrom(actual);
    }
    message_ += createButWasString(sexp, sact);
```

mmidor: After a while of consideration - the first one was better.

arstrube: Personally I think it can't be put into SimpleString (and does belong here) because for it to work, I need to have both sexp and sact,  which I keep comparing until they differ....

mmidor: :+1:

basvodde: @arstrube Why not have a method in SimpleString for comparing and giving the index of difference. Isn't that the same?

In fact, we must have that knowledge already somewhere else also because of the STRCMP_EQUAL.

PR #886, opened 2016-02-23
one proposal to fix #884 Append scope to function name

### LINKED ISSUE #884: Output mock scope at failure
If a scoped mock is used, it' would be helpful to print out it's scope on failure.

## Problem

1. If mocks have the same name - but different scopes, it's hard to say which one is actually meant
1. If there's a typo in the scopes name, it's difficult to trace back the failure

## Examples

Assuming following test:
```cpp
void fn(int i)
{
    mock().actualCall("fn").withParameter("i", i);
}

void fnScope(int i)
{
    mock("scope").actualCall("fn").withParameter("i", i);
}


TEST_GROUP(ScopeTest)
{
    void teardown() override
    {
        mock().checkExpectations();
        mock().clear();
    }
};
```

### Example 1: Same name but different scopes

Both mocks have the same name, but different scopes:

```cpp
TEST(ScopeTest, noScope)
{
    mock().expectOneCall("fn").withParameter("i", 3);
    fn(4);
}

TEST(ScopeTest, withScope)
{
    mock("scope").expectOneCall("fn").withParameter("i", 3);
    fnScope(4);
}
```

***Output (shortened):***

The actual message is twice exactly the same.

```
test/ScopeTest.cpp:31: error: Failure in TEST(ScopeTest, withScope)
	Mock Failure: Unexpected parameter value to parameter "i" to function "fn": <4>
	EXPECTED calls that DID NOT happen related to function: fn
		fn -> int i: <3>
	ACTUAL calls that DID happen related to function: fn
		<none>
	ACTUAL unexpected parameter passed to function: fn
		int i: <4>

test/ScopeTest.cpp:25: error: Failure in TEST(ScopeTest, noScope)
	Mock Failure: Unexpected parameter value to parameter "i" to function "fn": <4>
	EXPECTED calls that DID NOT happen related to function: fn
		fn -> int i: <3>
	ACTUAL calls that DID happen related to function: fn
		<none>
	ACTUAL unexpected parameter passed to function: fn
		int i: <4>
```

### Example 2: Typo in scope name

There's a typo in the scope name:

```cpp
TEST(ScopeTest, typoScope)
{
    mock("Scope").expectOneCall("fn").withParameter("i", 3);
    /*    ^-- typo */
    fnScope(4);
}
```

***Output (shortened):***

It's difficult to say, what *really* caused the test to fail, since the function is called as expected.
```
test/ScopeTest.cpp:37: error: Failure in TEST(ScopeTest, typoScope)
	Mock Failure: Unexpected call to function: fn
		<none>
		<none>
```

cuitoldfish:
let's have a discuss. I would add more test for checking parameter part later

### DIFF
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -83,6 +83,7 @@ public:
     MockNamedValue getData(const SimpleString& name);
 
     MockSupport* getMockSupportScope(const SimpleString& name);
+    
 
     const char* getTraceOutput();
     /*
@@ -148,13 +149,18 @@ private:
 
     MockSupport* getMockSupport(MockNamedValueListNode* node);
     
+    SimpleString* getMockScope(MockNamedValueListNode* node);
+    
+    
     bool hasntExpectationWithName(const SimpleString& functionName);
     bool hasntUnexpectationWithName(const SimpleString& functionName);
     bool hasCallsOutOfOrder();
     
     SimpleString appendScopeToName(const SimpleString& functionName);
     
-    void withScope(const SimpleString& name);
+    
+    
+    
 };
 
 #endif
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -136,7 +136,6 @@ private:
     MockExpectedCallComposite compositeCalls_;
     MockNamedValueComparatorsAndCopiersRepository comparatorsAndCopiersRepository_;
     MockNamedValueList data_;
-    SimpleString scope_; 
         
     bool tracing_;
 
@@ -151,16 +150,11 @@ private:
     
     SimpleString* getMockScope(MockNamedValueListNode* node);
     
-    
     bool hasntExpectationWithName(const SimpleString& functionName);
     bool hasntUnexpectationWithName(const SimpleString& functionName);
     bool hasCallsOutOfOrder();
     
     SimpleString appendScopeToName(const SimpleString& functionName);
-    
-    
-    
-    
 };
 
 #endif
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -135,7 +135,8 @@ private:
     MockExpectedCallComposite compositeCalls_;
     MockNamedValueComparatorsAndCopiersRepository comparatorsAndCopiersRepository_;
     MockNamedValueList data_;
-
+    SimpleString scope_;
+        
     bool tracing_;
 
     void checkExpectationsOfLastCall();
@@ -150,6 +151,10 @@ private:
     bool hasntExpectationWithName(const SimpleString& functionName);
     bool hasntUnexpectationWithName(const SimpleString& functionName);
     bool hasCallsOutOfOrder();
+    
+    SimpleString appendScopeToName(const SimpleString& functionName);
+    
+    void withScope(const SimpleString& name);
 };
 
 #endif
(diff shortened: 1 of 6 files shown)


cuitoldfish:
those new test case has showed the updated failure report, such as
```.cpp
+TEST(MockCallTest, expectOneCallInScopeButNotHappen)
 +{
 +
 +    MockFailureReporterInstaller failureReporterInstaller;
 +
 +    MockExpectedCallsListForTest expectations;
 +    expectations.addFunction("scope::foobar");
 +    MockExpectedCallsDidntHappenFailure expectedFailure(mockFailureTest(), expectations);
 +
 +    mock("scope").expectOneCall("foobar");
 +    mock().checkExpectations();
 +
 +    CHECK_EXPECTED_MOCK_FAILURE(expectedFailure);
 +
 +}
```

but here is more clear picture:
```.cpp
test/ScopeTest.cpp:37: error: Failure in TEST(ScopeTest, typoScope)
    Mock Failure: Unexpected call to function: scope::fn
        <none>
        <none>

test/ScopeTest.cpp:31: error: Failure in TEST(ScopeTest, withScope)
    Mock Failure: Unexpected parameter value to parameter "i" to function "scope::fn": <4>
    EXPECTED calls that DID NOT happen related to function: scope::fn
        scope::fn -> int i: <3>
    ACTUAL calls that DID happen related to function: scope::fn
        <none>
    ACTUAL unexpected parameter passed to function: scope::fn
        int i: <4>

test/ScopeTest.cpp:25: error: Failure in TEST(ScopeTest, noScope)
    Mock Failure: Unexpected parameter value to parameter "i" to function "fn": <4>
    EXPECTED calls that DID NOT happen related to function: fn
        fn -> int i: <3>
    ACTUAL calls that DID happen related to function: fn
        <none>
    ACTUAL unexpected parameter passed to function: fn
        int i: <4>
```

include/CppUTestExt/MockSupport.h
```diff
@@ -135,7 +135,8 @@ class MockSupport
     MockExpectedCallComposite compositeCalls_;
     MockNamedValueComparatorsAndCopiersRepository comparatorsAndCopiersRepository_;
     MockNamedValueList data_;
-
+    SimpleString scope_;
```
basvodde: The name is already stored in the GetData. So, there shouldn't be need for another scope_ field, correct?

basvodde: I guess you can't get the name from GetData, but it would probably be best to eliminate the duplication, or did you already?

cuitoldfish: yes, I can't easily get the name from GetData, previously global mock has one uni-direction association to mockNamedValueNode which own scope mock and it's name. the data_ in scope mock is initialized as empty so getData in scope mock would return empty.

but you are right, scope name can be filled to the mockNamedValue list, the data_ member.

Let's say my updated commit for this proposl!

basvodde:
Hmm, yeah I liked the name version more so there is no dynamic memory allocations for it. However, it would need to use the name also in places where it was earlier used. Let me check the previous version again to be more specific.

basvodde:
Uhm, the other ones isn't open anymore?

cuitoldfish:
this PR can be restored to previous version, or do you prefer to open another PR then we can compare both at the same time?

arstrube:
I don't know about @basvodde but me, I'd like to another PR better :-)

basvodde:
Another PR to compare would be best.

PR #888, opened 2016-02-24
Pending fixes for cl2k (Cl2000, c2000)

arstrube:
We discussed these before. This is for documentation, rather than integration......


include/CppUTest/UtestMacros.h
```diff
@@ -249,8 +249,10 @@
 #define BITS_EQUAL_TEXT(expected, actual, mask, text)\
   BITS_LOCATION(expected, actual, mask, text, __FILE__, __LINE__)
 
+#define ROBUST_SIZEOF(type) (2*sizeof(type)/sizeof(short)) /* For the Cl2000, sizeof(char) == sizeof(short) == 1 */
```
basvodde: Why would sizeof not be robust?

arstrube: Ah, good question. It trips over the fact that the TI processor has no byte, in fact nothing of 8 bits wide. So the calculations are off. Maybe robust isn't exactly the best way to describe it....

I must say I am not fond myself of doing this for just one quirky processor that nobody seems to be using anymore. Theoretically, other system without uint8_t or equivalent are possible; but this is the only one I've ever seen.

arstrube: The weird calculation in the macro makes up for this by multiplying by 2 and dividing by sizeof(short), thus for a type that ought to be a byte, 1 is returned.

arstrube: It didn't use to be necessary to consider this, but some of the newer features of CppUTest make it necessary. That's why I am not really keen on supporting the TI processor / Cl2k anymore.

arstrube: Maybe `CHEATING_SIZEOF()` would be more appropriate :D. Pretending there's data 8 bits wide when there isn't ;-)

arstrube:
If memory serves me, sizeof(char) == 1, sizeof(short) = 1, sizeof(long) == 2 etc. In other words, sizeof() returns multiples of the smallest unit, which is a 16 bit word. I really can't think of any other way of dealing with this than the crazy macro above.

include/CppUTest/UtestMacros.h
basvodde: What is the problem that is caused by the sizeof? Is it just in the test, a different output?

arstrube: I don't remember exactly. There are definitely test failures. There are display errors, e.g. when 1110111100000001 should have been displayed, 00000001 would be displayed, and so on
For the TI, a "byte" is 16 bit, hence sizeof is off compared to other platforms, by a factor of 2.

I think ifdefing testcases is not appropriate here, because when I display a short, the expected result really is 0xABCD and not 0xCD, and so on.

The problem wasn't there before, so I think only the BITS_EQUAL stuff is affected.

Same as the need for a typedef for a C++ function became only necessary with supporting functionPointerValue (something I am a little critical of anyway, because a function that's not void(void) still needs to be cast).

arstrube: I like the macro, because it can smooth over the difference without the need for lots of #ifdef.

arstrube: Perhaps call it NORMALIZED_SIZEOF()?

arstrube: Since the use of sizeof() will be consistent internally for a given compiler, I think it is safe to say that the only problem is the assumption that a size of one means 8 bits (vs actually 16 bits) when doing bit comparison and display.

As before, I am not sure this needs to be fixed. Version 3.7.2 of CppUTest is still fine with the TI. Nobody has complained so far.

basvodde: I'd actually prefer to have the test case under #if BYTE_IS_16_BITS or so.

I don't like the sizeof macro and don't think we need to shrink the sizeof a byte when the platform clearly has 16 bit bytes...

basvodde:
And, I'd like to support platforms such as these. They keep the code more portable and thus will make porting to other platforms easier.

Exactly because of this kind of problems :)

src/CppUTest/SimpleString.cpp
```diff
@@ -631,7 +631,7 @@ SimpleString StringFromBinaryWithSizeOrNull(const unsigned char* value, size_t s
 SimpleString StringFromMaskedBits(unsigned long value, unsigned long mask, size_t byteCount)
 {
     SimpleString result;
-    size_t bitCount = (byteCount > sizeof(unsigned long)) ? (sizeof(unsigned long) * 8) : (byteCount * 8);
+    size_t bitCount = (byteCount > ROBUST_SIZEOF(unsigned long)) ? (ROBUST_SIZEOF(unsigned long) * 8) : (byteCount * 8);
     const unsigned long msbMask = (((unsigned long) 1) << (bitCount - 1));
 
     for (size_t i = 0; i < bitCount; i++) {
```
arstrube: If we get rid of the macro, we need to get rid of the magical number "8" as well. Perhaps define a macro like BITS_PER_BYTE which will be 8 for most systems and 16 for systems with 16 bit bytes.

How and where do you suggest this be defined?

include/CppUTest/UtestMacros.h
arstrube: You have a point, of course, but in this particular case, I object to #ifdefing the test case, because for the 16 bytes platform, the result displayed will be WRONG. So we do need to fix the display. If you have a cleaner idea than the macro, I'm willing to implement that.

src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -32,6 +32,8 @@
 #include "CppUTestExt/MockSupport.h"
 #include "CppUTestExt/MockSupport_c.h"
 
+typedef void Cpp_Func();  /* Cl2000 requires casts to C++ function */
+
```
arstrube: You objected to this typedef as well. It would have to be part of the package, but we might want it in a place where it is more generally accessible. BTW I tried all kinds of ways to cast the function pointer without a typedef and all compilers went beserk. So, if you can cast somehow other than with a typedef, that would be okay. But a typedef is cleaner in any case.

I know it sounds silly, but the cl2000 is fussy about this. Perhaps the warning can be disabled, if so, I'd go for that.

arstrube:
If we want to do this for the sake of portability, then in my book, getting rid of the magical number "8" is the way to go :)

arstrube:
GCC has __CHAR_BIT__ which "you should not use directly".

arstrube:
The cl2000 has CHAR_BIT. Gcc does, too, in '<limits.h`. So next week, I'll change the code to use CHAR_BIT.

src/CppUTestExt/MockSupport_c.cpp
basvodde: I'd prefer the cast without a typedef if that is possible

include/CppUTest/UtestMacros.h
basvodde: So, why can't we fix the display for 16 byte platforms then? That feels like the right and obvious thing to do....

basvodde:
Why CHAR_BIT ?

arstrube:
You ask me? Ask those who decide such things ;-). We can, of course, define our own CPPU_TEST_... , based on the above.

src/CppUTestExt/MockSupport_c.cpp
arstrube: Okay, you tell me how :-)

arstrube: (it's not that I didn't try)

basvodde:
#if sizeof(char) == 8 ?

src/CppUTestExt/MockSupport_c.cpp
basvodde: Did you try: (void (*)(void)) 

That works under clang at least ;)

Andne:
Doesn't `sizeof(char)` always return 1?  Also, I think sizeof is a compiler operation, so preprocessor logic based on it wouldn't work I think.

src/CppUTestExt/MockSupport_c.cpp
arstrube: Yes, but that's besides the point.

Tthe point is that the cast must be __to a c++ function__. I dont'know any syntactically correct way to do that inline.

The typedef works, because it is in the context of C++ code. A typdef could also be put with in extern "C++" {}.

arstrube:
@Andne  is right. That's why I propose to use CHAR_BIT, because it is defined to 8 normally and to 16 for the TMS320 & friends.

(@basvodde sorry I got you wrong, I thought you were asking why it is called CHAR_BIT)

basvodde:
Ok, but please avoid using features that are said not to be used :)

src/CppUTestExt/MockSupport_c.cpp
basvodde: The cast is also in C++ context, so it would be the same :)

src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -382,7 +384,7 @@ MockExpectedCall_c* withConstPointerParameters_c(const char* name, const void* v
 
 MockExpectedCall_c* withFunctionPointerParameters_c(const char* name, void (*value)())
```
arstrube: This is a C context, so the pointer is c.

src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -382,7 +384,7 @@ MockExpectedCall_c* withConstPointerParameters_c(const char* name, const void* v
...
 {
-    expectedCall = &expectedCall->withParameter(name, value);
+    expectedCall = &expectedCall->withParameter(name, (Cpp_Func*)value);
```
arstrube: And this is a C++ context, so the above c pointer must be cast to c++.

arstrube: I can post the original compiler complaints :P

arstrube: I agree that (void(*)*()) __should__ work.....

basvodde: At least (void (*)(void)) compiled without any problem. And this cast is in C++ context. Could you try it for CL2k?

arstrube: This can be done; however, since I may not add a comment to explain what these apparently redundant casts are for; some bright mind will remove it in the future, at which point, since Travis will be fine (and nobody is going to complain on the Cl2000 side) the Cl2000 build will silently break :P.

basvodde: I doubt anyone will remove the casts. It is so weird that he would at least check the history. And besides, you are checking all commits anyway :P

arstrube: Apologies. Spoke too soon. Had to edit above comment.

A (void(*)()) cast in C++ context, unfortunately, does not cast away the C-ness. Only a typedef does that. I knew I had tried this one before the typedef -- what don't I do to please BV ;-)

arstrube: So, what to do now?

arstrube:
I suggest I make a new pull request from the part we can (hopefully!) agree on, and keep the function pointer stuff in this one for now.

Also, because running all test on the simulator takes forever, that will probably only happen on Wednesday....

arstrube:
Oops. The function pointer stuff somehow fell out of this one. So I'll do it the other way round :P

include/CppUTest/CppUTestConfig.h
```diff
@@ -189,6 +189,16 @@
 #endif
 #endif
 
+/* Handling of systems with a different byte-width (e.g. 16 bit).
+ * Since CHAR_BIT is defined in limits.h (ANSI C), use default of 8 when building without Std C library.
```
basvodde: This should be defined by autotools in the future :)

basvodde:
Excellent. Can I merge it?

arstrube:
Yes, you can merge this one. Unfortunately, there is still the other one before all will compile again ;-)

PR #890, opened 2016-02-26
CHECK_EQUAL_C_BITS for comparing bits in pure C

arstrube:


include/CppUTest/TestHarness_c.h
```diff
@@ -121,6 +124,8 @@ extern void CHECK_EQUAL_C_STRING_LOCATION(const char* expected,
         const char* actual, const char* fileName, int lineNumber);
 extern void CHECK_EQUAL_C_POINTER_LOCATION(const void* expected,
         const void* actual, const char* fileName, int lineNumber);
+extern void CHECK_EQUAL_C_BITS_LOCATION(unsigned int expected, unsigned int actual,
```
arstrube: Not sure why `int` is used by existing code? So I use `unsigned int` here. But aren't we comparing `long` in the assertion methods, and implying `long` in the names of the Cpp macros? So why `int` for C?

arstrube:
@klusark you might find this one useful, too.

klusark:
@arstrube Thanks! That does look useful!

PR #891, opened 2016-02-26
Support for 64/128 bit integer comparison

mmidor:
As the default LONGS_EQUAL act different depending on platform,
user should be warned that this macro may not work as expected.
Further on, more universal macro should be provided, like
MEMCMP_EQUAL(), but presenting output as hexadecimal number,
not a hexadecimal memory dump.

I created INTS_EQAL() macro - just a proposal. 
It works like a MEMCMP_EQUAL(), but produces following output:
    expected <0x00000000000000000000000000000001>
    but was  <0x00000000000000000080000000000000>

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -628,6 +628,23 @@ SimpleString StringFromBinaryWithSizeOrNull(const unsigned char* value, size_t s
     return (value) ? StringFromBinaryWithSize(value, size) : "(null)";
 }
 
+SimpleString StringFromAnyInteger(const unsigned char* value, size_t size)
+{
+    SimpleString result = "0x";
+
+    for (int i = size-1; i >= 0; i--)
+    {
+        result += StringFromFormat("%02X", value[i]);
+    }
+
+    return result;
+}
+
+SimpleString StringFromAnyIntegerOrNull(const unsigned char* value, size_t size)
+{
+    return (value && size) ? StringFromAnyInteger(value, size) : "(null)";
+}
+
 SimpleString StringFromMaskedBits(unsigned long value, unsigned long mask, size_t byteCount)
 {
     SimpleString result;
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -632,9 +632,8 @@ SimpleString StringFromAnyInteger(const unsigned char* value, size_t size)
 {
     SimpleString result = "0x";
 
-    for (int i = size-1; i >= 0; i--)
-    {
-        result += StringFromFormat("%02X", value[i]);
+    for (size_t i = 0; i < size; i++) {
+        result += StringFromFormat("%02X", value[size-1-i]);
     }
 
     return result;
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -318,6 +318,15 @@ BinaryEqualFailure::BinaryEqualFailure(UtestShell* test, const char* fileName, i
 	}
 }
 
+IntsEqualFailure::IntsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, const unsigned char* expected,
+                                       const unsigned char* actual, size_t size, const SimpleString& text)
+: TestFailure(test, fileName, lineNumber)
+{
+    message_ = createUserText(text);
+
+	message_ += createButWasString(StringFromAnyIntegerOrNull(expected, size), StringFromAnyIntegerOrNull(actual, size));
+}
+
 BitsEqualFailure::BitsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, unsigned long expected, unsigned long actual,
                                    unsigned long mask, size_t byteCount, const SimpleString& text)
 : TestFailure(test, fileName, lineNumber)
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -322,6 +322,19 @@ IntsEqualFailure::IntsEqualFailure(UtestShell* test, const char* fileName, int l
                                        const unsigned char* actual, size_t size, const SimpleString& text)
 : TestFailure(test, fileName, lineNumber)
 {
+    if (size <= sizeof(long))
+    {
+        long lexp = 0, lact = 0;
+        for (size_t i = 0; i < size; i++)
+            ((unsigned char*)&lexp)[i] = ((const unsigned char *)expected)[i];
+        for (size_t i = 0; i < size; i++)
+            ((unsigned char*)&lact)[i] = ((const unsigned char *)actual)[i];
+
+        LongsEqualFailure f(test, fileName, lineNumber, lexp, lact, text);
+        message_ = f.getMessage();
+        return;
+    }
+
     message_ = createUserText(text);
 
 	message_ += createButWasString(StringFromAnyIntegerOrNull(expected, size), StringFromAnyIntegerOrNull(actual, size));
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -318,7 +318,7 @@ BinaryEqualFailure::BinaryEqualFailure(UtestShell* test, const char* fileName, i
 	}
 }
 
-IntsEqualFailure::IntsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, const unsigned char* expected,
+AnyIntsEqualFailure::AnyIntsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, const unsigned char* expected,
                                        const unsigned char* actual, size_t size, const SimpleString& text)
 : TestFailure(test, fileName, lineNumber)
 {
--- a/tests/AllTests.cpp
+++ b/tests/AllTests.cpp
@@ -33,6 +33,11 @@ int main(int ac, const char** av)
     CHECK(true);
     LONGS_EQUAL(1, 1);
 
+#ifdef __GNUC__
+//    __int128_t i128 = 1;
+//    LONGS_EQUAL(1, i128);
+#endif
+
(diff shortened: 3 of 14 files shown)


mmidor:
Anybody knows how to fix problem generated by Travis `wcl` compiler ?
It rejects construction like this:

``` C
long func() { return -1; }

void main() 
{
    // here x becomes eq sizeof(long), 
    // the func() is not called as sizeof is evaluated at compile-time !
    int x = sizeof(func()); 
}
```

arstrube:
I might be able to look into this on Monday.

What is actually the reason you are using sizeof(func()) rather than sizeof(long) ?

mmidor:
I changed the macro LONGS_EQAL to warn the user:

```C
  #define LONGS_EQUAL_LOCATION(expected, actual, text, file, line)\
   { if ((sizeof(expected) > sizeof(long)) || (sizeof(actual) > sizeof(long))) \
         UtestShell::getCurrent()->print( \
            "WARNING:\n\tSize of `expected` or `actual` is bigger than size of `long int`\n\tConsider use of INTS_EQUAL().\n", \
             file, line); \
     UtestShell::getCurrent()->assertLongsEqual((long)expected, (long)actual, text, file, line); }
```

Now let's imagine test like this:

```C
long long func(int a) { return a * a * a; }
LONGS_EQUAL(125, func(5));
```
It is correct, however testing `sizeof(func())` seems to be unacceptable for the mentioned compiler.

mmidor:
The problem would disappear if assertLongsEqual() accepted long long argument. But long long int is not a standard :-(

mmidor:
What about this issue ? I saw you added some compiler parameter to ignore the `sizeof` side effect. Did it helped ?

arstrube:
Of course it did :). Now I see why you killed Appveyor.... ;-)

arstrube:
Will only help you though, if

1. you merge cpputest/master into your PR -- or --
2. you rebase your PR onto cpputest/master

mmidor:
Yupiyayeeey!

Now, let's discuss how to proceed. To be honest, I have no idea how to support all kinds of usage, including:
* ANYINTS_EQUAL(1, 1)
* ANYINTS_EQUAL(integer, 1)
* ANYINTS_EQUAL(1, integer)
* ANYINTS_EQUAL(1, func(123))

Ability to use C++11 and `decltype` would solve all the problems. But now, the macro ANYINTS_EQUAL can accept only pointers to variables of the same size. Exactly as MEMCMP_EQUAL. Very inconvenient.

And, of course, my idea was not to introduce another new special macro (there are already a lot of them), but make the existing LONGS_EQUAL more universal and robust (to not loose silently int64 data on 32bit platforms anymore).

How about templates? They may help solving this.

src/CppUTest/SimpleString.cpp
```diff
@@ -628,6 +628,22 @@ SimpleString StringFromBinaryWithSizeOrNull(const unsigned char* value, size_t s
     return (value) ? StringFromBinaryWithSize(value, size) : "(null)";
 }
 
+SimpleString StringFromAnyInteger(const unsigned char* value, size_t size)
```
basvodde: Should this be a hex string from any integer?

src/CppUTest/TestFailure.cpp
```diff
@@ -318,6 +318,28 @@ BinaryEqualFailure::BinaryEqualFailure(UtestShell* test, const char* fileName, i
...
 
+AnyIntsEqualFailure::AnyIntsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, const unsigned char* expected,
+                                       const unsigned char* actual, size_t size, const SimpleString& text)
+: TestFailure(test, fileName, lineNumber)
+{
+    if (size <= sizeof(long))
+    {
+        long lexp = 0, lact = 0;
+        for (size_t i = 0; i < size; i++)
+            ((unsigned char*)&lexp)[i] = ((const unsigned char *)expected)[i];
```
basvodde: Can you add this loop in a helper function (inside the class, private) that describes what it does and call it twice?

tests/AllTests.cpp
```diff
@@ -33,6 +33,19 @@ int main(int ac, const char** av)
...
+        __int128_t a128 = 1;
+        __int128_t b128 = 1;
+        LONGS_EQUAL(a128, b128);
+
+        // expected <0x00000000000000000000000000000001>
+        // but was  <0x00000000000000000080000000000000>
+        __int128_t c128 = 1;
+        __int128_t d128 = 1L << 55;
+        ANYINTS_EQUAL(&c128, &d128, sizeof(c128));
+#endif
```
basvodde: This can't be here like this. What is it and why is it #it 0-ed

tests/TestFailureTest.cpp
```diff
@@ -333,6 +333,38 @@ TEST(TestFailure, BinaryEqualFullWidth)
...
+    FAILURE_EQUAL("expected <256 0x100>\n\tbut was  <512 0x200>", f);
+}
+
+TEST(TestFailure, IntsEqualFourBytes)
+{
+    const unsigned char expectedData[] = {0x00, 0x01, 0x00, 0x00};
+    const unsigned char actualData[] = {0x00, 0x02, 0x00, 0x00};
+    AnyIntsEqualFailure f(test, failFileName, failLineNumber, expectedData, actualData, sizeof(expectedData), "");
+
+    if (sizeof(long) >= 4)
```
basvodde: Split this into two tests. An if in the test is a smell

basvodde:
I couldn't find a test related to the new warning that was introduced?

tests/TestFailureTest.cpp
arstrube: 1. In my mind, the first part of the "if" just means making a bad result okay. If I understand correctly what you are doing, then the outcome of your comparison should be the same, irrespective of size of long? After all, you are passing in the integer data as an array of bytes?

2. There  is no test for 64 bit or 128 bit.

3. An int can be negative. I don't think your methods can handle negative values (yet?).

4. What about little-endian vs. big-endian?

arstrube:
I'm really not sure it can be done, with the constraints that we have set ourselves (must work without Std C++ library, must not use templates, must not use long long).

Of these I think allowing long long would be the easiest (would have to be a cpputest option, like using/not using Std Cpp library) or perhaps handle by Automake?

Without that, some tests will not be possible (e.g. passing in an actual 64/128 bit integer to the macro).

tests/TestFailureTest.cpp
basvodde: Endianness seems an issue.... Did you take that into account @mmidor

mmidor:
This is not completed yet - I made this to show my idea, to go around and try to find better solution than LONGS_EQUAL and support not only int32 but int64 and int128 as well if the platform supports it. 

But I don't want to create just another macro, there are o much of them already. Bytes, Longs, Ints, where is it going ? 
One test for integers should be enough, but able to accept numbers of any size and type (value like 15, variable, function call result). Its difficult. 

Maybe LONGS_EQUAL could use input parameter dependent on what maximum length int the current platform supports ? CMake performs tests during the build to find out what features the current platform support - can we use these results ?

basvodde:
@arstrube An #if is possible, as long as it isn't in the header and changing the class memory layout, like we had in the previous pull request...

arstrube:
@mmidor - if I am not entirely mistaken, your approach needs to pass a pointer to the integer (so it can be interpreted as an array). This would probably not be possible to reconcile with

* ANYINTS_EQUAL(1, func(123))

src/CppUTest/TestFailure.cpp
mmidor: Of course

tests/AllTests.cpp
mmidor: This is temporary, will be removed

arstrube:
I totally agree with @mmidor that the goal should be to make LONGS_EQUAL more flexible.

*Normally* we use function overloading, but we can't for long long, because we'd need to #ifdef one of the overloaded methods, and we have seen this is not a good idea.

The "old-fashioned" style, on the other hand, would have been (and C compilers still do that) with low-level macros, something like:
```c++
#define STRING_FROM_ANY_INTEGER(x) \
    ((sizeof(x) == 8) ? StringFromAnyInteger((const unsigned char*)&x, 8) : \
    (siseof(x) == 16) ? StringFromAnyInteger((const unsigned char*)&x, 16) : \
    StringFrom(x))
```
This way, if I am not entirely mistaken, it would not be necessary to pass the size to the test failure, and it should be okay to compare different size values. It would also be easier to put in #ifdef's and suitable warnings to handle different platforms.

arstrube:
For example,
```c++
#ifdef CPPUTEST_PLATFORM_SUPPORTS_LONG_LONG 
    #define STRING_FROM_ANY_INTEGER(x) \
         ((sizeof(x) == sizeof(long long)) ? StringFromAnyInteger((const unsigned char*)&x, sizeof(long long)) : \
         StringFrom(x))
#else
    #define STRING_FROM_ANY_INTEGER(x) StringFrom(x))
#endif
```
This would keep the standard behavior for anything but long long, which would under windows be 64bit, and under 'nix probably 128bit.

basvodde:
With long long support being integrated, is this still relevant?

mmidor:
I dont know what you are talking about, I see no long long in function arguments.
And I hate `unsigned long long` in the code, it's ugly and makes problems when porting code between platforms. Someone invented uint64_t which is nice clear and always of expected size. Just a digression. 

Anyway, this pull request is supposed to be abandoned.

basvodde:
There is a seperate pull request that adds long long support. It doesn't use long long directly, but not the uint64_t either :)

arstrube:
Yeah.... uintxx_t is nice and clean :-/. Practically every embedded project I know uses it in some form. Some prefer their own typedefs (s8, u16, or whatever they call it), which can lead to problems. I know this approach is not without its drawbacks either, but it works well in situations where bit width matters a lot (which, in embedded development, is most situations :-P)

PR #893, opened 2016-03-02
fix #884 Option2 for append scope name

### LINKED ISSUE #884: Output mock scope at failure

cuitoldfish:


cuitoldfish:
I would say I like this option most.
In the current design, `mockNamedValue` is the class to describe the "value", and it's close for changing the class content, the relation among members, and the relation between `mockNamedValue` and it's member,  but open for the modification for the value members, the value can have it's own discretion on type(void*, int, and so on) and off course behavior, member variable.
mock support is just one instance of that "value", and it deserve it's own scope_ member to make itself more definitely.
what's more, the duplication here is not so bad because scope in `mockNamedValue` and `mockSupport` both from the same sources then no possibility to have the non-uniformity between the two scope.

basvodde:
Shouldn't you set the scope in getMockSupportScope, when the scope gets created?

cuitoldfish:
I currently do set scope  in getMockSupportScope, just after clone()

basvodde:
Ah yes, I missed that. Only thing, could you rename the withScope to just setScope or so ? I believe the withScope/setScope isn't going to be used by CppUTest users themselves, correct? Now it suggests it is similar to the other with methods, whereas it probably isn't.

cuitoldfish:
I agree, this function should be confined to be used inside the mockSupport class.

how about change the name to `completeMockSupportWithScope`?  and shall we declare this small function to inline?

basvodde:
How about adding the scope to clone ?

cuitoldfish:
I was thinking this yesterday, and also declare the scope_ to const, but I have concern to change the meaning of clone() because we need to pass the scope name to clone() function

basvodde:
Clone is only used there, right?

cuitoldfish:
yeah, I mean the `clone` has one world known meaning, extra parameter will change that and then add misunderstanding to the readebility

basvodde:
I think it is ok, especially if you consider scope to be kinda equal to the name of the Mock. So, lets just change clone and have a default parameter to "".

cuitoldfish:
OK. and I prefer to rename scope_ to mockName_, and make it const which can only be set when mockSupport is constructed, no possibility to get changed

basvodde:
Thanks

PR #894, opened 2016-03-07
Pending fixes for cl2k (Cl2000, c2000): C++ vs. C function pointer problem

arstrube:
Using a typedef inside the correct space (C or C++) is the only way to make a function pointer C-ish or Cpp-ish. It is also elegant and simple.

I won't waste any more time on attempted hacks that invariably don't work.


arstrube:
In MockSupport_cTest.cpp,
```c++
extern "C" {
    typedef void C_Func();
}
```
could be replaced with:
```c++
extern "C" {
    void (*defaultFunctionPointer)() = (void(*)()) 1;
    void (*expectedFunctionPointer)() = (void(*)()) 10;
}
```
Instead of testing `(void(*)())1` one then tests `defaultFunctionPointer`
for testing. This does most definitely NOT make for cleaner code, though :-[

A similar method for MockSupport_c.cpp would NOT work, since on would need to assign a C++ function pointer to a C function pointer at some point, which is an error (not a Werror).

arstrube:
@basvodde why are you so opposed to using a typedef?

1. It is clean
2. It is self-documenting (name could be made even more explicit)
3. It is limited to source file scope and not propagated via headers

basvodde:
Eheh

Well, 
1) I don't think it is cleaner
2) In this case, I'd prefer seeing the cast than having to look for the typedef
3) So is a cast.

There were 3 casts or so, I don't see how adding a typedef makes a difference. I felt it to be less clear.

arstrube:
Yes, but the cast has one distinct drawback. It is impossible :-).

arstrube:
Cast without typedef is just simply no option, because C++ doesn't really provide a way to cast a function pointer to either C or C++ (because for all 'normal' compilers, there's no difference to start with).

The main problem is that the compiler cannot find a 'matching' method (e.g. one that takes a C function pointer). By trying to provide one, that would move the problem to the header, which is worse. 

Frankly, I am getting very tired of this discussion. I am willing to do what can be done, but this is a dead end and really, what I did in the tests to avoid the typedef stinks. Big.

arstrube:
Cast without typedef doesn't work.

basvodde:
Well, in that case we need a typedef.

Please rename it thouhh, Cpp_Func is a bit to general. But are you sure the cast didn't work? I'm surprised about that, it is weird...

arstrube:
Yes, renaming is fine with me. Do you have a suggestion, perhaps? Perhaps it should have _for_Cl2000 in the name? I would rather take a name that you see fit, seeing that I can't make the typedef go away.

And can I leave the typedef in the tests, or would you prefer the code as is in this PR? I don't like it, but it does work, so I'm okay with it.

Yes, unfortunately, by now I am absolutely 110% sure. There was only one context where it did work (which I moved to the other PR), and in the tests, I can cheat by defining functions to pass as pointers within extern "C". (That cheat won't work for the source code, because at some point I would have to cast the argument (C) and assign it to the variable (C++), and for that cast I would need the typedef.... you see, it's a lot like running in circles really fast :-P

The Cl2000 compiler is weird indeed. Apparently it considers the body of a function defined as extern "C" to be "c-ish", even in C++. That's why a normal cast doesn't make it "c++-ish". Also, generally in C++, It is also not possible to declare anything as extern "C" within a function body. 

Usually a cast is the hint we give the compiler to choose a particular overloaded function, right? Is there any other way to guide / force the compiler's hand? If so, that might be one last option that could be explored. Someone was talking about it in another context, but it was a bit over my head.

arstrube:
> The Cl2000 compiler is weird indeed. Apparently it considers the body of a function defined as extern "C" to be "c-ish", even in C++. That's why a normal cast doesn't make it "c++-ish". Also, generally in C++, It is also not possible to declare anything as extern "C" within a function body.

I have to take that statement back. It is fussy, yes, but not weird. It is actually behaving as told :-P

basvodde:
cpputest_cpp_function_pointer ?

arstrube:
I thought I was onto something.... unfortunately, no luck. I always end up with one of the two problems below, eventually requiring a cast with a typedef somewhere. But perhaps anyone has an idea?

__This won't link__
```c++
// no extern "C" context; linker fails
{
    expectedCall = &expectedCall->withParameter(name, value);
    return &gExpectedCall;
}
```
```
 undefined                         first referenced 
  symbol                               in file
 ---------                         ---------------- 
 _withFunctionPointerParameters_c  .......\\MockSupport_c.obj
```

__This won't compile__
```c++
extern "C" MockExpectedCall_c* withFunctionPointerParameters_c(const char* name, void (*value)())
{
    expectedCall = &expectedCall->withParameter(name, (void(*)())value);
    return &gExpectedCall;
}
```
```
"MockSupport_c.cpp", line 389: error #306: no instance of overloaded function "MockExpectedCall::withParameter" matches the argument list
            argument types are: (const char *, void (*)() C)
            object type is: MockExpectedCall
1 error detected in the compilation of "MockSupport_c.cpp".
```
Interestingly, all other functions (that do not contain a function pointer) link perfectly fine without being defined in an extern "C" context (the declaration in the header alone is enough).

arstrube:
> cpputest_cpp_function_pointer

Will do. Unless you have an idea regarding my comment above.

basvodde:
How about creating a small function that does the assignment. Put that outside of the extern "C" block. Would that work?

arstrube:
I had that idea, too. There is one snag. The assignment ends up needing a cast... :-(. 

What I'd really like to be able to do is
```c++
extern "C++" 
{
}
``` 
inside a function body, While extern "C++" {} is perfectly legal, it is not syntactically possible to use inside a function body.

I was hoping that extern "C" on the declaration line the function body would be 'spared', but that isn't the case.

The typedef seems the only way to create a type in C-land and then use it in Cpp-land, and vice versa.

basvodde:
But it is ok if the assignment needs a cast, right? Can we move the assignment outside the extern "C" block and then have the cast in just one place?

arstrube:
I was wondering, too - should we really have the function pointer parameter stuff in CppUTest? The colleague who contributed it obviously had a similar problem with on a specific platform. But, shouldn't that kind of problem be solved on user level? Now we kind of pulled the problem (casting function pointers) inside CppUTest and we have to tackle it there, actually breaking support for the Cl2000 (unless we work with the typedef). What do you think? Then again, the cast from void* to function pointer and vice versa can lead to undefined behavior on some platforms (TI apparently being one of them), so maybe we really ought to include this functionality.....

arstrube:
We can have the cast in just one place, yes, but we would still require a typedef for it.

basvodde:
Why? The cast would be outside of the extern "C" so then we wouldn't need the typedef, correct?

arstrube:
You need a function that accepts the one and returns the other. That can never be done within the same context, if it could be done, we wouldn't need the function or the cast.

arstrube:
this works:
```c++
extern "C"
{
    typedef void (*cpputest_c_function_pointer)();  /* Cl2000 requires cast to C++ function */    
}

void (*c_function_pointer_to_cpp(cpputest_c_function_pointer value))()
{
    return (void(*)())value;
}
```
Personally, I don't see an advantage though.

arstrube:
The above function gets called where the casts would have been.

arstrube:
Should I make the above change? The typedefed cast would then only be in one place, although the code would look pretty much the same. If we ever get to remove this, then it's much of a muchness to remove the casts, or to remove the function calls, in my opinion.

arstrube:
P.S. Appveyor needs to be restarted, had a hickup:
```
Build started
git clone -q https://github.com/cpputest/cpputest.git C:\projects\cpputest
fatal: unable to access 'https://github.com/cpputest/cpputest.git/': 
Failed to connect to github.com port 443: Timed out
Command exited with code 128
```

basvodde:
No, I mean you actually pass the pointer (reference or pointer pointer) to the function as a parameter and set it in the function. That way you don't need any typedef anymore, correct?

arstrube:
Hmm. How would you change my function above to do that? 

And how would you dereference / assign the pointer (ultimately fp "C") to the other pointer (ultimately fp "C++" without the compiler knowing about it?

I mean, isnt that what my above function does? Accept a C pointer (with the aid of a typedef) and return it as a C++ pointer?

arstrube:
How would passing it as a reference / pointer pointer change that?

arstrube:
Cast the pointer  pointer to void* ? Isn't that entering the domain of potentially dangerous hacks? Personally, I'd rather have one typedef that's used in one place :-/

arstrube:
@basvodde voilà monsieur: https://github.com/arstrube/cpputest/pull/4

If you want to go that route, you may copy that code, put it in your own pull request and change it in any way you see fit.  YOU MAY NOT ASSOCIATE IT WITH OR CREDIT IT TO MY NAME.

I will run your code past the Cl2000 for you for testing.

basvodde:
No, I didn't mean that, I mean something like set_pointer( void (**)(void), void (*)(void))

Not sure if that would work, but the implementation would basically be:

out = in;

Plus perhaps a cast. Or would you then get the compile error when passing the function pointer to this helper function?

arstrube:
Could be done with a typedef. Typedef necessary because it has to be either C (both pointers C) or C++ (both pointers C++). So one needs a typedef. Do you like that better than:
```c++
extern "C"
{
    typedef void (*cpputest_c_function_pointer)();  
}

void (*c_function_pointer_to_cpp(cpputest_c_function_pointer value))()
{
    return (void(*)())value;
}
```
It could be done, but the typedef would be needed as well. I don't mind which one.

basvodde:
Seeing all the alternatives, I think this one is best.

Thanks for your patience

arstrube:
There is one more alternative:

If you put `extern "C" {}` around each and every header and source that involves function pointers, then everything will be fine, because C++ does the right thing for methods, and everything else will turn C-ish everywhere. At least theoretically.

basvodde:
That will avoid the cast and the typedef?

arstrube:
Yes. Instead, it will obfuscate some 18 files and nobody will know if or when or why what breaks.

Also, Clang hasn't seen it yet. Neither Ms Studio.

arstrube:
You may admire this great work of ingenuity and clean code here: https://github.com/arstrube/cpputest/pull/5.

arstrube:
Typedefs are evil; let's do obfuscation instead :D

arstrube:
I knew it! Travis reveals more obfuscation problems:
```
In file included from tests/CppUTestExt/ExpectedFunctionsListTest.cpp:32:
In file included from tests/CppUTestExt/MockFailureReporterForTest.h:31:
./include/CppUTestExt/MockSupport.h:43:14: error: 'mock' has C-linkage
      specified, but returns user-defined type 'MockSupport &' which is
      incompatible with C [-Werror,-Wreturn-type-c-linkage]
MockSupport& mock(const SimpleString& mockName = "", MockFailureReporter...
             ^
1 error generated.
make[2]: *** [objs/tests/CppUTestExt/ExpectedFunctionsListTest.o] Error 1
make[2]: Leaving directory `/home/travis/build/arstrube/cpputest'
make[1]: *** [extensions] Error 2
make[1]: Leaving directory `/home/travis/build/arstrube/cpputest'
make: *** [check_examples] Error 2


The command "../scripts/travis_ci_build.sh" exited with 2.
```

arstrube:
Yay! We can break much existing test code this way, but we do avoid a typedef.

Never mind, we tell our users they must wrap all their tests in `extern "C"` from now on :))

arstrube:
@basvodde I'm just mocking you, sorry, shouldn't do that. Take one of the alternatives here, or a combination thereof, and all will be good.

basvodde:
Its ok :) I'll accept the mockery, as I've been keeping you very busy with this pointer for weeks now :P

basvodde:
The extern "C" things you did peeked my interest. It is ugly, but I think there must be a better solution than the current one.

Question: The casts in the MockSupport_c.cpp can be done with a normal cast, right? They do *not* require a typedef because is *that* space a function pointer is by default a C++ function pointer. Correct?

arstrube:
Right, but not quite. And, unfortunately, this seems to be largely uncharted territory in C++, so the Cl2000 is currently the only compiler I know that does this cleanly. __There are no casts (within CppUTest), because all function pointers in all declarations are C__.

The basic idea is:
```c++
extern "C" {
    class A {
    public:
         void method_that_takes_an_fp(void(*arg)()); // the pointer will be C
    }
}
```
The cl2000 does that cleanly, in that everything that belongs to C++ classes keeps C++ linkage.

Gcc and other compilers do weird things to varying extent, including giving C++ linkage to member objects in the header, and then giving C linkage to their initialization in the source.

__The basic idea is to wrap all C++ code in `extern "C"` so everything that is not classes gets C linkage, including any function pointer arguments.__

The Bad News is that our most common compilers do really weird things with that. Also, it is obfuscated (nobody will know what you are doing) and a lot of code needs to be wrapped in C (and no normal programmer will guess why).

If you try to wrap individual classes, rather than entire headers, things get very messy very quickly. And remember, there is only one compiler that we no of, that even cares whether a function pointer is C-ish or Cpp-ish.

arstrube:
A cast with a typedef is clear, easy, and clean :-P

arstrube:
You might try:
```
extern "C++" {
#include "MockSupport_c.h"
}
```
in MockSupport_c.cpp. It would make the C interface C++-ish. Then no casts would be needed anywhere. My guess is that it won't link, though.

basvodde:
Why does cl2000 make all function pointers C by default ? I don't understand that part yet...

arstrube:
No no, it doesn't do that. It __distinguishes__ C function pointer versus C++ function pointers. Their C-ishness or Cpp-ishness depends on the context: C++ in *.cpp, C in *.c.

arstrube:
So, in MockSupport_c.h, / .cpp they are C-ish (because of extern "C").

But in MockSuport.cpp (which it interfaces to) they are Cpp-ish.

arstrube:
Loosely speaking, by putting all of MockSupport.cpp in extern "C" they become C-ish there, too. Only, the C-ishness then doesn't stop there....... hence changes to some 18 or so files.

arstrube:
Let's agree on the fact that modern C++ compilers just couldn't care less ;-)

basvodde:
@arstrube Proposed a pull request that should make this one a lot smaller. Correct?

arstrube:
Right. I had similar changes in one PR, only I called the functions "some_function...", "default_function..." and 'expected_function...." which I believe clearer for the tests involving default values. But, otherwise, it should be fine.

basvodde:
Right, sorry about that. Now I recognize this as much nicer than the cast and/or the typedef.

Could you change this PR into what is still needed? Just the typedef/casts in the MockSupport_c right? Let me think a little bit more about how to resolve these, otherwise we'll merge that (finally!)

arstrube:
umm, I already did (rebased & dropped what isn't needed)?

platforms/CCStudio/CppUTestExtRunAllTests1.pjt
```diff
@@ -38,7 +38,6 @@ Source="..\..\src\CppUTestExt\MockExpectedCallsList.cpp"
 Source="..\..\src\CppUTestExt\MockFailure.cpp"
 Source="..\..\src\CppUTestExt\MockNamedValue.cpp"
 Source="..\..\src\CppUTestExt\MockSupport.cpp"
-Source="..\..\src\CppUTestExt\MockSupport_c.cpp"
```
arstrube: This file was still in there although unneeded (MockSupport_c is tested in CppUTestExtRunAllTests2.pjt). I removed it here since we are coming up against memory limitations.

basvodde: Yeah, I didn't do anything with it :)

basvodde:
Ok, its been long enough :) I really can't think of another way as it isn't possible to make a function with both C and C++ style pointers.

Again, thanks for your patience on this one. This might seem like a slow process, but I think it is these in-depth discussions that will keep CppUTest to the right quality level. Thanks!

arstrube:
Cl2000 compiles without complaints now.

PR #896, opened 2016-03-11
Pending fixes for cl2k (Cl2000, c2000): C++ vs. C function pointer problem - version 2

arstrube:
This works.


src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -32,6 +32,16 @@
 #include "CppUTestExt/MockSupport.h"
 #include "CppUTestExt/MockSupport_c.h"
 
+extern "C"
+{
+    typedef void (*cpputest_c_function_pointer)();
+}
+
```
arstrube: I'd prefer something shorter like `typedef void (*fp_c)();` now that it is used only once.

basvodde: I'd prefer not as this is more descriptive and your editor should autocomplete anyway :)

arstrube: hehe

PR #897, opened 2016-03-12
Gcc 4.9 floating point conversion issue (#871).

offa:
See #871

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -235,7 +235,7 @@ void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
 /* MinGw GCC 5.3.0 isnan / isinf macros are not working properly, causing a conversion
  * warning / error
  */
-#if defined(__GNUC__) && __GNUC__ >= 5
+#if __GNUC_PREREQ(4, 9)
 #pragma GCC diagnostic ignored "-Wfloat-conversion"
 #endif
 
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -232,12 +232,18 @@ void (*PlatformSpecificFree)(void* memory) = free;
 void* (*PlatformSpecificMemCpy)(void*, const void*, size_t) = memcpy;
 void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
 
-/* MinGw GCC 5.3.0 isnan / isinf macros are not working properly, causing a conversion
+/* GCC 4.9+ isnan / isinf macros are not working properly, causing a conversion
  * warning / error
  */
-#if __GNUC_PREREQ(4, 9)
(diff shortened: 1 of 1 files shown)


arstrube:
Hmm. Looks like MInGW doesn't know `__GNUC_PREREQ`...

offa:
Maybe it has another name? If i remember correctly, there's such a macro in MinGW.

arstrube:
__MINGW_GNUC_PREREQ :-(

offa:
Thanks, i have added separate checks for Gcc and Mingw now.

offa:
Seems this did more damage than help ...

src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -232,12 +232,18 @@ void (*PlatformSpecificFree)(void* memory) = free;
 void* (*PlatformSpecificMemCpy)(void*, const void*, size_t) = memcpy;
 void* (*PlatformSpecificMemset)(void*, int, size_t) = memset;
 
-/* MinGw GCC 5.3.0 isnan / isinf macros are not working properly, causing a conversion
+/* GCC 4.9+ isnan / isinf macros are not working properly, causing a conversion
  * warning / error
  */
-#if defined(__GNUC__) && __GNUC__ >= 5
+#if ( defined(__GNUC_PREREQ) && __GNUC_PREREQ(4, 9) )
```
arstrube: Maybe check whether it's defined on a separate line? The preprocessor is a bit simple-minded ;-)

arstrube: Perhaps then also move it into CppUTestConfig.h, define CPPUTEST_NEED_CONV_FIX there and just check it here?

arstrube: Actuall, the && forces evaluation of the second term, which results in a syntax error, because the macro is not defined (one of them always isn't). Maybe we actually don't need the MinGW part at all; I think it has behaved itself so far.

arstrube:
I think this should do nicely:
```c++
#if defined(__GNUC_PREREQ)
  #if __GNUC_PREREQ(4, 9)
    #pragma GCC diagnostic ignored "-Wfloat-conversion"
  #endif
#endif
```
If Mingw suffers from the isnan() problem as well, we can do the full shebang in CppUTestConfig.h.

arstrube:
Not so fast.... there's clang, too, and it has __GNUC__ defined, but doesn't know GNUC_MINOR__ so we arrive at this:
```c++
/* GCC 4.9.x introduces -Wfloat-conversion, which causes a warning / error
 * in its own (macro) implementation of isnan() and isinf()
 */
#if defined(__GNUC__) && defined(GNUC_MINOR__)
 #if __GNUC__ >= 5 || ( __GNUC__ == 4 && GNUC_MINOR__ > 8 )
  #pragma GCC diagnostic ignored "-Wfloat-conversion"
 #endif
#endif
```

offa:
via #901

PR #898, opened 2016-03-13
Changed values to real functions to get rid of warnings

basvodde:
Changed the values to functions to avoid casts


arstrube:
This should work for the cl2000. Can test on Monday.

PR #901, opened 2016-03-14
GCC Float conversion fix (#871); separate version checks for Gcc

arstrube:
@bukulin, @offa - this seems to work for all situations.....


offa:
Perfect, thanks. :+1:

PR #902, opened 2016-03-15
Avoid segfaults when checkExpectations is done after teardown for stack allocated parameters

megaloni:
When using stack-allocated parameters when setting up expectedCalls, calling checkExpectations in teardown has the potential to access invalid memory when calling toString() on the MockNamedValue that was stack-allocated (for example, when the expected call did not occur).

To get around this, cache the toString() result when the parameter value is set to avoid having to access it after it has gone out of scope.


arstrube:
A test case would be nice, to demonstrate the problem & validate the fix.

src/CppUTestExt/MockNamedValue.cpp
```diff
@@ -115,6 +115,8 @@ void MockNamedValue::setObjectPointer(const SimpleString& type, const void* obje
     {
         comparator_ = defaultRepository_->getComparatorForType(type);
         copier_ = defaultRepository_->getCopierForType(type);
+        if(comparator_ && copier_)
```
arstrube: What is the significance of the copier here? Does it cause the same problem? What if I have a comparator, but no copier? Or a copier, and no comparator?

arstrube:
Any progress on a test case?

megaloni:
I got pulled off on another task, but I will get one written up and will also circle back on why I added the check for _copier.

arstrube:
'cause I tried to come up with a test case, and it crashed, but for the wrong reasons. I couldn't make it crash with stack-allocated data gone out of scope, or with dynamically allocated data that was freed.

The reason for the crashes was, essentially, absence of a comparator (another issue, actually, because it would be nicer if that didn't crash), which happens regardless of caching. When I fixed that, I couldn't make it crash anymore.

So I am curious about your use case.

basvodde:
I'm a bit confused by the code and definitively not like the cached string in the mock value. Would love to hear more, so I can understand it :)

megaloni:
I need to do more investigation on this - we may be running into this issue only when dealing with smart pointers.

PR #904, opened 2016-03-17
Fix SimpleString Tests for 16bit per byte systems (cl2000)

arstrube:
Forgot to fix these tests -- some of them were still failing on the cl2000.

### DIFF
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -917,8 +917,18 @@ TEST(SimpleString, MemCmpFirstLastNotMatching)
     CHECK(0 != SimpleString::MemCmp(base, lastNotMatching, sizeof(base)));
 }
 
-
-TEST(SimpleString, MaskedBits1byte)
+#if (CPPUTEST_CHAR_BIT == 16)
+TEST(SimpleString, MaskedBitsChar)
+{
+    STRCMP_EQUAL("xxxxxxxx xxxxxxxx", StringFromMaskedBits(0x00, 0x00, 1).asCharString());
+    STRCMP_EQUAL("xxxxxxxx 00000000", StringFromMaskedBits(0x00, 0xFF, 1).asCharString());
+    STRCMP_EQUAL("xxxxxxxx 11111111", StringFromMaskedBits(0xFF, 0xFF, 1).asCharString());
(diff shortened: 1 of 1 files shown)


basvodde:
Coverage decreased by adding a test!

arstrube:
Yeah, but it decreased by 0,0% :-P

basvodde:
So what should I do?

arstrube:
You could merge it. Or I could "fix" it :-). There are 92 hits less, in some 11.000, decreasing coverage from 100% to 100%.

arstrube:
Actually, I have no idea how to fix it. The #if - #else'd test (the part that's compiled) is the same as before. The other tests are just doing a calculation differently, which ends up with the same value as before on this platform.

arstrube:
I take this to mean that I can point the coverage badge at cpputest.github.io to cpplutest/cpputest now ;-).

arstrube:
A person not providing tests will get a red flag from now on, it seems.....

arstrube:
@basvodde 

1. How did you include the Coveralls check in the build? I see no commit. But I also don't see how to configure this from Github or Coveralls. The check is not active on my fork.

2. Is there a way to configure the Coveralls check? I think even the smallest change in percentages is significant. However, I don't think decreased line counts need to be taken into account. In the extreme, decreased line count could mean more efficient code... What makes it worse is that Coveralls doesn't give any indication in the code (at least I couldn't find any) that tells us exactly which lines were executed fewer times, and it would be potentially bothersome to fix this by means of guesswork.

basvodde:
1) No idea, I just did what you told me last time :)

2) No idea either :)

Previously the @coveralls github account answered you, perhaps it can do that again :P

arstrube:
There seems to be an open issue against Coveralls regarding 2). So I think it will be addressed in due time :-)

As for 1), it turned out to be surprisingly simple. It's a new feature, and you got it, because you linked to Coveralls fairly recently. After I removed Coveralls from my fork and added it again, I am now getting the same feedback. Only on pull requests though, so in order to see it, I must create a PR against my own fork.

basvodde:
So, nothing I should do?

arstrube:
I guess not. Except accepting the occasional PR that 'decreases' coverage by 0.00000% ;-)

PR #906, opened 2016-03-18
Interface for checking an IEEE754 exception flag in the debugger

arstrube:
When a test fails due to an an IEEE754 exception flag being set,
it can be hard to find exactly where the error occurred.
By entering the new static method in the debugger as a watch,
e.g. IEEE754ExceptionsPlugin::checkIeee754ExeptionFlag(4), its
value will change to true at the exact point where the problem
occurred.

In most cases, it won't then be necessary to enable floating point
trapping in order to find the offending line (the other PR that
mysteriously fails under Cygwin on Appveyor).

### DIFF
--- a/include/CppUTestExt/IEEE754ExceptionsPlugin.h
+++ b/include/CppUTestExt/IEEE754ExceptionsPlugin.h
@@ -40,6 +40,7 @@ public:
 
     static void disableInexact(void);
     static void enableInexact(void);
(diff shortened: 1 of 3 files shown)


arstrube:
@basvodde - no coverage check??

arstrube:
ahh, there it is. An increase. Nice :-)

PR #907, opened 2016-03-20
argument line option to enable OPTRUN_TEST

cuitoldfish:
this idea is originated from @PaulBussmann, I have this try since Paul seem busy. 

After discuss, it is more reasonable to add one command line to let the `OPTRUN_TEST` get run.

I  currently prefer `OPTRUN_TEST` rather than `OPTIN_TEST` because I think there should only be two possibility, get run then run counted, or get ignored then ignore counted. so, optional run by specifying the command line option `-optRun`

### DIFF
--- a/include/CppUTest/CommandLineArguments.h
+++ b/include/CppUTest/CommandLineArguments.h
@@ -45,6 +45,7 @@ public:
     bool isColor() const;
     bool isListingTestGroupNames() const;
     bool isListingTestGroupAndCaseNames() const;
+	bool isOptRun() const;
     int getRepeatCount() const;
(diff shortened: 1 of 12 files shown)


basvodde:
I think I mentioned that I would prefer to just use IGNORE_TEST and give an ability to run IGNORE_TESTS ?

That way we do not need to make another macro and it will keep it simpler.

cuitoldfish:
ok, seems i have a misunderstanding. I agree you and would open another PR

basvodde:
Yup, so what is needed is a --run-ignored. If you can combine it with -g and -n, we'll be able to easily run one test that was ignored.

PR #908, opened 2016-03-20
add command line option to Run ignore test

cuitoldfish:
using "-runIgnore" to run IGNORE_TEST


include/CppUTest/CommandLineArguments.h
```diff
@@ -45,6 +45,7 @@ class CommandLineArguments
     bool isColor() const;
     bool isListingTestGroupNames() const;
     bool isListingTestGroupAndCaseNames() const;
+	bool isRunIgnore() const;
```
basvodde: Hey, whats with the spacing? :)

basvodde:
I'd suggest to make the option -i (or so) and make the longer option --ignore (with double-dash). That is kinda the practice in unix apps. Or just have the -i option.

arstrube:
Hey @cuitoldfish you spooked the Dos build :D. One of its 6 part-exes ran out of memory linking. @basvodde Guess we'll just have to drop the tests, hehe ;-)

basvodde:
7th part :)

cuitoldfish:
hey, only "-i" option can describe clearly?

what's the failure? I have slow network and could not response to those failure immediately. what's does this "ran out of memory linking" mean? I didn't do any thing to consume the size.

arstrube:
@cuitoldfish well you didnt. but your code did ;-).
@basvodde - that means lots of shunting around :-/ but probably the best thing to do so it will take some time before we run out of space again.....

cuitoldfish:
@arstrube, so seems I can not pass dos build unless you drop those tests files? :)

@basvodde, I prefer "-ri" if clean is the rule for command line option

basvodde:
-ri will be fine. Not sure why the r ?

arstrube:
@cuitoldfish yeah, unless you feel like fiddling with `platforms\Dos\sources.mk` yourself a little ;-)

cuitoldfish:
—ri is the abbreviation of run ignore, i think it is easier to remember

basvodde:
Ah run :) Ok

basvodde:
The rest looks good. But would appreciate fixing the formatting. If you look at the pull request diff, you can see it looks kinda ugly now :(

cuitoldfish:
Oh, I forget to mention that formatting error. That is because wrong configuration in my editor. I have corrected those. Will commit soon

include/CppUTest/CommandLineArguments.h
```diff
@@ -69,6 +70,7 @@ class CommandLineArguments
     bool runTestsAsSeperateProcess_;
     bool listTestGroupNames_;
     bool listTestGroupAndCaseNames_;
+	bool runIgnore_;
```
basvodde: Hum hum :)

basvodde:
Another simplification. I'm a bit annoyed by the "setRunIgnored" flag, but haven't figured out yet how to get rid of that. But at least, you can remove the _runIgnored member field out of the UtestShell and only have it in the IgnoredShell...

Then you can also remove the isIgnored from the UtestShell interface.

cuitoldfish:
the current implementation in TestRegistry program to the UtestShell interface, if isRunIgnored only exist IgnoredShell, then how to call it from TestRegistry?

basvodde:
:)

It isn't called from TestRegistry as far as I can see :)

include/CppUTest/CommandLineArguments.h
cuitoldfish: yeah, I'm sorry for this missing one, thanks for your reminder!

src/CppUTest/TestRegistry.cpp
```diff
@@ -50,6 +50,7 @@ void TestRegistry::runAllTests(TestResult& result)
     result.testsStarted();
     for (UtestShell *test = tests_; test != NULL; test = test->getNext()) {
```
cuitoldfish: @basvodde , here we can see, in the TestRegistry, all test shell is taken to process one by one based on the UtestShell interface. so i need to add setRunIgnore in the UtestShell interface, or I have to segregate the UtestShell interface to get a few new classification

arstrube: @basvodde it is in fact a new interface we are adding to UtestShell, and that's why in my opinion it needs to be there (like @cuitoldfish pointed out). It is the same as runInSeparateProcess_ or currentRepetition_. We need to be able to tell UtestShell from the "outside world" how it is supposed to behave.

arstrube:
I was just wondering what the output will be of an ignored test that is run.

Currently, the output would be"
```bash
.!                        # normal test followed by ignored test
TEST(group, test1)        # normal test with -v
IGNORE_TEST(group, test2) # ignored test with -v
```
My expectation is that this, when ignored tests are run, changes to:
```bash
..                        # normal test followed by ignored test that is run
TEST(group, test1)        # normal test with -v
TEST(group, test2) # ignored test that is run with -v
```
Is this actually the case? Does everyone agree that this is how it should be? In my opinion, one should be able to tell from the output at a glance whether an ignored test was run or not.

Also, statistics in the summary line should change accordingly, because the ignored tests which are run are no longer ignored. Ignored tests which are not run should stay as is.

arstrube:
I see there is a test for the summary counts. But there isn't one for the output.

cuitoldfish:
i prefer 
```.cpp
..                       # normal test followed by ignored test
TEST(group, test1)        # normal test with -v
IGNORE_TEST(group, test2) # ignored test with -v
```

because user know he is running the IGNORE_TEST, the output should tell this truth.

The statistics has been changed accordingly, when IGNORE_TEST run, ignore count will not increase but the run count will increase

arstrube:
@cuitoldfish hmmm. I think replacing ! by . but not replacing the -v output is inconsistent (if that's really what you meant)

arstrube:
Another, rather small thing -- grammatically speaking, I would prefer runIgnore**d**Something in the code to runIgnoreSomethin`.

cuitoldfish:
I actually used the runIgnore**d**, but later change my mind because I think I'm run the IGNORE TEST, not sure which will be the better name

arstrube:
I was wondering about that (IGNORE TEST)myself, then resolved that
IGNORE TEST as in "ignore! test" (imperative) but
runIgnoredTest as in "run this ignored test", because else it would turn out like "run! ignore! test" which is kind of a contradiction in itself.....
Maybe @basvodde has a preference one way or the other.

src/CppUTest/TestRegistry.cpp
basvodde: The isRunIgnore is not called everywhere outside the class, so adding it is not useful. In fact, it shouldn't have been added if the code was test-driven....

Also the member varialble is never used and can be moved down in the class hierarchy. I recommend to do that as it isn't needed to have that state in the UtestShell.

I wish we could move the setRunIgnored also out of the UTestShell interface, but I wouldn't know how. The reason is that usually this kind of information is given at object construction time, but the UtestShell is constructed at program start, so that isn't possible. So, I guess we'll need to keep that *one* method in the UtestShell interface.

basvodde:
You guys are funny :)

I wouldn't care much, but remember that if we change the output, we also need to change it as input as the -v format can be used as input for CppUtest also.

src/CppUTest/TestRegistry.cpp
cuitoldfish: yeah I agree. 
`UtestShell::isRunIgnore()` is not called outside the class, I will remove it. 
For the `setRunIgnored ` and that member variable, I also think defined in UtestShell is not the best design, but due to current class hierarchy(UtestShell and IgnoreTestShell) and inter-class dependency(TestRegistry and UtestShell Interface), we can not get bettern place to hold the `setRunIgnored ` and the member variable, unless we do some refactorying on the class hierrachy, eg. segregate the UtestShell interface. But currently it's not deserved to refactorying the interface. We can do this as the interface beome fatter.

cuitoldfish:
if this, I think we need to keep "IGNORE_TEST" in the output so that the output can be used as input string

arstrube:
Does it make any difference for the input?

arstrube:
I mean, if I ask CppUTest to "run" an ignored test case, it will know internally that it is ignored, right?

arstrube:
@basvodde pursue a career as comedians, perhaps? Double-crossed C and The Code Carmeleons"s, haha :)

arstrube:
@basvodde There is no need to change input, as can be seen here:
```bash
$ ./CppUTestTests.exe -v "IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture)"
IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture) - 0 ms

OK (613 tests, 0 ran, 0 checks, 1 ignored, 612 filtered out, 0 ms)

$ ./CppUTestTests.exe -v "TEST(UtestShell, IgnoreTestAccessingFixture)"
IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture) - 0 ms

OK (613 tests, 0 ran, 0 checks, 1 ignored, 612 filtered out, 0 ms)
```
Adding the -ri flag would run the test. No difference there.

cuitoldfish:
@arstrube if I have a ignore test

IGNORE_TEST(group, test2) 

then if I run cpputest using this command `./CppUTestTests.exe -v "IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture)"`

your expectation is 
```.cpp
TEST(group, test2) - 0 ms
OK (613 tests, 1 ran, 0 checks, 0 ignored, 612 filtered out, 0 ms)
```
seems strange incongruity :)

arstrube:
Yeah, but this "feature" is not part of the requirements that we agreed on :P The requirements say that only if you do
`./CppUTestTests.exe -ri -v "IGNORE_TEST(UtestShell, IgnoreTestAccessingFixture)"` you can expect 
```bash
TEST(group, test2) - 0 ms
OK (613 tests, 1 ran, 0 checks, 0 ignored, 612 filtered out, 0 ms)
```

arstrube:
It may not be what a user might expect, but it is consistent with how we said this thing should work :). I mean, if I specify the same test case by means of -sg / -sn, it isn't going to be run, either.

arstrube:
Did you merge cpputest/master / rebase to cpputest/master? It's still spooking the Dos build ;-)

arstrube:
@basvodde if I recall correctly, TestUTestMacro is rather large also. It may need breaking up; or perhaps @cuitoldfish can reduce the size of his tests by 27 bytes :P....

src/CppUTest/TestRegistry.cpp
basvodde: Can you make setRunIgnored virtual. Let the UtestShell implement return true and move the member variable in the subclass.

You do not need the member variable in the base class and removing that makes it simpler.

basvodde:
@arstrube Ok, I'll look at splitting some other time.

What I meant with the output is that is you change it to IGNORE_TEST_RUN_AWAY(gorup, name) then you also need to add that as possible input.

src/CppUTest/TestRegistry.cpp
cuitoldfish: yeah, I also plan to do that. and what's your thinking of interface segregate for UtestShell? maybe We need to do it in case UtestShell interface become fatter.

basvodde: Yeah, I don't like the setIgnoreTest in the interface, but for now I think it is ok.

As I mentioned, usually this kind of info is passed on creation, but in our case we cannot do that because the creation happens in the macro. So, for now its ok.

cuitoldfish:
@arstrube , I haven't rebase or merge master, I would like to finalize current implementation and then do that to pass the CI.  But did you mention currently there is still issue in TestUTestMacro  to hamper the CI?

arstrube:
@cuitoldfish maybe better if you merge. If the Dos problem goes away, we're good. @basvodde If not, this won't pass before the UtestMacroTests are split in some way.

arstrube:
@cuitoldfish why, you removed a method didn't you. But the code grew fatter and coverage went down :-)

cuitoldfish:
I removed IgnoredUtestShell::isRunIgnore() because it's not needed as discussed

cuitoldfish:
The decrease may because the virtual method void UtestShell::setRunIgnore(). This is just a interface, it's body will not be executed so that coverage decrease.

arstrube:
You'll need to test it. So it doesn't do anything fishy :-P. Or at least enclose it in
```c++
// LCOV_EXCL_START
// LCOV_EXCL_STOP
```
Of course you had to remove  IgnoredUtestShell::isRunIgnore(). I was just joking about removing a method and still the code size in Dos grew by some 1.8 kbytes :)

And, may I point out that the class is  Ignore**d**UtestShell and not IgnoreUtestShell ;-). So, personally, I'd like the method and variable names to be consistent with that.

arstrube:
No problem about passing Dos later, but I'd say the 1.8k problem is different to the original one and only splitting up the UtestMacro tests will make it go away. Not necessarily the files, but the test groups. Because one test group has hit the 64k limit.

cuitoldfish:
I haven't notice the extra `d` in IgnoredUtestShell :),   I will add the `d` consistently.

basvodde:
Excellent, thanks!

arstrube:
@cuitoldfish nice :)
@basvodde now we need to to split up TestUTestMacro so the Dos build won't fail...

arstrube:
Actually, I'm not entirely clear what the linker means complaining about `TestUTestMacro_TEXT2`. Looks like it corresponds to the file name, not the test group. So we may have to split the file after all :-(.

cuitoldfish:
I think at least from test design point of view, there should be separate test file for UtestShell and IgnoreTestShell, right?

arstrube:
Sounds good to me.

arstrube:
After looking at it more closely, there were some IgnoredUtestShell tests in there already that conceptionally belong to UtestTest.cpp. Along with your new tests, I will move them there. It is a rather small file. So we don't have to go to the trouble to split files at this point.

PR #910, opened 2016-03-21
Match Ieee754Exceptions tests to CppUTest coding style

arstrube:


basvodde:
What happened to cygwin?

arstrube:
It is acting up :-(. Very strange. Complaining about /usr/bin/bash, which should be there (at least, it used to be there).

arstrube:
Or, rather autoreconf conced out somehow, and then there's no ../configure...

PR #911, opened 2016-03-21
Reshuffle Dos tests into 7 EXEs

arstrube:
Voila.

(That Makefile really is a work of Genius) ;-)


basvodde:
Ah, and the failure is still the cygwin one....

PR #913, opened 2016-03-24
Move IgnoredUtestShell tests to UtestTest.cpp

arstrube:
Currently openwatcom.org is not responding (DoS attack?). So I'm not 100% sure the Dos build will work, however, the problem did go away when compiling with this change locally.


arstrube:
@Andne Cygwin on Appveyor is giving us grief. It's conking out right when attempting to run autoreconf -i .. . Perhaps you have an idea what might be the problem? Nobody else appears to have this problem (Google), and it showed up very suddenly, literally from one day to the next.

good: https://ci.appveyor.com/project/basvodde/cpputest/build/3.7.0-ci702/job/2ytmx0j9qmaf8lw3
bad: https://ci.appveyor.com/project/basvodde/cpputest/build/3.7.0-ci703/job/r4um7x8emtgbesys

I must admit I have no idea what might be going on....

arstrube:
@basvodde - finally openwatcom is back & I restarted the job to see it pass on my fork :-)

basvodde:
Bah, now it passed but due to the white spaces it can't be automatically merged :(

arstrube:
how annoying.

arstrube:
DON'T MERGE PLEASE - the merge went all wrong!

basvodde:
Uh, ok, I'll merge it :)

arstrube:
Okay. Now it's the way it was supposed to be.

Andne:
I think I've gotten it working again by forcing cygwin to upgrade it's installed packages.  Now I need to figure out if that was really the fix or what and back out the stuff that didn't fix it (but causes other problems).  Looks like autoreconf was somehow performing an illegal operation (exit code 143, SIGTERM it seems?).

Andne:
Found this, it took effect on the date things broke - https://www.appveyor.com/updates/2016/03/08

Guess the new image has something wrong in the cygwin install?

Andne:
Well, forcing the image back to Visual Studio 2013 seems to fix the problem.  It's not like we need VS2015 yet anyways, so not a big deal.

arstrube:
Thanks @Andne, much appreciated. I suspected something like that, but it wasn't at all obvious...

Andne:
Gah, had nothing to do with the images, it was because I was running cygwin install to make sure that the needed packages are available (they are by default too I've since learned) and the setup command isn't blocking.  End result, a build was using cygwin while the installer was updating cygwin - not very good.  My cleanup PR #917 doesn't try to do the setup anymore, so that's why it works.

PR #917, opened 2016-03-25
Appveyor cleanup

Andne:
Includes a fix to keep Cygwin working for now.  For some reason it doesn't work on the VS2015 image correctly so can't build with the latest version of visual studio (not that such a configuration was in use anyways).

Also includes a fair amount of cleanup and restructuring of the appveyor scripts to try and reduce the amount of duplication in them and have the variables be more consistent across the different configurations.  Basically, MinGW and Cygwin are now seen as platforms the same as Win32 or x64 is.  PlatformToolset is only to determine the version of the compiler to use.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ environment:
     PlatformToolset: v140
 
 install:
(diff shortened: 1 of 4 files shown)


arstrube:
@Andne thanks a lot!

PR #920, opened 2016-03-28
Fix CMake absolute paths when CppUTest is nested.

charlesnicholson:
A previous 'fix' of mine has CMake hard-coding CppUTest test executables to "${CMAKE_BINARY_DIR}/tests". That's ok when CppUTest is built as the CMake root, but when CppUTest is a component in a more complex project, that path is no longer correct.

This PR changes `cpputest_set_test_output_location` to `cpputest_normalize_test_output_location`, and no longer takes an output path. Instead, the new function simply pushes ${CMAKE_CURRENT_BINARY_DIR} as the build target for all configuration variants, which was the intent of my original clumsy PR.

This is still a WIP so that @jerryryle can see it, please don't merge yet!


jerryryle:
This looks good. I tested with my reproduction case and it worked as expected.

PR #923, opened 2016-04-06
CMake provide cpputest_buildtime_discover_tests to clients of the CppUTest libs

asgeroverby:
There are two part to this pull request

1) Current cmake builds of CppUTest use the function cpputest_buildtime_discover_tests. This pull request makes the  cpputest_buildtime_discover_tests available to projects using CppUTest as an installed lib.

2) Now all include files under include/CppUTest and include/CppUTestExt are installed.
The include paths are also automatically add to the include paths of the lib clients.
Before SimpleMutex.h was not in the install list of include/CppUTest.
Before MockSupport.h was included twice in the install list of include/CppUTestExt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,10 @@ if(PkgHelpers_AVAILABLE)
     DESTINATION ${LIB_INSTALL_DIR}/CppUTest/cmake )
   install(EXPORT CppUTestTargets
     DESTINATION ${LIB_INSTALL_DIR}/CppUTest/cmake)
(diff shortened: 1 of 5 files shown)


asgeroverby:
The Travis CI builds fails because  target_include_directories was introduced in CMake version 2.8.11 and the project has:
cmake_minimum_required(VERSION 2.8.7)

Is  it OK to increase the minimum version?

asgeroverby:
Found an issue when building with MSVC. 
Closing this request and splitting it in two.

PR #924, opened 2016-04-06
Cmake discover tests

asgeroverby:
Provide build-time discover tests to clients of the CppUTest libs when the libs are build and installed using cmake. i.e. cpputest_buildtime_discover_tests 

Basically it lets clients do like the lib it self i.e.

```
add_executable(myTests ${myTests_src})
target_link_libraries(myTests CppUTest ${THREAD_LIB})
...
cpputest_buildtime_discover_tests(myTests)

```


asgeroverby:
This PR is obsolete. CppUTest has moved on to cmake_minimum_required(VERSION 3.1) and the cmake usages has been modernized.  
PR #1220 gives the same benefits against current CppUTest.

Issue #916, opened 2016-03-24
Compiler warning Visual Studio 2005 & 2013 


BillDenton:
Compiler warning:
 ... \src\CppUTestExt\IEEE754ExceptionsPlugin.cpp(74) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
This looks similar to https://github.com/cpputest/cpputest/issues/198

arstrube:
Hi there,

Shouldn't this show up in Appveyor when building Head? I see a bunch of other warnings (also interesting, actually) but not this one?

arstrube:
Do you have a better suggestions to fix this than an explicit cast to bool? Would an explicit cast make this warning go away?

arstrube:
It seems the recommended way is 

return (fetestexcept(flag) != 0);

They say that will make the compiler happy.

BillDenton:
It does make the Visual Studio 2013 compiler happy. Not checked if the CppUTest internal tests pass and I don't we are using this in our unit test. So don't know if it is correct.

arstrube:
It is probably the most correct way on the planet to get 'real' bool out of an int return value :)

arstrube:
P.S. There are tests in the test suite that would fail if it wasn't correct. You can check by running the tests with -v -gFE_ to see if they run in VS. I believe they do.

arstrube:
@basvodde the Git installer reset my system config (a nasty habit; it doesn't take the existing one but forces you to make some decisions on the fly and replaces the whole thing, it had autocrlf=true instead of false, but changing that still doesn't restore my previous settings. Which are the correct settings for CppUTest?

arstrube:
It keeps on doing:
```bash
warning: LF will be replaced by CRLF in src/CppUTestExt/IEEE754ExceptionsPlugin.cpp.
```
:-(

Andne:
The .gitattributes file is supposed to override the system setting I think, that's why it got added at one point.  Otherwise, under windows I think I just have the default settings (autocrlf=true I believe) and that seems to work correctly most of the time.

Maybe you just need to force a clean checkout of the files?  Sounds like it's trying to commit without conversion a file that has CRLF but maybe a clean checkout would set the file to LF instead?  Also, git settings are screwy with where they can be overridden, so may need to checkout how many places they've been set when inside the cpputest repository.

arstrube:
[...]

PR #925, opened 2016-04-06
Msvc fix warnings

asgeroverby:
Fix two warnings that breaks MSVC builds.

### DIFF
--- a/tests/CppUTestExt/MockSupport_cTestCFile.c
+++ b/tests/CppUTestExt/MockSupport_cTestCFile.c
@@ -56,7 +56,7 @@ void all_mock_support_c_calls(void)
             withStringParameters("string", "string")->
             withPointerParameters("pointer", (void*) 1)->
             withConstPointerParameters("constpointer", (const void*) 1)->
-            withFunctionPointerParameters("functionpointer", (void(*)()) 1)->
+            withFunctionPointerParameters("functionpointer", (void(*)(void)) 1)->
             withMemoryBufferParameter("name", (void*) 1, 0UL)->
             ignoreOtherParameters();
 
@@ -68,7 +68,7 @@ void all_mock_support_c_calls(void)
             withStringParameters("string", "string")->
             withPointerParameters("pointer", (void*) 1)->
             withConstPointerParameters("constpointer", (const void*) 1)->
-            withFunctionPointerParameters("functionpointer", (void(*)()) 1)->
+            withFunctionPointerParameters("functionpointer", (void(*)(void)) 1)->
             withMemoryBufferParameter("name", (void*) 1, 0UL)->
             hasReturnValue();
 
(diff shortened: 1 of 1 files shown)


src/CppUTestExt/IEEE754ExceptionsPlugin.cpp
```diff
@@ -71,7 +71,7 @@ void IEEE754ExceptionsPlugin::enableInexact()
 
 bool IEEE754ExceptionsPlugin::checkIeee754ExeptionFlag(int flag)
 {
-    return fetestexcept(flag);
+    return !! fetestexcept(flag);
```
basvodde: not not ?

arstrube: Apparently an interesting, different fix to the bool problem ^^. This pull doesn't seem to be based on the latest master though, or this change would have been unnecessary. Shouldn't there be a merge conflict?

arstrube: Ummm, apparently not.... where did my fix go?

basvodde: I'd prefer a different solution than not not, which feels a bit obscure.

arstrube: It would appear that I simply forgot to make the pull request to fix #916 :-(. It was supposed to be
```c++
return (fetestexcept(flag) != 0);
```

arstrube: This is the recommended way, see discussion for #916.

basvodde: Ok can either you make that pull request or @asgeroverby change the pull request

arstrube: Was going to make PR but my git global setup is messed up and commit CRLF to the repo :-(. Had to play around with Cygwin and it is all messed up, bash, git and all. Trying to fix asap....

asgeroverby: I will drop 312e5d3 from the pull request - so there is no merge conflict.

arstrube:
@asgeroverby as a matter of interest - I'd like to find out why these warning seemingly did not cause the Appveyor build to fail, yet they broke your build. It looks like the sort of thing that the Appveyor build could / should have caught? @basvodde what do you think?

basvodde:
@arstrube Do we run VC with maximum warnings in AppVeyor?

arstrube:
@basvodde I don't know for sure - probably not with warnings as errors? What I don't understand is, it should work the same whether on Appveyor or on a user's machine, shouldn't it?

arstrube:
@basvodde I strongly suspect we haven't enabled warnings as errors, do projects can break that do.

asgeroverby:
@arstrube, @basvodde my build is most probably different. I never edit MSVC projects by hand if I can be avoided, I use CMake doing something like this:

~~~~~~~~~~~~~~~~~~.{cmd}
> mkdir C:\Work\builds\build-msvc-cpputest
> pushd C:\Work\builds\build-msvc-cpputest
> cmake -DCMAKE_INSTALL_PREFIX=c:/Work/local-install/msvc-cpputest ../../cpputest
-- Building for: Visual Studio 14 2015
...
(Lots of output)
...
-- Build files have been written to: C:/Work/builds/build-msvc-cpputest
> cmake --build . --target install
~~~~~~~~~~~~~~~~~~

I found an Appveyor configuration to do something [similar](https://github.com/jibsen/brieflz/blob/master/appveyor.yml)

PR #927, opened 2016-04-06
Cmake auto set include path

asgeroverby:
Provide automatic includes for lib clients that link with the CppUTest libs.

Below is a minimal `CMakeLists.txt`

``` .cmake
cmake_minimum_required(VERSION 2.8.11)
project(mytest)

include(CTest)
enable_testing()

# Provide access to installed CppUTest
# if installed in non-standard place add "HINTS <path-to-installed-cpputest-lib>"
FIND_PACKAGE(CppUTest)

add_executable(mytest mytest.cpp)
target_link_libraries(mytest CppUTest)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: cpp
 sudo: false
+addons:
+  apt:
+    sources: ['kubuntu-backports']
+    packages: ['cmake']
 matrix:
     include:
     - compiler: clang
@@ -38,10 +42,6 @@ matrix:
           - dosbox
 global:
 - os: linux
-  addons:
-    apt:
-      sources: ['kalakris-cmake']
-      packages: ['cmake']
 - rvm: '1.9.3'
 - secure: |-
     P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ matrix:
           - valgrind
     - compiler: clang
       env: BUILD=cmake
+      addons:
+        apt:
+          packages:
+          - cmake
     - compiler: clang
       env: BUILD=autotools_gtest
     - compiler: clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,6 @@ matrix:
           - valgrind
     - compiler: clang
       env: BUILD=cmake
-      addons:
-        apt:
-          packages:
-          - cmake
     - compiler: clang
       env: BUILD=autotools_gtest
     - compiler: clang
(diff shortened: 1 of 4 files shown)


asgeroverby:
The build fails because Travis-CI defaults to CMake version 2.8.7 and the commit relies on `target_include_directories` which was included in CMake version 2.8.11. c.f. https://community.kde.org/Frameworks/Epics/CMake_target_usage_requirements.  I assumed that the feature was safe to use because the existing `CMakeLists.txt` files were already using `target_link_libraries`, which was also introduced in CMake version 2.8.11.

c.f. fx. `src/CppUTest/CMakeLists.txt`
 ```{.cmake}
if (WIN32)
      target_link_libraries(CppUTest winmm.lib)
endif (WIN32)
```
The above is hidden from Travis-CI because it is only seen in WIN32 mode.

Is CppUTest required to support CMake from version2.8.7? if so I'll try to come up with a workaround.
I have not found a nice way to make Travis-CI use a newer version of CMake.

**Comment by arstrube** [comment]
Hi, have you tried
```yml
      addons:
        apt:
          packages:
          - dosbox
          - cmake
```
in .travis.yml? I haven't tried, but I think it might get you a later version.

asgeroverby:
@arstrube: Your appeared to have no effect. I, however, did a few experiments.

Minimalist examples to test the Tracis-CI CMake version:

- CMake 2.8.12 from kubuntu-backports container-based infrastructure [.travis.yml.container-based](https://github.com/asgeroverby/travis-cmake/blob/2159d5129b27d688d061d6fe350ce7d34bea66b4/.travis.yml.container-based)
  I was only able to get the effect when putting the `addons` directive at the top-level.

- CMake 2.8.12 from kubuntu-backports on standard infrastructure [.travis.yml.standard](https://github.com/asgeroverby/travis-cmake/blob/5dc830d45576024313a02c38bfbfa77e4c1cc698/.travis.yml.standard)

- Standard CMake=3.2.2 on trusty standard infrastructure [.travis.yml](https://github.com/asgeroverby/travis-cmake/blob/bf34b3aa41d0f716bce47ea560527d23c8d5d8d0/.travis.yml)

- Was unable to run trusty container-based infrastructure (maybe it does not exist?)
  (https://travis-ci.org/asgeroverby/travis-cmake/builds/121403436)

The current PR uses the first approach, but it fails because the included gtest/gmock version does not provide CMake install directives for the libs and headers.

gtest/gmock added install support only last year and added the `target_include_directories` in this commit  [https://github.com/google/googletest/commit/f601ee162e23eb3fe49446d13ccd6d30ca687218]. The commit only exposes the directive to cmake version that support it. If it do something similar the PR will pass the travis-ci builds with its current cmake version, but it will break for people trying to with a CMake >= 2.8.11 and the GTEST option enabled. Pulling a fresh (not released - yet) gtest should however fix the problem.

What do you prefer?

arstrube:
Any comment from @basvodde or anyone who routinely uses Cmake?

basvodde:
I'm not understanding the drawbacks and benefits well yet, so it is hard to decide. However, the gtest/gmock should also work with the old versions, so pulling the latest (not released) version seems to not be an option. What other options are there and what are the benefits/drawbacks?

asgeroverby:
CppUTest has moved on to `cmake_minimum_required(VERSION 3.1)`. 
This PR is obsolete. PR #1220 gives the same benefits against current CppUTest.

PR #929, opened 2016-04-07
Return proper bool

arstrube:


arstrube:
@basvodde this is needed to fix the Msvc errors (#925) as @asgeroverby removed this part from #925.

PR #932, opened 2016-04-11
Add mock support for bool (C++ only).

uecasm:
Mock support for bool in C++ only, split from #847.

Note that since the C API does not support bool at all in this version, I have made one additional change: bool parameters and return values are now implicitly convertible to int parameters and return values respectively, and vice versa.  (This only applies to int, not other types such as long or unsigned int.)

This has a consequence that even C++-only tests will pass where the expectation is for a bool and the actual call is an int (or the reverse) -- opinions will probably differ on whether this is a benefit or a drawback, but it was something that I was trying to avoid in the prior PR.

However the main reason this has been done is specifically to allow the case where a C++ test is expecting a bool parameter/return but the actual call is made via the C API, which can only express this as an int.  (There is however no unit test for this as the current tests don't verify compatibility between C and C++ APIs in this fashion.)

### DIFF
--- a/src/CppUTestExt/MockNamedValue.cpp
+++ b/src/CppUTestExt/MockNamedValue.cpp
@@ -46,6 +46,12 @@ MockNamedValue::~MockNamedValue()
 {
 }
 
+void MockNamedValue::setValue(bool value)
+{
+    type_ = "bool";
+    value_.boolValue_ = value;
+}
+
 void MockNamedValue::setValue(unsigned int value)
 {
     type_ = "unsigned int";
@@ -136,6 +142,12 @@ SimpleString MockNamedValue::getName() const
 SimpleString MockNamedValue::getType() const
 {
     return type_;
+}
+
+bool MockNamedValue::getBoolValue() const
+{
+    STRCMP_EQUAL("bool", type_.asCharString());
+    return value_.boolValue_;
 }
 
 unsigned int MockNamedValue::getUnsignedIntValue() const
@@ -268,7 +280,9 @@ bool MockNamedValue::equals(const MockNamedValue& p) const
 
     if (type_ != p.type_) return false;
 
-    if (type_ == "int")
+    if (type_ == "bool")
+        return value_.boolValue_ == p.value_.boolValue_;
+    else if (type_ == "int")
         return value_.intValue_ == p.value_.intValue_;
     else if (type_ == "unsigned int")
         return value_.unsignedIntValue_ == p.value_.unsignedIntValue_;
@@ -312,7 +326,9 @@ bool MockNamedValue::compatibleForCopying(const MockNamedValue& p) const
 
 SimpleString MockNamedValue::toString() const
 {
-    if (type_ == "int")
+    if (type_ == "bool")
+        return StringFrom(value_.boolValue_);
+    else if (type_ == "int")
         return StringFrom(value_.intValue_);
     else if (type_ == "unsigned int")
         return StringFrom(value_.unsignedIntValue_);
--- a/include/CppUTestExt/MockActualCall.h
+++ b/include/CppUTestExt/MockActualCall.h
@@ -43,6 +43,7 @@ public:
 
     virtual MockActualCall& withName(const SimpleString& name)=0;
     virtual MockActualCall& withCallOrder(int callOrder)=0;
+    MockActualCall& withParameter(const SimpleString& name, bool value) { return withBoolParameter(name, value); }
     MockActualCall& withParameter(const SimpleString& name, int value) { return withIntParameter(name, value); }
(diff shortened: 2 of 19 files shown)


arstrube:
I don't know. I used to be in favor of support bool (and long long). But, admittedly, I don't like the looks (and implications) of this.

In many ways, I wish we could use `stdint.h` and `stdbool.h`.

Since apparently we can't, in my opionion, perhaps it is better not to try and support bool or long long...

uecasm:
I could remove the conversion support, but then if you're setting expectations or doing actual calls via the `mock_c()` interface then you must either use that on the other side as well (which will prevent you from using bools) or you must remember to use the integer method on the other side even for things that are logically bool.

(That sounds more complicated than it actually is.  I'm personally on the fence whether conversion is a good idea or not [so I'm happy to remove it if that's the consensus], but discussion in the prior PR suggested it might be useful.)

arstrube:
I guess treating `bool` as a distinct type in C++ definitely makes sense, so conversion support for C makes sense, too.

arstrube:
Btw, Coveralls is saying that some of your new code has not been exercised by any tests yet ;-)

uecasm:
Unfortunately due to the code structure I don't think I can restrict the conversion to just the C code -- the C++ code will get it too.  Otherwise there would be certain usages that didn't work, and it would be weird.

I'm ignoring the coveralls result for now until decisions are made on which approach is preferred: this way or the previous PR, and whether to retain conversion or remove it.

src/CppUTestExt/MockNamedValue.cpp
```diff
@@ -138,6 +144,17 @@ SimpleString MockNamedValue::getType() const
     return type_;
 }
 
+bool MockNamedValue::getBoolValue() const
+{
+    if(type_ == "int")
+        return (value_.intValue_ != 0);
```
basvodde: Usually this happens accidentally, so I'm not sure whether we would want the conversion. What was your thought behind this one?

src/CppUTestExt/MockNamedValue.cpp
```diff
@@ -151,8 +168,13 @@ unsigned int MockNamedValue::getUnsignedIntValue() const
 
 int MockNamedValue::getIntValue() const
 {
-    STRCMP_EQUAL("int", type_.asCharString());
-    return value_.intValue_;
+    if(type_ == "bool")
+        return (value_.boolValue_ ? 1 : 0);
```
basvodde: Likewise :)

src/CppUTestExt/MockNamedValue.cpp
```diff
@@ -265,10 +287,16 @@ bool MockNamedValue::equals(const MockNamedValue& p) const
         return (value_.longIntValue_ >= 0) && (value_.longIntValue_ == (long) p.value_.unsignedLongIntValue_);
     else if((type_ == "unsigned long int") && (p.type_ == "long int"))
         return (p.value_.longIntValue_ >= 0) && ((long)value_.unsignedLongIntValue_ == p.value_.longIntValue_);
+    else if((type_ == "bool") && (p.type_ == "int"))
+        return (value_.boolValue_ == (p.value_.intValue_ != 0));
+    else if((type_ == "int") && (p.type_ == "bool"))
+        return (p.value_.boolValue_ == (value_.intValue_ != 0));
```
basvodde: Likewise

src/CppUTestExt/MockNamedValue.cpp
uecasm: I covered that in the description.  The particular cases to imagine:

- A C++ test uses `mock().expectOneCall("a").withBoolParameter("x", true).andReturnValue(false);`
The corresponding C implementation uses `return mock_c()->actualCall("a")->withIntParameter("x", true)->intReturnValue();`.

- The reverse case, where C sets an expectation and C++ the actual call.  (This is probably rarer due to the way the library is designed.)

In this variant, the C API cannot use the bool type, so it must use the int type.  The C++ side _could_ use the int type as well but this is weird where the type is logically bool.

uecasm:
Pondering the issue of C and C++ bools a bit more, an alternate approach has occurred to me.

Assuming we still want to avoid the "real" C99 bool for some reason, we can still define bool C API (`withBoolParameters`, `boolReturnValue`, etc), but use `int` as the actual storage type on the C side and `bool` on the C++ side.

This should allow everything to Just Work™, even without the conversions, especially since C bools are int-like anyway.  Let me know if you want me to update this PR to that approach, or make a separate PR for it.

basvodde:
Sounds reasonable.

uecasm:
Ok.  Updated to int-based on the C side (no conversions needed), and has full coverage. :grinning: 

Though coveralls is acting a bit odd.

basvodde:
MinGW failure seems unrelated...

PR #934, opened 2016-04-13
Make mock_c() return values type safe

arstrube:


arstrube:
@uecasm - continuing from the issue you raised -

Basically, it means that most often (and somewhat confusingly) you get the incorrect return type (int instead of <correct_type>) because you used xxxxReturnValue() calls when there was not actual call happening (for example, because of ignored calls, or disabling mocking). This has always been the case in C++, and my PR propagates that behavior to C. If you revert my changes to the tests, you will see failures of that kind.

The xxxxReturnValueOrDefault() versions (which I also made available in the C API) are meant to be used in situations where an actual call may never occur.

My changes in this PR cause failures for incorrect usage, so I had to change some tests, to eliminate incorrect usage.

arstrube:
P.S. It would be nice if MockNamedValue could somehow fail when it wasn't assigned a type, rather than set int as default type. Less confusing for the user...

uecasm:
The C++ methods do an assert on the return type, so you should get a test failure if you try to fetch the "wrong" return type.  This seems better than silently getting the wrong value (although hopefully that would also then fail a test in the user code, but possibly in a more obscure way).

src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -635,7 +635,7 @@ MockValue_c returnValue_c()
 
 int intReturnValue_c()
 {
-    return getMockValueCFromNamedValue(actualCall->returnValue()).value.intValue;
+    return actualCall->returnIntValue();
 }
 
 int returnIntValueOrDefault_c(int defaultValue)
```
uecasm: Couldn't this just be:

    int returnIntValueOrDefault_c(int defaultValue)
    {
        return actualCall->returnIntValueOrDefault(defaultValue);
    }

?  It shouldn't be legal to call this without a prior `actualCall` having been made at some point, so it shouldn't matter that it's initially null.  (And you deserve the crash if you try it.)

Note that `MockSupport::actualCall` (and `actualCall_c`) cannot return null.  So this can only occur if you're doing it wrong.

PR #935, opened 2016-04-18
Test support for long-long

uecasm:
Unit testing support for long long, extracted from #847 (mock support excluded).

Support is enabled by default for the Visual Studio projects (but still needs to be manually enabled on the user side by defining `CPPUTEST_USE_LONG_LONG` in the user project settings).

AutoTools and CMake disable support by default but include an option to enable it (which will end up in `config.h` and will thus be automatically picked up by user code).

Support is enabled when Travis compiles for coverage testing.

If support is disabled, the APIs still exist but have a dummy placeholder type such that user code should fail to compile if they're trying to use long longs.  If support is enabled on the user side but disabled on the library side then the code will compile but will raise an "unsupported feature" test failure at runtime.


uecasm:
(BTW if you want to make the changes easier to review, in the **Files changed** tab on GH you can click on _showing changes from **All commits**_ to select each commit to view the changes separately.)

basvodde:
Yup, always use the Files changed tab :)

The coverage has increased... interesting ;)

include/CppUTest/UtestMacros.h
```diff
@@ -200,6 +200,29 @@
 #define UNSIGNED_LONGS_EQUAL_LOCATION(expected, actual, text, file, line)\
   { UtestShell::getCurrent()->assertUnsignedLongsEqual((unsigned long)expected, (unsigned long)actual, text, file, line); }
 
+#ifdef CPPUTEST_USE_LONG_LONG
```
basvodde: This doesn't need to be in #ifdef anymore actually, right? I'm thinking what would be better, have an undefined compiler error or a run-time error? Although I guess there won't be a runtime error, but an assignment to struct error that would be confusing, right?

Should we have an #else which will give a better error message when compiling without long long support?

basvodde: Actually, thinking it over, I suggest to remove the #ifdef completely. Reason is that when using CppUTest as a library, then it is not unlikely that the header is included without the CPPUTEST_USE_LONG_LONG define. So they'll get an error, *even though* they might have compiler CppUTest with long long support. So, this #ifdef might force the user of CppUTest to #define USE_LONG_LONG

uecasm:
The coverage increased because more code was added but the code was already covered.  So the pre-existing uncovered line of code is now less significant percentage-wise.

src/CppUTest/SimpleString.cpp
```diff
@@ -478,6 +478,40 @@ SimpleString HexStringFrom(unsigned long value)
...
+}
+
+SimpleString HexStringFrom(void (*value)())
+{
+    return HexStringFrom((cpputest_ulonglong) value);
+}
+
+#else   /* CPPUTEST_USE_LONG_LONG */
+
 static long convertPointerToLongValue(const void* value)
```
basvodde: Not quite understand why this isn't needed when we add long long support. It feels unrelated ?

include/CppUTest/UtestMacros.h
uecasm: The `#ifdef` is there so that they'll get a compiler error immediately, instead of a test failure at runtime (which is what would happen otherwise).

The second comment doesn't make sense to me.  There are two cases:

1. If the library was built using AutoMake/CMake, it produces a `config.h` file that "bakes in" the correct setting.   Since applications must always use these header files when using the library (they're part of the devel package), this will Just Work™ and they can't get it wrong unless they're being deliberately silly (in which case it will react as sensibly as it can).  It's never legal for applications to download a separate set of headers when using a pre-built package -- they **must** use the ones from the corresponding devel package.

2. If the application and library are built in VS then it is possible for the two to disagree on the setting.  In this case it makes sense for the user code to fail to compile if they try to use LONGLONG asserts without enabling the feature in their user code (this is entirely independent of whether it's supported in the library or not).  If, after doing so, it turns out that they enabled it in the user code without enabling it in the library, then they'll get a test failure at runtime that tells them this (but this is unlikely as the feature is enabled by default for VS library builds).

tests/UtestTest.cpp
```diff
@@ -232,6 +232,33 @@ TEST(UtestShell, DestructorIsCalledForLocalObjectsWhenTheTestFails)
 
 #endif
 
+#ifndef CPPUTEST_USE_LONG_LONG
+// otherwise these are tested by UnitTestMacros
```
basvodde: These can also be tested via the macros when the #ifdef is removed :)

include/CppUTest/UtestMacros.h
uecasm: Although, I should qualify the above a bit.

With the `#ifdef`, attempting to use an assert macro with the feature unsupported will produce an `undefined symbol LONGLONGS_EQUAL` or similar compiler error.

Without the `#ifdef`, attempting to use an assert macro with the feature unsupported will produce an `cannot convert 5 to cpputest_longlong for parameter expected in assertLongLongsEqual` or similar compiler error.

So I guess it's just a question of which error message is more obvious to the user (and of course different compilers will have different messages, of varying understandability).

basvodde: I don't know about cmake, but... when you install the apt-get package or the homebrew package, at the moment the config.h that is generated by autoconf is NOT installed yet. This has been something to-do as it would require renaming the config.h to something else (as the filename config.h is a bit too easy to lead to mistakes) and then prefix the configure results and install it.

That means that currently the system environment #defines discovered by autotools aren't kept after it is installed. Thus, it could be very common for someone to install CppUTest and not have USE_LONG_LONG #defined. The question is whether we want to let them #define it in their project or try to get it to "just work". By removing the #ifdef, we might be able to get a run-time (assigning an int to a struct, a linker error (not defined the StringFrom methods), or a run-time error (unsupported feature))

My guess is that this would be better, as in most cases such a thing shouldn't happen.

uecasm: For a specific example, using VS, with the `#ifdef` produces:

    error C3861: 'UNSIGNED_LONGLONGS_EQUAL': identifier not found

And without the `#ifdef` produces:

    error C2664: 'void UtestShell::assertUnsignedLongLongsEqual(cpputest_ulonglong,cpputest_ulonglong,const char *,const char *,int,const TestTerminator &)' : cannot convert argument 1 from 'unsigned __int64' to 'cpputest_ulonglong'
          No constructor could take the source type, or constructor overload resolution was ambiguous

basvodde: Yes, the problem is... it is quite likely that someone didn't define USE_LONG_LONG and it is much less likely that someone writes long long code on a system where it isn't supported.

But that said, related to the above error. Can you add a constructor to the struct, so that you get the runtime error?

uecasm: No, because there is no type that the constructor could accept, by definition.  (If someone has a compiler that genuinely does not support the `long long` type, then doing anything else would make them unable to use CppUTest at all, which is obviously undesirable.)

It's actually nearly impossible to get into a situation where the runtime error will trigger, except in CppUTest's own unit tests (since they're carefully written to work in both cases).  Casual mistakes will almost certainly trigger either a compile-time or link-time error.  But that's a good thing, if perhaps slightly less self-explanatory.

basvodde: If you add an int to the constructor, wouldn't the long long get downcasted to int by default?

src/CppUTest/SimpleString.cpp
uecasm: They're defined differently in the `#else` path (as originally defined prior to longlong support).

Basically, if `long long` is enabled then it converts pointers to longlongs prior to formatting to avoid possibly trying to convert a 64-bit pointer to a 32-bit long (on Windows), as previously discussed as a problem in `convertPointerToLongValue` just below.

On Linux this might result in unnecessarily converting a 64-bit pointer to a 128-bit integer prior to formatting, but other than marginal performance loss this should be harmless.

It does assume that `unsigned long long` is the same size or larger than a pointer on all platforms, but as far as I am aware that is true on all current platforms.  In the event it fails to be true on some future platform (which seems unlikely) then it should generate a warning or error in that case and we can worry about it then.

tests/UtestTest.cpp
uecasm: True, but that doesn't reduce the NLOC for the tests, since this tests the disabled case while the other tests the enabled case, and they behave differently.

include/CppUTest/UtestMacros.h
uecasm: No, it's still an error.  The only time it would work is if an integer constant of smaller size is used (eg. `1LL` will error, `1L` or `1` would work).  Using a variable or return value of `long long` type will still error regardless.

Additionally, this code is also used for C, so constructors are illegal.  If one was added it'd have to get protected in a further `#ifdef __cplusplus` block, which just complicates things further.

basvodde: Ok, nevermind that idea then :)

Anyways, I still suggest to remove the #ifdef even though the error is much less clear. Reason is that not defining USE_LONG_LONG is quite likely, whereas calling the long long assert on a system that doesn't support long long is rather unlikely. So, to get the best default, the #ifdef can be removed.

uecasm: Sorry, looks like I missed one of your earlier messages.

> I don't know about cmake, but... when you install the apt-get package or the homebrew package, at the moment the config.h that is generated by autoconf is NOT installed yet. This has been something to-do as it would require renaming the config.h to something else (as the filename config.h is a bit too easy to lead to mistakes) and then prefix the configure results and install it.

If so, that seems like a packaging error.  The config.h file is just as much a part of the required headers for the devel package as any of the others, since the user won't be running configure again.

It doesn't need renaming as it is only included by a file in the same directory so cannot possibly conflict with anything else in CppUTest itself, and shouldn't conflict with anything else that is built with CppUTest in the include path since they should have similar local conventions (at least on anything POSIX-like).

It shouldn't be included in source packages, of course -- for those the user is expected to run `configure` themselves (though not `autoreconf`, which is why `configure` is included even though it's a generated file).

basvodde: It is definitively a packaging issue, but it isn't a trivial issue. Autotools actually doesn't recommend installing the generated config.h in the /usr/local/include (which is where the library is installed). It comes with a lot of #defines that, unless renaming, could do some serious harm. This is why I left it out when working on autotools and left it for future work.
(see e.g. http://stackoverflow.com/questions/19586211/correct-installation-of-config-h-for-shared-library-using-autotools)

So, I try to let as little of CppUTest headers be dependent on config. That tends to lead to nicer headers anyways as avoiding #ifdef is almsot always a good idea. So, in this case...

uecasm: > Anyways, I still suggest to remove the #ifdef even though the error is much less clear. Reason is that not defining USE_LONG_LONG is quite likely, whereas calling the long long assert on a system that doesn't support long long is rather unlikely. So, to get the best default, the #ifdef can be removed.

I'm not quite sure what you mean by this.  Defining the macros in cases where they can't work seems misleading (I conceded on this point for the virtual calls, since there can be situations where leaving them undefined causes trouble, but that is not the case here).

And if someone does not define `USE_LONG_LONG` and does not use the long long asserts, then there's obviously no errors.

So really, which error message is "better" seems like the most important consideration.

uecasm: > It is definitively a packaging issue, but it isn't a trivial issue. Autotools actually doesn't recommend installing the generated config.h in the /usr/local/include (which is where the library is installed).

True.  But I assumed that it should end up as `/usr/local/include/CppUTest/config.h` when installed, along with most of the other headers.  Then there wouldn't be any problem.

basvodde: Not really, because the question is which situation happens most often and what do we ask the user.

So, two scenarios:

1) People call LONG_LONG when their system doesn't support that. How likely? Can't imagine anyone doing that, except when you port code. But then the long long declaration should give an error already

2) People call LONG_LONG without defining USE_LONG_LONG. How likely? Extremely likely, will happen for every installation on Mac and Linux. Can be fixed by letting the user manually define LONG_LONG. Do we want to do that. Not really.

Then, simply related to code clarity, NOT having a #idef is always preferred.

Now, unless you can convince me that scenario #1 is very common too... I'd suggest to remove it :)

basvodde: re: installing config.h

It would be, because the CppUTestConfig.h would need to include it, so there *all* the #defines that are generated in the config.h will be in the same namespace.

E.g. if I do, in my test code:

#include <CppUTest/TestHarness.h>  <--- will include config.h
#include <myapp.h>

Not the behavior of myapp.h *might* be influence by the config.h #defines. If that would happen, it would be exceptionally painful to find. The solution is to rename all the #defines in the config.h (which autotools provides some support for), but it would require quite some additional autotools work.

uecasm: Regarding the scenarios, I agree that the first is unlikely and the second is what we're trying to cover.

My point is that _within_ the second scenario, there's two further cases:

1. If we have the `#ifdef`, then the user gets an error that the macro doesn't exist.  Confused, they look at the header file, immediately spot that it's inside an `#ifdef` block, and realise that they need to enable it.  They do so, everything now works.

2. If we don't have the `#ifdef`, then they get an error that a long long value can't be assigned to a long long parameter (they'll probably miss at first that it's `cpputest_longlong` rather than a "real" long long).  Confused, they look at the macro definition.  It gives them no hints.  They look at the method declaration.  It gives them no hints.  Finally, they notice the typedef and look at the type definition, then finally notice that it's in an `#ifdef` (or see the nearby comment about being a placeholder) and realise what they need to do.

The former seems less confusing than the latter to me.  Especially since following the latter chain assumes the use of an IDE with a good "go to definition" implementation -- it'll require a lot more hunting around otherwise.

Making the header file very slightly "prettier" doesn't seem like an important consideration to me compared to that.

basvodde: Yes, but for

1) We are asking the user to add a CppUTest include to *their* Makefile or project file. I would like to avoid that as much as possible as most Makefile/project files already have way too many #defines.

2) Unlikely. If we don't have the #ifdef and someone would try to use the macro, then they'll need to actually define a long long. If they do so, it would give an error. Unless they try to pass it with a fixed int constant, then we can add a constructor to make a proper run-time error.

basvodde: Making headers more pretty is always a consideration for me, as it is likely to reduce maintenance ;)

uecasm: I still don't really agree (the cases where the user will get the error are the same regardless of whether the `#ifdef` is there or not -- it just changes which error they get and how they discover how to fix it).

But you seem set on this and I'm tired of arguing about it, so I'll remove it in a few minutes.  At which point this whole thread will disappear. :laughing:

basvodde: I'll miss the thread though :)

Could you then also change the non LONG LONG cases using the macros? (and delete the comment)

uecasm: > Could you then also change the non LONG LONG cases using the macros? (and delete the comment)

I'm not sure what you mean by this.

tests/UtestTest.cpp
basvodde: I mean, test these using the macros :)

include/CppUTest/CppUTestConfig.h
```diff
@@ -199,6 +199,27 @@
 #define CPPUTEST_CHAR_BIT 8
 #endif
 
+/* Support for "long long" type */
+#ifdef CPPUTEST_USE_LONG_LONG
```
basvodde: Thinking about all the previous discussion. It still won't work because of this. Did you try to tell me that earlier or did we both miss this?

uecasm: What won't work?

basvodde: If CPPUTEST_USE_LONG_LONG is not #defined, then the header will contain a struct to the assert, and therefore you will still get a compiler error.... even when your code does support long long.

uecasm: By definition, if it's not defined, then long long is not supported.  That's the whole point of it.

basvodde: My stupid mind thought you referred to when you try to use the checks on systems that don't support long long. I *thought* it would work also without #defining this, but now I see that isn't possible (yet)

basvodde: Ok, I can't figure out (yet) how this could work without #defining USE_LONG_LONG (or, in other words, by moving the usage to the cpp file). I won't ask you to put the earlier things back, let me later (tonight, my time) clean up this mess I created :)

basvodde: I guess the only way would be to memcpy it in a big enough buffer. That might work, but isn't nice.

uecasm: Ah.  No.  In the original January version of the patch I tried to be more clever and autodetect compiler support (there's still a few traces of that lingering in this version, such as `AC_TYPE_LONG_LONG_INT` that's not actually used other than to taunt people who read `config.h`).  But as you may recall that sometimes ended up with weird behaviour when eg. the C++ and C compilers detected differently.  So now it's just a simple manually-specified toggle.

Having said that, there **is** a basic auto-detection I could add back in, possibly -- just a few lines up you do (conditionally) include `limits.h`.  This does contain something that should theoretically tell us whether the compiler supports `long long`s (if `LLONG_MAX` is defined).

Although it's possible that some CRT libs define this even if the compiler doesn't support it, or that the application wants to build without them because they want to compile against an older C/C++ standard.  So for those reasons it might still be better to stick with a manual define.

(Though again, in the January PR I had it tristate -- you could define `CPPUTEST_USE_LONG_LONG` to force support on, `CPPUTEST_NO_LONG_LONG` to force support off, and neither to let it try to auto-detect.)

basvodde: Even memcpy wouldn't work as it wouldn't work with literals.

It seems that the only way around this is through fixing the packaging. So, during configure, we'll need to generate a file that specifies CppUTest specific #defines and then install that also. It would need to be empty by default in order not to break the builds of other systems.

I think I'll be looking into this tonight, or did you have happened to do a similar thing yet? We might jsut use CppUTestConfig or need to add CppUTestConfigGenerated.h or so and include it in CppUTestConfig.h

uecasm: I did consider letting it actually copy data around as an opaque memcpy'd type.  But there are technical problems with that (you'd need to add conversion macros, and there's still problems with formatting, especially cross endian).  And even after all of that it wouldn't support large numeric literals, which inevitably someone will want to use in these sorts of tests.

So it makes things very messy for very little gain, since if you're using a compiler that really doesn't support long longs, you're obviously not going to have code that has variables of that type.  Even if it's a cross-platform app where **most** platforms support long long, if there's one that doesn't you'll either have to avoid using it or do some conditional code for that platform, in which case the tests won't need to use long longs on that platform either.

As it stands this seemed like a reasonable compromise: if you define the option then you get support (and things fail to compile if your compiler doesn't support the type, as you'd expect).  If you don't define the option then you don't get support (even if the compiler could support it) but everything else works as expected.

And if you somehow end up defining the option differently between compiling CppUTest and compiling the unit test application, then you get a link error, due to name mangling differences between the symbols.  I'm not sure if there's a good way to avoid this (you probably don't want to avoid it in the case where the lib doesn't support but the app wants it, but the reverse is possibly more reasonable).

uecasm: No, I haven't tried messing with the packaging at all.  I'm not sure where to look. :confused:

uecasm: Hmm.  Actually there might be a way to bypass the linker errors in the mixed case.  I'll have to think about that tomorrow though, it's getting pretty late here.

basvodde: Ok. Let me explore this packaging tonight (a couple of hours) if I'm still awake by then. I'll let you know. I think most of this PR is in the right direction so I'll make sure this is merged at least by the end of the week. It has been open too long already!

re memcpy. Entertained the idea and decided it causes more trouble. Especially when doing LONG_LONG_CMP(10, 10) you can't do a memcpy as the &10 will unlikely be a good thing :)

basvodde: Ok. Let me explore this packaging tonight (a couple of hours) if I'm still awake by then. I'll let you know. I think most of this PR is in the right direction so I'll make sure this is merged at least by the end of the week. It has been open too long already!

re memcpy. Entertained the idea and decided it causes more trouble. Especially when doing LONG_LONG_CMP(10, 10) you can't do a memcpy as the &10 will unlikely be a good thing :)

basvodde: Ok. Made 3 pull requests related to packaging. Now it will install the config.h but renamed and prefixed, so that it shouldn't lead to conflict. I'm not using it in the CppuTestConfig.h, but that is the next step. It turned out easier than I had expected (good).

The same changes will need to be made on the CMake side, I guess. Delivering CppUTest with an empty CppUtestGeneratedConfig.h ought to be able to deal with the others, such as VC++.

uecasm: I haven't had a chance to try it yet (soon!), but I think I can make it so that you can centrally install one version with `USE_LONG_LONG` enabled, and then link against it both with or without `USE_LONG_LONG`.  Though it requires a couple of minor shenanigans.

If you centrally install a version without longlongs then you'll still get link errors if you try to `USE_LONG_LONG` in the app, but that seems reasonable too.

uecasm: I've added a new commit (separate, so it's easily readable) that allows the library to be built with `USE_LONG_LONG` and successfully link test applications both with and without it (though of course trying to use the long-long APIs from an app that doesn't enable them won't work).

I haven't added tests for them yet (because that's a bit tricky) so coverage will go down a bit, but I'm mostly just pushing it at this point to get your thoughts on the general concept and implementation.  I can remove it again if you're happier with the baked-in-packaging approach (at least on Linux).

uecasm:
I've added a new commit (separate, so it's easily readable) that allows the library to be built with `USE_LONG_LONG` and successfully link test applications both with and without it (though of course trying to use the long-long APIs from an app that doesn't enable them won't work).

I haven't added tests for them yet (because that's a bit tricky) so coverage will go down a bit, but I'm mostly just pushing it at this point to get your thoughts on the general concept and implementation.  I can remove it again if you're happier with the baked-in-packaging approach (at least on Linux).

src/CppUTest/TestDownlevelApi.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+void UtestShell::assertLongLongsEqual(cpputest_longlong expected, cpputest_longlong actual, const char* text, const char* fileName, int lineNumber, const TestTerminator& testTerminator)
+{
+    failWith(LongLongsEqualFailure(this, fileName, lineNumber, expected, actual, text), testTerminator);
+}
+
+void UtestShell::assertUnsignedLongLongsEqual(cpputest_ulonglong expected, cpputest_ulonglong actual, const char* text, const char* fileName, int lineNumber, const TestTerminator& testTerminator)
+{
+    failWith(UnsignedLongLongsEqualFailure(this, fileName, lineNumber, expected, actual, text), testTerminator);
+}
```
uecasm: Sigh.  Should have double-checked this on Linux first I guess.  This is working perfectly in VS, but compiling with either GCC or Clang in Linux causes link failures with `CPPUTEST_USE_LONG_LONG` defined on both sides now:

    lib/libCppUTest.a(lib_libCppUTest_a-Utest.o): In function `UtestShell::assertUnsignedLongLongsEqual(unsigned long long, unsigned long long, char const*, char const*, int, TestTerminator const&)':
cpputest_build/../src/CppUTest/Utest.cpp:446: multiple definition of `UtestShell::assertUnsignedLongLongsEqual(unsigned long long, unsigned long long, char const*, char const*, int, TestTerminator const&)'
    lib/libCppUTest.a(lib_libCppUTest_a-TestDownlevelApi.o):cpputest_build/../src/CppUTest/TestDownlevelApi.cpp:48: first defined here

I don't understand why this is happening.  One definition is for `struct cpputest_ulonglong` and the other is for `unsigned long long` so they should mangle to different identifiers and not conflict, which is how it works in VS.

So maybe this won't work after all.  Or maybe this is a VS-only fix and the packaging is the GCC/Clang fix?  Seems messy.

uecasm: Although wait.  Travis seemed to build ok in this case.  So maybe something just isn't being cleaned properly on my system or something?

uecasm: Argh.  I just don't get this.  I've cleaned everything I can think of, deleted everything in the build directory, and still I get the link error in Linux.

What's really weird is this:

    $ nm -C cpputest_build/src/CppUTest/lib_libCppUTest_a-TestDownlevelApi.o | grep LongLong
    ...
    00000000000000b0 T UtestShell::assertUnsignedLongLongsEqual(unsigned long long, unsigned long long, char const*, char const*, int, TestTerminator const&)

There is no way that this parameter should be `unsigned long long` type.  `CPPUTEST_USE_LONG_LONG` and `CPPUTEST_BUILD_DOWNLEVEL_API` are both defined, so it should fall through to the version that defines it as a struct instead.  WTF?

uecasm: Apparently building with autotools causes a link error while building with cmake does not.  But both are defining the symbol wrong when building TestDownlevelApi.cpp, and I still don't know why.

basvodde: Nice monologue. Any comments on the packaging changes I made? I'll merge them, then we can assume USE_LONG_LONG is set to something.

uecasm:
Since #941 is merged, this isn't needed any more.

PR #938, opened 2016-04-18
Install the config.h upon make install

basvodde:
Install a CppUTest prefixed config.h upon installation so that the system configuration can be kept there.


uecasm:
The PR isn't clean; it contains commits from other PRs.  And the one change that's in ca0dcaa just seems to be adding a file that is not actually used, so seems insufficient for the task.

basvodde:
The file is a m4 macro that is added to the configure.ac. Without it, it will fail :)

uecasm:
I'm not disputing that.  But configure.ac would need to be changed to use the file, as would any usage of the now-prefixed config.h macros.  Those changes are not present.

basvodde:
No, autoconf automatically loads the macros from the m4 directory. So the added line AX_ is actually using the macro

basvodde:
The line I can't find. Lemme check :)

basvodde:
Ok, now they are there.

uecasm:
There **is** no added `AX_` line in the PR.  That is my point.  `configure.ac` has not been modified at all.

basvodde:
Yes, you were right :) I had not committed them. My mistake.

Makefile.am
```diff
@@ -80,7 +80,8 @@ include_cpputest_HEADERS = \
 	include/CppUTest/TestResult.h \
 	include/CppUTest/TestTestingFixture.h \
 	include/CppUTest/Utest.h \
-	include/CppUTest/UtestMacros.h
+	include/CppUTest/UtestMacros.h \
+	CppUTestGeneratedConfig.h
```
uecasm: Recommend putting this into `include/CppUTest` as well.  Certainly it should be in that relative location when installed.

Also there should be a default file for VS and cmake builds.  Including regenerating it if autotools overwrote it, then someone does a subsequent cmake build.

basvodde: It gets installed in include/CppUTest. The file is the source file and that one is in the local directory because it is a generated file.

PR #940, opened 2016-04-19
Used the installed generated config.h file

basvodde:
Only done for autotools. Probably needs similar changes for cmake.


basvodde:
These are pretty deep changes to CppUTest and its install. They need to be mimicked in CMake and please review them well.

uecasm:
Regarding cmake, it already generates its config.h in the "right" format with nothing that will upset other software, so this could just be renamed to `generated/CppUTestGeneratedConfig.h`.

uecasm:
Note that most of the build failures with the current version are because tests for `HAVE_FORK` need to be changed to `CPPUTEST_HAVE_FORK` (Gcc/UtestPlatform.cpp, UtestPlatformTest.cpp, and UtestTest.cpp).

Also `CMakeLists.txt` line 14 needs to be adjusted similarly for the cmake builds.

basvodde:
Thanks. Made the changes. I assume that CMake also supports an installation. If so, it'll need to install the generated config file too then.

PR #941, opened 2016-04-20
Long long test support

basvodde:
Some minor changes (and reverts) from @uecasm long long support pull request. Now it uses the CppUtestGeneratedConfig when installing, which will by default support long long then.


include/CppUTest/CppUTestConfig.h
```diff
@@ -216,8 +218,6 @@ typedef struct { char dummy[16]; } cpputest_ulonglong;
 typedef struct { char dummy[8]; } cpputest_longlong;
 typedef struct { char dummy[8]; } cpputest_ulonglong;
 #endif
-#define CPPUTEST_LONGLONG_DEFAULT cpputest_longlong()
-#define CPPUTEST_ULONGLONG_DEFAULT cpputest_ulonglong()
```
uecasm: These are essential if you want coverage in the case where `CPPUTEST_USE_LONG_LONG` is not defined (it will not compile without them).  If you don't care about coverage in that case then many tests could be removed to eliminate use of this value, but as it stands it's not safe to remove these.

include/CppUTest/CppUTestConfig.h
```diff
@@ -200,11 +200,13 @@
 #endif
 
 /* Support for "long long" type */
+#ifdef CPPUTEST_HAVE_LONG_LONG_INT
+#define CPPUTEST_USE_LONG_LONG 1
+#endif
```
uecasm: This is exactly the sort of auto-detection that was objected to in my first revision of the PR...

basvodde: Partly because it didn't work without the previous changes to autotools :)

include/CppUTest/CppUTestConfig.h
basvodde: Lets see if we can think of a better way than using this DEFAULT #define :) So far, it wasn't a problem (yet). The coverage decreased a bit, need to check still why. Thanks!!

uecasm: Declaring a variable of type `cpputest_[u]longlong` is fine as long as it's used exclusively in tests where `CPPUTEST_USE_LONG_LONG` is undefined.  It's more dodgy in the other case.  Having said that, I tried to not use the "default" value where real longlongs were available, so it might be ok.

uecasm: The coverage decrease is because you put the `#ifdef`s back around the macros, and removed the corresponding macro tests, so when `CPPUTEST_USE_LONG_LONG` is not defined it's not exercising `UtestShell::assertLongLongsEqual` any more.  I had extra unit tests in UtestShellTest for that case; want me to dig them up?

The other cause is that coverage is being done without long-long support, because that uses the CMake build and the CMake build hasn't yet had config.h renamed to generated/CppUTestGeneratedConfig.h, so CMake's options aren't included any more.

basvodde: Right, let me fix that. Thanks!

basvodde: Ok, got the non-long-long support tests back and think I fixed the default value problem for mocks probably too. Still didn't do CMake, so expecting the build to still fail.

basvodde:
@uecasm Anything that I missed still with this pull request?

configure.ac
```diff
@@ -366,6 +368,11 @@ else
 	CPPUTEST_CPPFLAGS="${CPPUTEST_CPPFLAGS} -include ${srcdir}/include/CppUTest/MemoryLeakDetectorMallocMacros.h"
 fi
 
+# Disablong long long support
+if test "x${use_longlong}" = xno; then
+  AC_DEFINE([DISABLE_LONG_LONG], 1, [enable long long])
+fi
```
uecasm: `LONG_LONG_DISABLED` might be a more consistent name with some of the other options.  Though it doesn't really matter.  Also the comment text is a little misleading.

include/CppUTest/CppUTestConfig.h
```diff
@@ -211,13 +213,31 @@ typedef unsigned long long cpputest_ulonglong;
...
+
+struct cpputest_ulonglong
+{
+#if defined(__cplusplus)
+  cpputest_ulonglong() {}
+  cpputest_ulonglong(int) {}
+#endif
+  char dummy[CPPUTEST_SIZE_OF_FAKE_LONG_LONG_TYPE];
+};
+
```
uecasm: These won't work with C code -- they have to be typedefs to avoid syntax changes.  eg:

    typedef struct cpputest_longlong_
    {
    #if defined(__cplusplus)
        cpputest_longlong_() {}
        cpputest_longlong_(int) {}
    #endif
        char dummy[CPPUTEST_SIZE_OF_FAKE_LONG_LONG_TYPE];
    } cpputest_longlong;

Although I'm still unconvinced of the merit of including an int constructor.  It won't prevent real tests from failing with compile errors (since the only tests that would compile are ones that use small integer constants, and other than in CppUTest's own tests that will never happen).  So all it really gains is the ability to use `0` and `1` in CppUTest's own tests instead of declaring a dummy `cpputest_longlong` default-constructed variable as done previously (via the macro in my version, or on a separate line in yours).

uecasm: And once the mock code comes along, it will need explicitly declared variables anyway in order to ensure that the correct overload gets called.

configure.ac
arstrube: Agree with @uecasm about the name. Tiny typo in the comment, too - `Disablong` -> `Disabling`

include/CppUTest/CppUTestConfig.h
arstrube: @uecasm I'm not sure I can follow about the not working in C code (or how it does work when it does). Could you suggest a C test that would / wouldn't work, depending on typedef / no typedef? Obviously, whatever is tested in C at present, does work. And generally, it is possible to use structs in C (e.g. as arguments) without typedef.

tests/TestUTestMacro.cpp
```diff
@@ -359,6 +359,30 @@ IGNORE_TEST(UnitTestMacros, UNSIGNED_LONGLONGS_EQUAL_TEXTWorksInAnIgnoredTest)
     UNSIGNED_LONGLONGS_EQUAL_TEXT(1, 0, "Failed because it failed"); // LCOV_EXCL_LINE
 } // LCOV_EXCL_LINE
 
+#else
+
+static void _LONGLONGS_EQUALFailsWithUnsupportedFeatureTestMethod()
+{
+    LONGLONGS_EQUAL(1, 1);
```
arstrube: This is more a general question (applying to other tests as well) - when testing signed integers, wouldn't it make more sense to test a signed value such as -1?

configure.ac
uecasm: Actually I meant the `[enable long long comment]`.  But that too. :+1:

basvodde: Thanks, fixing.

tests/TestUTestMacro.cpp
basvodde: Uhm, personally I don't think it matters much. When I see a -1, I'd start getting curious as it would feel like a special value :)

include/CppUTest/CppUTestConfig.h
uecasm: If you declare `struct cpputest_longlong`, in C++ you can refer to it just as `cpputest_longlong` but in C it must be referred to as `struct cpputest_longlong` unless there is also a typedef.

So yes, it's still possible to use it in C either way, but it's cleaner if the `struct` keyword is not required when using it, and this makes it more consistent with C++.  This is also important for unity with the case where `cpputest_longlong` is actually a typedef to the real type -- in this case obviously using `struct` is forbidden.  So it really does have to be a typedef.

As for working in C as-is, I've just realised that I only implemented long longs (and bools!) in the C mock API; I did not do so in the C test API, as I somehow missed that file entirely.  So it's working because they're not there. :sweat: 

Would it make things easier if I went back to the other PR and updated it to match this one, plus C-longlong support?

tests/TestUTestMacro.cpp
uecasm: In general yes, although in this specific case it doesn't really matter since the value isn't actually used.

include/CppUTest/CppUTestConfig.h
basvodde: I'd suggest to merge this one and move on to the Mock and/or C versions, so we keep the PRs small. It still has your original commits, just made a few changes on that ;)

tests/TestUTestMacro.cpp
arstrube: @basvodde ummm, yeah, but it could provoke a compiler error if (by mistake) it gets passed to code that only takes unsigned. I agree about -1 being used as a special value, so -17 perhaps :-) ?

include/CppUTest/CppUTestConfig.h
arstrube: @uecasm okay, understood about the `struct`. In that case second the `typedef` approach.

PR #942, opened 2016-04-22
Update cmake builds for new CppUTestGeneratedConfig.h file.

uecasm:
This will make cmake convert `config.h.cmake` into `generated/CppUTestGeneratedConfig.h` during build and then to `PREFIX/include/CppUTest/CppUTestGeneratedConfig.h` on install, to match the new autotools setup.

I don't know how to make it have an equivalent of `AC_TYPE_LONG_LONG_INT` so `long long` support is still opt-in for cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,10 +47,10 @@ include("${CppUTestRootDirectory}/cmake/Modules/CppUTestNormalizeTestOutputLocat
 
 configure_file (
     "${PROJECT_SOURCE_DIR}/config.h.cmake"
-    "${PROJECT_BINARY_DIR}/config.h"
+    "${PROJECT_BINARY_DIR}/generated/CppUTestGeneratedConfig.h"
     )
 include_directories(${PROJECT_BINARY_DIR})
-add_definitions(-DCPPUTEST_HAVE_CONFIG_H)
+add_definitions(-DHAVE_CONFIG_H)
 
 include_directories(${CppUTestRootDirectory}/include)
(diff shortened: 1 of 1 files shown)


offa:
So this will address issues like these?

```
In file included from /usr/local/include/CppUTest/TestHarness.h:31:0,
                 from /usr/local/include/CppUTest/CommandLineTestRunner.h:31,
                 […]
/usr/local/include/CppUTest/CppUTestConfig.h:32:37: fatal error: CppUTestGeneratedConfig.h: No such file or directory

compilation terminated
```

I've build using cmake (C++11 enabled):

    - mkdir _build && cd _build
    - cmake -DC++11=ON -DTESTS=OFF ..
    - make -j4 && sudo make install

offa:
It seems **yes**!

Using your branch solves the issue mentioned above (tested on Linux, C++11 enabled and Gcc 4.8, 4.9, 5.x,  Clang 3.6, 3.7).

offa:
Clang 3.8 is working too.

arstrube:
Umm - the Coveralls results seem rather strange to me. I think they can safely be ignored.

uecasm:
It's not strange.  The percentage has changed because the number of compiled lines has changed (because this makes the coverage include long longs, whereas it didn't before).  The actual coverage is the same as before, but fewer lines are compiled so the percentage of uncovered lines got slightly larger.

But yes, the net result is that the coverage failure should be ignored -- coverage is not worse in this branch, just slightly different.

PR #944, opened 2016-04-23
add description to numerical check failure

cuitoldfish:
example:
```.cpp
class globalConfigureFactory
{
public:
    static int getUserHighestPriority()
    {
        return 4;
    }
};

class usersPriorityGenerator
{
public:
    usersPriorityGenerator()
    {
        usersPriority[0] = 4;
        usersPriority[1] = 2;
    }
    int(&getUsersPriority())[2]
    {
        return usersPriority;
    }
private:
    int usersPriority[2];
};

TEST(testGroup, testUserPriorityCalculation)
{
    usersPriorityGenerator uPG;
        
    ...//some code here to calculate users priority using class usersPriorityGenerator
    LONGS_EQUAL(globalConfigureFactory::getUserHighestPriority(), uPG.getUsersPriority()[0]);
    LONGS_EQUAL(globalConfigureFactory::getUserHighestPriority() - 1, uPG.getUsersPriority()[1]);
}
```
in this example, I have more than one numeric check in TEST, and none use the simple and readable magic number as in the cppUTest framework's TEST such as `LONGS_EQUAL(1, 0xff);`, so if one of my check fail, I will get something like `expected <3 0x3>, but was  <2 0x2>`, this will not let me locate exactly and quickly which check really fail, because in my LONGS_EQUAL there isn't 0x3 or 0x2 there, I need to debug dynamically to find which function's result mapping to 0x3 and which mapping to 0x2.

I think a description to those numerical check will be efficient to take me to the exact failure CHECK, such as `LONGS_EQUAL(globalConfigureFactory::getUserHighestPriority() - 1, uPG.getUsersPriority()[1]) failed!   expected <3 0x3>, but was  <2 0x2>`

LONGS_EQUAL_TEXT is not my choice because I don't want to send any special message to the failure, I just want to know which numerical check fail.

This PR only add description to LONGS_EQUAL, and let's see if extend to other numerical check too.
This TEST `TEST(UnitTestMacros, FailureWithLONGS_EQUALSUsedForNonMagicValue)` tell what this PR has done.

### DIFF
--- a/include/CppUTest/TestFailure.h
+++ b/include/CppUTest/TestFailure.h
@@ -72,6 +72,7 @@ protected:
     SimpleString createButWasString(const SimpleString& expected, const SimpleString& actual);
     SimpleString createDifferenceAtPosString(const SimpleString& actual, size_t position, DifferenceFormat format = DIFFERENCE_STRING);
     SimpleString createUserText(const SimpleString& text);
+    SimpleString createNumericalCheckDescription(const SimpleString& longsEqualString, const SimpleString& expected, const SimpleString& actual);
 
     SimpleString testName_;
     SimpleString testNameOnly_;
@@ -127,6 +128,7 @@ class LongsEqualFailure : public TestFailure
 {
 public:
     LongsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, long expected, long actual, const SimpleString& text);
+    LongsEqualFailure(UtestShell* test, const char* fileName, int lineNumber, long expected, long actual, const SimpleString& longsEqualString, const SimpleString& expectedString, const SimpleString& actualString, const SimpleString& text);
 };
 
 class UnsignedLongsEqualFailure : public TestFailure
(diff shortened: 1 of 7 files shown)


uecasm:
While this does seem useful, I'm not sure about the motivation.  You can use the line number of the failure to quickly find the line with the failing check.  This only gets a little ambiguous when used in a loop or a helper function that's called repeatedly, but the text parameter is supposed to help with disambiguating that (or you can just debug the test).

basvodde:
I'm a bit afraid it is too much information ? Haven't experienced this yet :)

cuitoldfish:
@uecasm, yes  I can use the line number of the failure to find the line with the failing check, so no need ti have extra codes to do the location

basvodde:
So quickly, I wanted to hear @arstrube opinion on this change before deciding what to do.

arstrube:
I just glanced at it quickly, so I may have missed something, but I couldn't quite figure out the motivation, either. At a glance, I would use the line number of the failure to quickly know which one failed.

The project I am supporting has checks in a loop (checking large arrays of floats as it were), in that case, it could be useful to know in which iteration you failed, and the extra text could provide that information. I wonder whether @basvodde would say you shouldn't have tests like that ;-)

Issue #946, opened 2016-04-25
Visual Studio 2005 build fails on IEEE754ExceptionsPlugin.cpp


BillDenton:
...\cpputest\src\CppUTestExt\IEEE754ExceptionsPlugin.cpp(34) : fatal error C1083: Cannot open include file: 'fenv.h': No such file or directory

This is the latest code pulled today. Could you fix this as soon as possible. Note: we also compile with 2013 for some projects, but I haven't test that yet.

Andne:
Unfortunately, VS2005 is not available on Appveyor, so that build doesn't get checked normally.  The project files do get checked since the VS2008 build uses them, just the VS2008 (VC8.0) compiler never gets run.

Actually, looks like the IEE754 files never got added to any of the Visual Studio projects.  With a bit of trial and error, I've narrowed down the version `fenv.h` was added in to 2013.  I'll submit a pull request containing the various changes to update the project files and build this logic when possible.

BillDenton:
Any update? Thanks.

arstrube:
@BillDenton do you actually need the plugin? Or just want to compile? If it is the first, I'm afraid this is not a trivial matter, because the availability of `fenv.h` is essential to the plugin. If the latter, just remove the plugin and its associated tests from your project.

@Andne if they were never added, how come they are part of the VS2005 build? I think the best option would be to remove them for now. Does VC support checking the IEEE754 floating point flags in any form at all? If so, as of which version?

@basvodde would it be okay with you if we omit the plugin for older version of Ms VC?

basvodde:
@arstrube Fine. Also, the plugin fails on my Mac. One test :) should_not_fail_again_when_test_has_already_failed. But only on some builds.

BillDenton:
I'm updating our internal copy of CppUTest the latest version so we have some of the newer macros. So I just need it to compile. I don't think we have used this plugin has I don't think it was in release 3.7.2 so we can't have used it. If it can't be supported in 2005, then just remove it from the build.

Andne:
@BillDenton What do you use to run the build?  Are you using the Visual Studio solution or CMake?

@arstrube Looks like it supports fenv.h as of 2015, but there's a test failure then.  I have the float exception logic suppressed in all other versions, just introduced a bad comment that broke everything else.  Excluding the files from the solution works, but not if someone is using cmake I suspect and we'd have to always remember why they're excluded.

arstrube:
[...]

PR #948, opened 2016-04-26
Add IEEE754 files to Visual Studio

Andne:
Also adds a check for _MSC_VER to disable the FENV logic for versions prior to 2013.  Older versions don't have the fenv.h header file that is required for this to work properly.

This should fix the issue seen in #946.

### DIFF
--- a/tests/CppUTestExt/IEEE754PluginTest_c.c
+++ b/tests/CppUTestExt/IEEE754PluginTest_c.c
@@ -76,4 +76,4 @@ void set_everything_c()
     set_inexact_c();
 }
 
-#endif // CPPUTEST_HAVE_FENV
+#endif /* CPPUTEST_HAVE_FENV */
--- a/tests/CppUTestExt/IEEE754PluginTest_c.c
+++ b/tests/CppUTestExt/IEEE754PluginTest_c.c
@@ -25,8 +25,6 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <CppUTest/CppUTestConfig.h>
-
 #include "IEEE754PluginTest_c.h"
 #include <math.h>
 
--- a/tests/CppUTestExt/IEEE754PluginTest_c.c
+++ b/tests/CppUTestExt/IEEE754PluginTest_c.c
@@ -27,8 +27,6 @@
 
 #include <CppUTest/CppUTestConfig.h>
 
-#ifdef CPPUTEST_HAVE_FENV
-
 #include "IEEE754PluginTest_c.h"
 #include <math.h>
 
@@ -42,14 +40,14 @@ void set_divisionbyzero_c(void)
 
 void set_overflow_c(void)
 {
-    f = 1000.0f;
-    while (f < INFINITY) f *= f;
+   f = 1e38f;
+   f *= f;
 }
 
 void set_underflow_c(void)
 {
-    f = 0.01f;
-    while (f > 0.0f) f *= f;
+   f = 1e-38f;
+   f *= f;
 }
 
 void set_invalid_c(void)
@@ -75,5 +73,3 @@ void set_everything_c()
     set_invalid_c();
     set_inexact_c();
 }
-
-#endif /* CPPUTEST_HAVE_FENV */
--- a/tests/CppUTestExt/IEEE754PluginTest_c.c
+++ b/tests/CppUTestExt/IEEE754PluginTest_c.c
@@ -25,6 +25,10 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <CppUTest/CppUTestConfig.h>
+
+#ifdef CPPUTEST_HAVE_FENV
+
 #include "IEEE754PluginTest_c.h"
 #include <math.h>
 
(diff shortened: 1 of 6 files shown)


tests/CppUTestExt/IEEE754PluginTest_c.c
```diff
@@ -71,3 +75,5 @@ void set_everything_c()
     set_invalid_c();
     set_inexact_c();
 }
+
+#endif // CPPUTEST_HAVE_FENV
```
arstrube: Just remove the comment. That will make the compiler happy, and @basvodde, too ;-)

arstrube:
Thanks for taking care of this :)

tests/CppUTestExt/IEEE754PluginTest_c.c
arstrube: It seems Wcl objects to the missing newline at the end of file, not sure though.

Andne:
@arstrube The build looks good now, but for some reason VS2015 (v140) has a couple failures.  Any thoughts on this?  Should we try to fix them or just disable the plugin completely on Visual C++ for now?

I've also noticed that since we run the tests as part of the build step normally, it's causing the failures to show as build failures instead of failed tests.  Should maybe look at what it would take to cleanup this behavior as well.  Not really specific to this problem, just a general build cleanup issue.

tests/CppUTestExt/IEEE754PluginTest_c.c
```diff
@@ -25,6 +25,10 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <CppUTest/CppUTestConfig.h>
+
+#ifdef CPPUTEST_HAVE_FENV
```
arstrube: Why this? There is nothing in this file that depends on whether or not fenv.h is present, or IEEE754 checks are supported.

arstrube:
Hi Andy, this looks indeed good (except for the one point noted above). The failures are due to the fact that compilers don't seem to implement some of the checks consistently. FE_INVALID is one of those. It doesn't mean that it's not working. We have several options here:

1. Remove FE_INVALID from all tests.
2. Use #ifdef in some way to exempt MS from these tests.
3. Set CPPUTEST_FENV_IS_WORKING_PROPERLY to 0 for MS (no tests will be run).

@basvodde what do you think?

tests/CppUTestExt/IEEE754PluginTest_c.c
Andne: It's because of the INFINITE define later on.  For at least some versions of Visual C++, INIFINITE isn't defined and so I was getting compiler errors.  This was the quickest way I thought of to avoid that issue (the file doesn't get used unless FENV is active), but the correct solution probably would have been to just block off the one function if INFINITE is missing.

arstrube: I probably shouldn't be using INFINITY in this file at all. Any idea how to rewrite the fucntion without INFINITY?

Andne: Looking online a bit it appears that there's an `isfinite(float)` function that could maybe be used instead.  I don't have access to the older compilers right now to check though.  Some of the information that I found implies that you're supposed to use `isfinite()` instead of comparing against `INFINITY` anyways.

Judging from online documentation, the closest in older VC++ implementations is going to be `_finite(double)`.  Looks like that provides the same functionality just under a different name.  It seems a bit ugly to wrap it in a macro just for Visual C, but I don't really know another solution and it's probably cleaner than excluding the entire file.  However, the functions in the file are unused unless FENV is enabled anyways, so how much does it actually matter if the file is compiled when FENV isn't active?

arstrube: Perhaps it doesn't :-). Actually I was hoping you might know some clever way to run into infinity without having to use a #define at all.... like FLOAT_MAX * 2 or something (I'm not really serious, because that would just be a different define ;-). Because it would be nice if this file contained nothing but plain C code.

arstrube:
```c++
void set_overflow_c(void)
{
    f = 1e38f;
    f *= f;
}
```
appears to work, at least on my machine. If it survives Travis, then I guess I could use that.

arstrube: Would you mind putting this in your PR (or I can submit a separate one):
```c++
void set_overflow_c(void)
{
    f = 1e38f;
    f *= f;
}

void set_underflow_c(void)
{
    f = 1e-38f;
    f *= f;
}
```
I couldn't find any good substitute for sqrt(-1), so math.h will have to stay.

Andne: Ok, I added that and removed the #ifdef from the file.

Andne:
I think I have `IEEE754PluginTest_c.c` cleaned up, any thoughts on how to handle VS2013 and VS2015?  They can pass some of the tests but a couple of them still fail.

arstrube:
I think both failures are due to the invalid flag. I see the following options:

- disable the tests alltogether (not a good idea since most pass)
- remove tests that involves invalid (get rid of math.h as well)
- #ifdef the invalid tests away for Msvc

I slightly prefer the second option over the third. Invalid will still work on those systems that pick it up, but it won't be documented as working by any tests. I would like to see sqrt() and math.h gone.

arstrube:
oooh. But just realized if we act on it and don't test it we will decrease coverage :-((

PR #951, opened 2016-04-28
Splitting TestUTestMacro.cpp

arstrube:
This isn't particularly elegant; I was trying to keep new files limited to one and avoid duplication of code.

1. Only the first commit is of interest; the rest will be taken care of by basvodde/long-long.

2. Better ideas?

3. How to divide up the file more sensibly? Or can we shrink the code (a lot of assembler instructions will be generated for doing 64 bit math on a 16 bit platform)?

### DIFF
--- a/include/CppUTest/TestTestingFixture.h
+++ b/include/CppUTest/TestTestingFixture.h
@@ -28,7 +28,6 @@
 #ifndef D_TestTestingFixture_H
 #define D_TestTestingFixture_H
 
-#include "CppUTest/TestHarness.h"
 #include "CppUTest/TestRegistry.h"
 #include "CppUTest/TestOutput.h"
 
--- a/include/CppUTest/TestTestingFixture.h
+++ b/include/CppUTest/TestTestingFixture.h
@@ -31,6 +31,8 @@
 #include "TestRegistry.h"
 #include "TestOutput.h"
 
+#define CHECK_TEST_FAILS_PROPER_WITH_TEXT(text) TestTestingFixture::CHECK_TEST_FAILS_PROPER_WITH_TEXT_LOCATION(text, fixture, __FILE__, __LINE__)
+
 class TestTestingFixture
 {
 public:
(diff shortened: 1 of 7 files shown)


include/CppUTest/TestTestingFixture.h
```diff
@@ -121,11 +123,40 @@ class TestTestingFixture
...
+        return lineOfCodeExecutedAfterCheck();
+    }
 
     TestRegistry* registry_;
     ExecFunctionTestShell* genTest_;
     StringBufferTestOutput* output_;
     TestResult * result_;
+    
+private:
+    static bool lineOfCodeExecutedAfterCheck(const bool* const value = 0)
```
arstrube: This is a bit of a hack, but it works without having to introduce a new .cpp file just for the static var to live in.

tests/TestUTestMacro2.cpp
```diff
@@ -0,0 +1,169 @@
...
+        TestTestingFixture::setLineOfCodeExecutedAfterCheck(false);
+    }
+
+    void runTestWithMethod(void(*method)())
+    {
+        fixture.setTestFunction(method);
+        fixture.runAllTests();
+    }
+};
+
```
arstrube: I just pulled out the long  long tests; this will work, but isn't particularly smart ;-)

basvodde:
Alternative split at #953

arstrube:
@basvodde fixing some errors this end - not sure how this can be:
```
In file included from ../tests/CppUTestExt/GTest1Test.cpp:83:
../include/CppUTest/TestTestingFixture.h:130:13: error: use of undeclared 
identifier 'FAIL_LOCATION'
FAIL_LOCATION(StringFromFormat("Expected one test failure, b...
^
../include/CppUTest/TestTestingFixture.h:135:13: error: use of undeclared 
identifier 'FAIL_LOCATION'
FAIL_LOCATION("The test should jump/throw on failure and no...
^
```
seeing that TestHarness.h is included before TestTestingFixture.h, and includes UTestMacro.h in turn? (of course it would be cleaner to implement this method in TestTestingFixture.cpp and #include TestHarness.h there, but that is an entirely different matter). The macro should be known? Why does it only cause problems in GTest1Test.cpp?

PR #953, opened 2016-05-02
Splitup the Test Macro file, alternative

basvodde:
Alternative split by splitting out String functions

### DIFF
--- a/src/CppUTest/CMakeLists.txt
+++ b/src/CppUTest/CMakeLists.txt
@@ -14,7 +14,7 @@ set(CppUTest_src
         MemoryLeakDetector.cpp
         TestFilter.cpp
         TestPlugin.cpp
-        TestTestingPlugin.cpp
+        TestTestingFixture.cpp
         SimpleMutex.cpp
         Utest.cpp
         ../Platforms/${CPP_PLATFORM}/UtestPlatform.cpp
--- a/src/CppUTest/CMakeLists.txt
+++ b/src/CppUTest/CMakeLists.txt
@@ -14,6 +14,7 @@ set(CppUTest_src
         MemoryLeakDetector.cpp
         TestFilter.cpp
         TestPlugin.cpp
+        TestTestingPlugin.cpp
         SimpleMutex.cpp
         Utest.cpp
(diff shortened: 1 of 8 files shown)


arstrube:
I have no objections to this solution, but could you please put the new files into sources.mk for Dos and see that the Dos build is indeed fixed. I am not sure about this, as the long long functions are prone to generate a lot of assembler code and I have no idea whether the string funcitons will outweigh them.

arstrube:
Needless to say that TestTestingFixture is cleaner this way :-)

basvodde:
It still needs to be cleaned up a bit though...

arstrube:
I'm kind of fond of my "hack" in the header though. I wouldn't have thought I could make the static functiosn work without a source file :P

basvodde:
I'd prefer moving everything to the source file anyway. It was already something that should have happened.

arstrube:
Can't really dispute that :D.

Are you going to make the changes needed to platforms/dos/sources.mk? Or how should we approach that? Should I submit a PR against basvodde/master?

basvodde:
Yes, will gradually :) Let me see the failure reasons.

arstrube:
For starters and obvious reasons, you need src/CppUTest/TestTestingFixture.cpp in the dos buid....

arstrube:
Same goes for Cmake builds. 

Interesting failures for Autotools, and also interesting that Cygwin succeded :-/

src/CppUTest/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(CppUTest_src
         MemoryLeakDetector.cpp
         TestFilter.cpp
         TestPlugin.cpp
-        TestTestingPlugin.cpp
```
arstrube: :-)

basvodde: /me an idiot not paying attention :)

arstrube:
The FAIL_LOCATION lines need to be commented with `// LCOV_EXCL_LINE` to make Coveralls happy

basvodde:
I'll merge this, the Vistual Studio files need fixing still though :(

Issue #954, opened 2016-05-02
FENV Test Failures under Visual Studio 2013/2015


Andne:
Visual Studio 2013 and newer contains the `fenv.h` header file as is able to compile the IEEE754 plugin.  However, something is different in the implementation and a couple of the tests fail.  During a previous pull request, the following were suggested as ways to handle the failure:
- disable the tests alltogether (not a good idea since most pass)
- remove tests that involves invalid (get rid of math.h as well)
- #ifdef the invalid tests away for Msvc

@basvodde Can you please weigh in on which of these you like best?

arstrube:
Thanks @Andne. Although I am not @basvodde, I'd cast my vote for removing tests for FE_INVALID entirely, including the C function that provokes it and math.h, which is only needed for that one function. That will get rid of both failures, and Msvc 2013 can then do valid tests.

I am supporting a large project where they do use the IEEE754 problem (implemented off CppUTest), and we did the same thing there (can't remember why, I think the needed to do a few tests with Msvc). I think in a real-world project, most of the time the output of the function that set FE_INVALID is likely to be input to some other operation, which causes one of the other flags to be set, so it isn't that the error wouldn't get caught. Floating point is Evil :D. That plugin always finds something somewhere ;-).

basvodde:
Removing the tests sounds ok, I guess. I got one failing test on Mac too.

Alternatively, we can leave them as IGNORE_TEST ?

arstrube:
@basvodde which one?

arstrube:
I think IGNORE_TEST wouldn't work, because one test fails indirectly because of the flag not being set, but that test isn't a test to be ignored.

arstrube:
@basvodde - no failure on my Mac (current head, Apple LLVM version 6.1.0)?

basvodde:
I had a failure on Mac on should_not_fail_again_when_test_has_already_failed.

arstrube:
How strange. Are you sure FE_INVALID didn't fail as well? Usually the two fail in concert.

arstrube:
Actually, had a look at should_not_fail_again_when_test_has_already_failed. That test is brittle (apart from the problem with Invalid). I try not to include fenv.h, but that comes at a price -- the hex values is not guaranteed to be the same across platforms. I wonder whether I could rewrite the test to make it more robust.

Andne:
[...]

PR #962, opened 2016-05-10
Refactor IEEE754 debugging functions

Andne:
As discussed in #954, I split the checkIeee754ExeptionFlag(int flag) function into specific functions for each flag.
This allows keeping the specifics of how the flags are defined in the class and avoid anything else depending on fenv.

### DIFF
--- a/include/CppUTestExt/IEEE754ExceptionsPlugin.h
+++ b/include/CppUTestExt/IEEE754ExceptionsPlugin.h
@@ -40,7 +40,10 @@ public:
 
     static void disableInexact(void);
     static void enableInexact(void);
-    static bool checkIeee754ExeptionFlag(int flag);
+    static bool checkIeee754OverflowExceptionFlag();
+    static bool checkIeee754UnderflowExceptionFlag();
+    static bool checkIeee754InexactExceptionFlag();
+    static bool checkIeee754DivByZeroExceptionFlag();
 
 private:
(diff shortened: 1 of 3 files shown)


arstrube:
Great! Thanks

arstrube:
How about having ` static bool checkIeee754InvalidtExceptionFlag()` as well and wrapping it in 
```
// LCOV_EXCL_START Not supported by all platforms
// LCOV_EXCL_STOP
```
It isn't strictly necessary, but would be nice to have for those who need it (similarly, the Plugin will detect this kind of error, if it is raised, but we don't test it anymore).

basvodde:
Seems ok like it is :)

arstrube:
I just updated github.io and I was wondering whether, since the user might now get exceptions for flag values not covered here, whether we shouldn't also offer:
```c++
static bool checkIeee754AnyExceptionFlag();
```
for cases where there is no specific checking function, or the user just doesn't care?

basvodde:
Not sure for me as I'm not a user of it :)

arstrube:
I think it would be important for those who are users ;-). The original interface with the ugly flag parameter could be used for any flag, including all flags or any combination thereof. Not that that in itself was all that useful. But right now, users are restricted to debugging just these 4 flags (because the remaining ones can differ between platforms. I will submit a PR when I have time....

PR #965, opened 2016-05-12
Add SimpleString::npos & use as default amount in subString()

arstrube:


basvodde:
Can we override the SubString with just one parameter instead, so we don't need the public npos static variable?

arstrube:
Public npos is gleaned from Standard C++. I find it rather useful (that was the whole point). Have a look at the next PR and let's then decide.

It needn't be public, of course, but I do find it usefu.....

arstrube:
From www.cplusplus.com:

public static member constant

std::string::npos

`static const size_t npos = -1;`

Maximum value for size_t
npos is a static member constant value with the **greatest possible value** for an element of type size_t.

This value, when used as the value for a len (or sublen) parameter in string's member functions, means "until the end of the string".

As a return value, it is usually used to indicate **no matches**.

This constant is defined with a value of -1, which because size_t is an unsigned integral type, it is the largest possible representable value for this type.

arstrube:
If you absolutely don't want **npos**, then we could use `size()` as default argument and still have a more convenient `subString()`, but when I raised the issue, I was really missing SimpleString::npos. I'd consider the default argument to subString rather as a nice-to-have.

basvodde:
This can be closed now, right?

arstrube:
yes

PR #966, opened 2016-05-13
Use size_t consistently with the aid of SimpleString::npos

arstrube:


arstrube:
The failure on Appveyor is due to Platform=MinGW32, PlatformToolset=5.3.0 failing to install.

arstrube:
@basvodde I do believe the use of `npos` is cleaner, at least internally. I would be open to remove it from the public interface, although I don't quite understand your objection, since std::string does it exactly the same way, and I fail to see what's 'bad' about it.

I think having a well defined value for "end of string" is an asset, not a liability?

tests/SimpleStringTest.cpp
```diff
@@ -405,7 +411,7 @@ TEST(SimpleString, findNormal)
     SimpleString str("Hello World");
     LONGS_EQUAL(0, str.find('H'));
     LONGS_EQUAL(1, str.find('e'));
-    LONGS_EQUAL(-1, str.find('!'));
+    LONGS_EQUAL(SimpleString::npos, str.find('!'));
```
arstrube: In my mind, this is a lot cleaner than returning an int where it should be size_t, and then comparing to -1. I find the "int approach" is old-school c-ish, not c++.

include/CppUTest/SimpleString.h
```diff
@@ -76,7 +78,7 @@ class SimpleString
     void replace(const char* to, const char* with);
 
     SimpleString lowerCase() const;
-    SimpleString subString(size_t beginPos, size_t amount) const;
+    SimpleString subString(size_t beginPos, size_t amount = SimpleString::npos) const;
```
basvodde: As mentioned, I'd rather just have 2 methods here than a default parameter

tests/SimpleStringTest.cpp
basvodde: I honestly don't care, and see very little benefit in changing it also.

arstrube: The benefit is using size_t where it should be used, and not mixing it with int (see further up in the code).

If you really don't want this, I'll do another PR just with the method.

I can also do
```c++
    SimpleString subString(size_t beginPos, size_t amount = size()) const;
```
Why would you prefer a second method?

basvodde: Ok, we can keep the npos, although I find npos a horrible name (consistent with Standard C++ library chosing horrible names). But I would like the additional method as IMHO it is just clearer in the interface to have the additional method.

Using the size_t consistently was a good argument.

arstrube: OK. Agreed. I will do the additional method.
Any suggestion for a better name?

basvodde: No suggestion for now. I guess npos will do as it is consistent with C++ lib. Though I don't like it :)

include/CppUTest/SimpleString.h
arstrube: On the way

arstrube:
The Appveyor failure is an choco problem.

PR #968, opened 2016-05-18
Add interface to compare signed byte-sized data

arstrube:
I noticed that CppUTest currently makes no provision for comparing signed byte-sized values. In my experience, in embedded development, sint8_t is frequently used.

This is an attempt at providing the missing functionality.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -232,6 +232,18 @@
 #define BYTES_EQUAL_TEXT(expected, actual, text)\
     LONGS_EQUAL_TEXT((expected) & 0xff, (actual) & 0xff, text)
 
+#define SIGNED_BYTES_EQUAL(expected, actual)\
+    SIGNED_BYTES_EQUAL_LOCATION(expected, actual, __FILE__, __LINE__)
+
+#define SIGNED_BYTES_EQUAL_LOCATION(expected, actual, file, line) \
+       { UtestShell::getCurrent()->assertSignedBytesEqual(expected, actual, NULL, file, line); }
+
+#define SIGNED_BYTES_EQUAL_TEXT(expected, actual, text)\
+    SIGNED_BYTES_EQUAL_TEXT_LOCATION(expected, actual, text, __FILE__, __LINE__)
+
+#define SIGNED_BYTES_EQUAL_TEXT_LOCATION(expected, actual, text, file, line) \
+        { UtestShell::getCurrent()->assertSignedBytesEqual(expected, actual, text, file, line); }
+
 #define POINTERS_EQUAL(expected, actual)\
     POINTERS_EQUAL_LOCATION((expected), (actual), NULL, __FILE__, __LINE__)
 
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -482,8 +482,17 @@ TEST(SimpleString, FunctionPointers)
 TEST(SimpleString, Characters)
 {
     SimpleString s(StringFrom('a'));
-    SimpleString s2(StringFrom('a'));
-    CHECK(s == s2);
+    STRCMP_EQUAL("a", s.asCharString());
+}
+
+TEST(SimpleString, NegativeSignedBytes)
+{
+    STRCMP_EQUAL("-15", StringFrom((signed char)-15).asCharString());
+}
+
+TEST(SimpleString, PositiveSignedBytes)
+{
+    STRCMP_EQUAL("4", StringFrom(4).asCharString());
 }
 
 TEST(SimpleString, LongInts)
@@ -554,6 +563,8 @@ TEST(SimpleString, Sizes)
 
 TEST(SimpleString, HexStrings)
 {
+    STRCMP_EQUAL("f3", HexStringFrom((signed char)-13).asCharString());
+
     SimpleString h1 = HexStringFrom(0xffffL);
     STRCMP_EQUAL("ffff", h1.asCharString());
 
--- a/include/CppUTest/SimpleString.h
+++ b/include/CppUTest/SimpleString.h
@@ -146,6 +146,7 @@ SimpleString StringFrom(long value);
 SimpleString StringFrom(unsigned long value);
 SimpleString StringFrom(cpputest_longlong value);
 SimpleString StringFrom(cpputest_ulonglong value);
+SimpleString HexStringFrom(signed char value);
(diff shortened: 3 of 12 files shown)


src/CppUTestExt/MockSupport_c.cpp
```diff
@@ -788,7 +788,7 @@ const void* returnConstPointerValueOrDefault_c(const void * defaultValue)
 
 void (*functionPointerReturnValue_c())()
 {
-    return actualCall->returnFunctionPointerValue();
+    return (void (*)()) actualCall->returnFunctionPointerValue();
```
arstrube: The cast has nothing to do with this PR - I checked the new code against the Cl2000 and it requires this cast. But it is also currently broken because of added features, and not easy to fix, so I can take this one out here, but there won't be a (working) PR for the Cl2000.

tests/SimpleStringTest.cpp
```diff
@@ -480,6 +480,13 @@ TEST(SimpleString, Characters)
     CHECK(s == s2);
 }
 
+TEST(SimpleString, CharactersAsDecimals)
+{
+    SimpleString s(DecimalStringFrom(-5));
+    SimpleString s2(DecimalStringFrom(-5));
+    CHECK(s == s2);
```
basvodde: This test doesn't really test much :)

basvodde:
Whats the difference between this and BYTES_EQUAL?

arstrube:
BYTES_EQUAL compares bytes as characters, e.g. 'a' or '¶'. 
CHARS_EQUAL

tests/SimpleStringTest.cpp
arstrube: It has many siblings :D

arstrube: In my opinion, we would do well to change all of them to something like:
```c++
SimpleString s(DecimalStringFrom(-5);
STRCMP_QUAL("-5", s.asCharString());
```
Any reason why these tests were written like this?

basvodde:
Why not have BYTES equal print the value, the hex *and* the character?

arstrube:
Why indeed not :) 
Sound like a much better solution to me.

arstrube:
My code doesn't do unsigned char correctly though. Will need to fix that as well.

tests/SimpleStringTest.cpp
basvodde: If you test the CHECK, it might be useful. But in this case, you test the DecimalStringFormat, then it isn't useful whatsoever :)

arstrube:
Actually, I got confused. Sorry about that. BYTES_EQUAL does unsigned. That's the problem.

The solution I'm trying to find is being able to do both signed and unsigned comparisons properly for byte-sized data.

arstrube:
@basvodde pls have another look. How could this still be improved?

arstrube:
I got some things mixed up in my comments. The current situation (before this PR) is:

1. `StringFrom(char)` will only print literal characters
2. `BYTES_EQUAL` will compare unsigned char, so signed values will be displayed wrong
3. The C interface will compare char as literal characters
4. Signed 8 bit data cannot be compared with proper failure messages

Normally, we have macros like "LONGS_EQUAL" vs. "UNSIGNED_LONGS_EQUAL.
Here, we have BYTES_EQUAL vs. SIGNED_BYTES_EQUAL, because BYTES_EQUAL was there first, and it does unsigned.

The C interface has two new macros: CHECK_EQUAL_C_UBYTE and CHECK_EQUAL_C_SBYTE. The C interface doesn't mirror all of the new macros we have for C++.

basvodde:
Uhm, but now the output is not "num hex char" like mentioned?

arstrube:
True. But it never included char, only in the C interface. So, if we do that, it will be like a new feature:

- do it only for the C interface? How?
- do it for BYTES_EQUAL (doesn't real make sense, it's not CHARS_EUAL)
- doesn't make sense for SIGNED_BYTES_EQUAL at all (this is numeric by definition

I suggest to give it some more thought and, perhaps, add it later. It would make sense for the C macro though, in which case there would be no new macros.

basvodde:
Uhm, well to me it is just 3 different representations of the same information. As sometimes showing it in a different representation will trigger an AH!, I think it is useful. I would do it for all of them, or perhaps only when it is in the printable ascii range :)

arstrube:
Hmm. So SIGNED_BYTES_EQUAL and BYTES_EQUAL would then display:
```
expected <-3 0xfd 'X'>
but was  <-2 0xfe 'Y'>

expected<255 0xff 'Z'>
but was  <253 0xfe 'Y'>
```
where X, Y, and Z would be the actual screen representation of that character. Similar for the C interface (will take a look next week).

basvodde:
Yes, and then we *could* check whether the character is printable or not, and not print it if it isn't.

arstrube:
I don't have much time right now, and if I do add the character, I will need to change the tests.

Since I want to test negative numbers and 'large' unsigned bytes, I would **have** to limit printing to printable characters for reasonable tests.

Can we perhaps merge this 'as is', and add printing the character later? (I can add an issue for that).

arstrube:
If I display the character along with numeric data, then 
```c++
   CHECK_EQUAL_C_CHAR()
```
which displays a literal character, would be redundant.
```+
   CHECK_EQUAL_C_UBYTE()
   CHECK_EQUAL_C_SBYTE()
```
would handle unsigned char and signed char, respectively. But we need to keep CHECK_EQUAL_C_CHAR(), because it's always been there.

Should we then do:
```c++
    CHECK_EQUAL_C_CHAR() // does signed char
    CHECK_EQUAL_C_BYTE() // does unsigned char
```
or leave all three (with CHECK_EQUAL_C_UBYTE() giving identical output to CHECK_EQUAL_C_CHAR()?

include/CppUTest/UtestMacros.h
```diff
@@ -232,6 +232,18 @@
 #define BYTES_EQUAL_TEXT(expected, actual, text)\
     LONGS_EQUAL_TEXT((expected) & 0xff, (actual) & 0xff, text)
```
basvodde: This one will need to be changed to one of the others. But you want to do that later, right?

arstrube: After the changes with the brackets, it could become UNSIGNED_LONGS_EQUAL_TEXT(). It would be easier to do this after the brackets thingie.

arstrube: yes.

PR #970, opened 2016-05-18
Cl2k

arstrube:
This is needed to get the Cl2k projects to compile and link again.

However, it results in a huge number of memory leaks and failed tests in the CppUTestExt binaries :-(


arstrube:
The memory leak thingie went away. So this can be merged.

This shall be my last contribution for the Cl2000 compiler. I am changing projects.

PR #973, opened 2016-05-19
ignoreAdditionalCalls (ignoreMultipleCalls?)

maxilai:
Hi, basvodde and arstrube
    My colleagues raised an requirements for "ignore additional (multiple, extra) calls".
    Here, we have ignoreOtherCalls which does not ignore the expected one.
    Then, we add a feature ignoreAdditionalCalls which behaviors like below. What I concern is the name "ignoreAdditionalCalls". Is this name a good one?
    Thanks!
    
TEST(MockCallTest, ignoreAdditionalCallsExceptForTheExpectedOne)
{
    mock().expectOneCall("foo");
    mock().ignoreAdditionalCalls();
    mock().actualCall("foo");

    /* The following calls will not fail, because the additional calls for expected is ignored. */
    mock().actualCall("foo");    
    mock().actualCall("foo").withParameter("foo", 1);
	
    mock().clear();
}

### DIFF
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -97,6 +97,7 @@ public:
     virtual void enable();
     virtual void tracing(bool enabled);
     virtual void ignoreOtherCalls();
+    virtual void ignoreAdditionalCalls();
 
     virtual void checkExpectations();
     virtual bool expectedCallsLeft();
@@ -133,6 +134,7 @@ private:
     MockExpectedCallsList expectations_;
     MockExpectedCallsList unExpectations_;
     bool ignoreOtherCalls_;
+    bool ignoreAdditionalCalls_;
     bool enabled_;
     MockCheckedActualCall *lastActualFunctionCall_;
     MockExpectedCallComposite compositeCalls_;
@@ -153,6 +155,7 @@ private:
 
     bool hasntExpectationWithName(const SimpleString& functionName);
     bool hasntUnexpectationWithName(const SimpleString& functionName);
+    bool hasntUnFulfilledWithName(const SimpleString& functionName);
     bool hasCallsOutOfOrder();
 
     SimpleString appendScopeToName(const SimpleString& functionName);
--- a/include/CppUTestExt/MockSupport_c.h
+++ b/include/CppUTestExt/MockSupport_c.h
@@ -189,6 +189,7 @@ struct SMockSupport_c
     void (*disable)(void);
     void (*enable)(void);
     void (*ignoreOtherCalls)(void);
+    void (*ignoreAdditionalCalls)(void);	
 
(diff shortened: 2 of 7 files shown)


basvodde:
I'm somewhat uncomfortable with this feature as I don't know in what case you would want to use it. Could you explain that to me? It feels it opens the door to making a lot weaker tests.

maxilai:
I'm not comfortable either.
The thing is. In they test cases, they call a mocked function many times.
They don't care how many times the calling happens. They just care
the behaviors of production code as expected based the returned value from
this mocked function.
In the production code, they will call this mocked function many times.
like below:
while (condition)
{
    ...
    importantValue = p->foo(parameters); // Here, the mocked funciton will
be called.
    ...
    otherImportantValue = whatEverFunctions();
    condition = caculatingCondition(importantValue, otherImportantValue);
}
Here, they just want to call the mocked function in order to get a
important value. But they don't care how many times this mock function are
called.
When they make some change from whatEverFunction() or any other related
functions. The times of happening of the mocked function will change. They
have to change the test code to revise the parameter amount from
expectNCalls method for lots of test cases. So , they raise this
requirement.

With ignoreOtherCalls and ignoreAdditionCalls, programmers will be able to
have cpputest to only check what they specified by expectedOne(N/No)Calls.
They want 5 calls, the cpputest will only check the first 5 calls, ignore
the 6th and 7th etc.

Or, do you guys have any good suggestion?





Thanks
Max MA

Chengdu Design Center, MSI

(86)28-8789-2373, (86)18602867063

2016-05-20 9:44 GMT+08:00 Bas Vodde <notifications@github.com>:

arstrube:
As a matter of interest, how do they determine pass / fail? I assume they have some way to check the value of `condition` in the test?

One thing I don't really understand is, if the first N calls influence the value of `condition`, how come the rest of the calls don't? This seems kind of strange to me.

Would it solve their problem if they set default return values for these mocks and then used `disable()` - enable()` to switch of mocking altogether, while the function under test is executed?

basvodde:
@maxilai It feels a bit like a hack to patch poorly written unit tests, so I'm very hesitant to add it. Even if we would add it, the way it is suggested now is generic so that it is for all expectations. That would make even less sense as you mentioned the ignoreAdditional calls ought to be per function.

I'd love to hear others opinion, but for me so far, it feels a bit wrong. Perhaps you could come up with an actual concrete test, so we could look at that?

maxilai:
Thanks for the advice from all of you.
I will contact with my colleagues to get more information about their test cases. Thanks again.

jgonzalezdr:
As a matter of fact, this behavior can be achieved with the functionality added in PR #1018, but in a more controlled way expectation by expectation instead of allowing all the expectations to get additional calls (which might be inadequate for some tests).

Issue #975, opened 2016-05-20
Why do some failures print the hex value twice?


arstrube:
An example would be:

```
      expected <97 (0x61) 0x61>
      but was  <98 (0x62) 0x62> 
```

Is there any point in printing it twice? This occurs, for example, in UNSIGNED_LONGS_EQUAL. Other failures just print the hex value once in brackets, which seems to be done elsewhere.

BYTES_EQUAL should actually call UNSIGNED_LONGS_EQUAL_LOCATION, instead of LONGS_EQUAL_LOCATION for clarity, but this is not possible, because of the strange output.

Or am I missing some finer point here?

cuitoldfish:
I don't think there is some finer point here, but some conflict to two class design here.

There is a general string class SimpleString existed in current design, for unsigned decimal number, it will be translate to "%llu (0x%llx)". All other will use SimpleString to construct string and so does the `UnsignedLongsEqualFailure`. But by Contrast to other client, `UnsignedLongsEqualFailure` has a exception -- it will append it's own hex format string which also follow the general design for TestFailure class.

so two class design for SimpleString and TestFailure make this print including hex format value twice, which seems stranger when showed as the failure report.

basvodde:
Eheh :) Needs to be fixed.

arstrube:
Okay. Will take a look at it next week.

maxilai:
Hi, arstrube
    I think the finer point here is "human readable", both dec and hex were
printed.
    example:

TEST(MockExpectedCallComposite, hasUnsignedLongIntParameter)
{
    composite.withParameter("param", (unsigned long int) 5);
    STRCMP_EQUAL("name -> unsigned long int param: <5 (0x5)>",
 call.callToString().asCharString());
}

3 types of value will be printed as dec and hex

1.
SimpleString StringFrom(cpputest_ulonglong value){
    return StringFromFormat("%llu (0x%llx)", value, value);
}
2.
SimpleString StringFrom(unsigned int i){
    return StringFromFormat("%10u (0x%08x)", i, i);
}
3.
SimpleString StringFrom(unsigned long i){
    return StringFromFormat("%lu (0x%lx)", i, i);
}


TEST(UnitTestMacros, TestUNSIGNED_LONGS_EQUAL)
{
    fixture.runTestWithMethod(_UNSIGNED_LONGS_EQUALTestMethod);
[...]

PR #979, opened 2016-05-22
#975 fix Move brackets format function out of StringFrom to keep general usage on StringFrom

cuitoldfish:


src/CppUTest/SimpleString.cpp
```diff
@@ -492,7 +513,7 @@ SimpleString StringFrom(cpputest_longlong value)
 
 SimpleString StringFrom(cpputest_ulonglong value)
 {
-    return StringFromFormat("%llu (0x%llx)", value, value);
+    return StringFromFormat("%llu", value, value);
```
basvodde: Fails here as you pass value twice to the StringFromFormat :)

cuitoldfish: will correct soon :)

arstrube:
@basvodde do we really need a bracketed hex string? I would prefer to get rid of it. Or, if we want bracketed hex strings in test failures, then add the brackets there (but personally, I feel they are redundant). I feel the few places where it was printed with brackets (plus without) were more of a slip than an intended feature.

@cuitoldfish if we get rid of the brackets, this PR will become a whole lot simpler.

cuitoldfish:
yes, but I'm hesitate if some modification will change what people has already currently accepted, just like this brackets format hex string.

I'm also curious why currently there is no brackets hex format for signed type(only for unsigned type)?

arstrube:
It's because the brackets were an inconsistancy to start with..... I feel they are not needed for anything, a hex number can easily be parsed looking for the 0x....

I was in fact wondering why we accepted tests like that, e.g. `<1 (0x1) 0x1>` when it was plain to see that all other tests were checking for `<1 0x1>`.

basvodde:
I'd somehow prefer the 1 (0x01) format for all. The brackets make it clearer that it is a representation of the same format.

arstrube:
@basvodde that would work, too.... would be still need the extra set of methods though? I would prefer to see the adding of brackets outside of SimpleString. I think SimpleString should do the simple thing :).

basvodde:
We passed the simple part long time ago :)

I think its ok in SimpleString as it would be at least in one place. Otherwise the same format would be in multiple places, which is how we ended up with the problem in the first place.

tests/TestUTestMacro.cpp
```diff
@@ -77,8 +77,8 @@ static void _UNSIGNED_LONGS_EQUALTestMethod()
 TEST(UnitTestMacros, TestUNSIGNED_LONGS_EQUAL)
 {
     fixture.runTestWithMethod(_UNSIGNED_LONGS_EQUALTestMethod);
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("expected <1 (0x1) 0x1>");
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <0 (0x0) 0x0>");
+    CHECK_TEST_FAILS_PROPER_WITH_TEXT("expected <1 0x1>");
+    CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <0 0x0>");
```
arstrube: @basvodde if we do the brackets thing, these need to be bracketed, too

arstrube: and I would have to make the same change to SIGNED_BYTES_EQAUL & friends.

basvodde: Correct

arstrube:
@basvodde okay you convinced me there ;-)

tests/TestUTestMacro.cpp
arstrube: This one will then need to be merged before my signed bytes one.

cuitoldfish:
how about cpputest_longlong type,  I didn't create a brackets format function for it because I didn't see where this type is printed in a string.

cuitoldfish:
So for 
```.cpp
CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <0 (0x0) 0x0>");
```
will be changed to 
```.cpp
CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <0 (0x0)>");
```
?

and for signed type, take -1 as an example,  it's hex string will be 0xFFFFFFFF, so in the above case, that would be
```.cpp
CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <-1 (0xFFFFFFFF)>");
```
still looks good?

arstrube:
yes, looks good to me :)

cuitoldfish:
if the brackets hex format string is part of all numeric type string, then I prefer to add this just inside StringFrom,

arstrube:
yes I think that's the idea, just the methods you already added.

arstrube:
umm, no, not in StringFrom(). That is used in many other places, and we don't want brackets there!

arstrube:
maybe I'm wrong. Try what happens if you put it into StringFrom, if any tests fail, etc. I am not a great fan of the brackets, but @basvodde likes them. Perhaps it is okay in StringFrom, although I would have preferred a 'normal' string format. Also, should now suddenly pointers be in brakcets, when it is the main displayed value? I think not....

arstrube:
I think your PR is fine with regard to SimpleString. Just need to use brackets consistently for all failures and in all tests, or? (persoanlly I like it better as is, though)...

basvodde:
Not in the StringFrom, that would be a mistake :)

PR #986, opened 2016-05-24
#975 fix  Add brackets hex string for tests and failure

cuitoldfish:
seems some conflict existed, will take a look later


tests/CppUTestExt/MockFailureTest.cpp
```diff
@@ -205,9 +205,9 @@ TEST(MockFailureTest, MockExpectedParameterDidntHappenFailure)
     MockExpectedParameterDidntHappenFailure failure(UtestShell::getCurrent(), "foo", *list);
     STRCMP_EQUAL("Mock Failure: Expected parameter for function \"foo\" did not happen.\n"
                  "\tEXPECTED calls that DID NOT happen related to function: foo\n"
-                 "\t\tfoo -> int bar: <2>, const char* boo: <str>\n"
+                 "\t\tfoo -> int bar: <2 (0x00000002)>, const char* boo: <str>\n"
                  "\tACTUAL calls that DID happen related to function: foo\n"
-                 "\t\tfoo -> int bar: <10>, const char* boo: <bleh>\n"
+                 "\t\tfoo -> int bar: <10 (0x0000000a)>, const char* boo: <bleh>\n"
```
arstrube: If this is really how CppUTest displays the value, then we have another inconsistency, because elsewhere small values would be displayed like, e.g. (0x1) or (0xa)

cuitoldfish: you mean the inconsistency between (0xa) and (0x0000000a) ?

arstrube: yes

arstrube: I noticed a while ago that it normally does 0xa.

cuitoldfish: Then there is another concern that, should all the formate follow the (0x0000000a)  or (0xa)?

Seems we get into endless option selecting

basvodde: I'd say 0xa as it feels much more clear :)

arstrube: Yes, please stick with 0xa. You can get into untold trouble doing 0x0000000a on different bit width systems....... 0xa meaning, only non-zero nibbles are displayed. Because it can be really tough to figure out now many 0 there should be and then testing that on different platforms (bad enough we have that problem with BITS_EQUAL.....)

basvodde:
Open close?

Issue #957, opened 2016-05-08
GCC 6: Missing sized deallocation functions


offa:
Compiling with GCC 6 (Cmake, C++11 enabled) emits following warning:

```
[…]/cpputest/src/CppUTest/MemoryLeakWarningPlugin.cpp:312:6: Warnung: the program should also define »void operator delete(void*, std::size_t)« [-Wsized-deallocation]
 void operator delete(void* mem) UT_NOTHROW
      ^~~~~~~~
[…]/cpputest/src/CppUTest/MemoryLeakWarningPlugin.cpp:332:6: Warnung: the program should also define »void operator delete [](void*, std::size_t)« [-Wsized-deallocation]
 void operator delete[](void* mem) UT_NOTHROW
      ^~~~~~~~
```

About this [warning](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html):

> -Wsized-deallocation (C++ and Objective-C++ only)
>     Warn about a definition of an unsized deallocation function
> 
>               void operator delete (void *) noexcept;
>               void operator delete[] (void *) noexcept;
> 
>     without a definition of the corresponding sized deallocation function
> 
>               void operator delete (void *, std::size_t) noexcept;
>               void operator delete[] (void *, std::size_t) noexcept;
> 
>     or vice versa. Enabled by -Wextra along with -fsized-deallocation. 

---------------------------------------------
**Update:** Same with C++11 disabled.

arstrube:
It seems that Gcc 6 makes `-fsized-deallocation` the default.

How about we explicitly specify `-fno-sized-deallocation` in the compiler options?

Of course, we could add the functions as well, but that seems to make little sense, in that we weren't intending to use them, anyway.

arstrube:
The option should only be in CXXFLAGS, not in CFLAGS.

offa:
I don't know if there any negative impacts, but I would rather solve the problem by adding the necessary functions than solving it by disabling the warning.

Unfortunately the documentation above doesn't list the rational behind the warning, but I guess the GCC dev's had an intention by adding this.

arstrube:
It's not like disabling a warning - it's telling Gcc we are not interested in this feature :)

basvodde:
I think we could add them, the problem of these is that compiler and compiler versions all deal with it differently. So adding it might not be as simple as it seems :)

offa:
Fixed, see #987

PR #987, opened 2016-05-24
Included operator delete with size for new gcc

basvodde:
As title.


arstrube:
nice.

arstrube:
ummm... tests? Coverage decereased by a whopping 0.07 percent ^.^

offa:
I have just tested on Gcc 6; all fine so far, the warning (#957) didn't show up.

basvodde:
Added now. Wanted to check if it compiles at all on all our compilers. Lets see about the test as the syntax for calling it is somewhat unusual :)

PR #988, opened 2016-05-25
#975 fix  Add consistent brackets hex string format for all test and failure

cuitoldfish:


tests/SimpleStringTest.cpp
```diff
@@ -742,12 +742,21 @@ IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
...
-TEST(SimpleString, BuildStringFromUnsignedInteger)
+TEST(SimpleString, BuildStringFromUnsignedLongInteger)
 {
     unsigned long int i = 0xffffffff;
 
     SimpleString result = StringFrom(i);
-    const char* expected_string = "4294967295 (0xffffffff)";
+    const char* expected_string = "4294967295";
+    CHECK_EQUAL(expected_string, result);
+}
```
arstrube: why?

arstrube:
I see the hex strings are giving you some grief... shouldn't it be possible to print them (and test them) exactly as always, just with brackets around them?

arstrube:
I mean, if it worked before, it should work the same way now, unless you change something (other than the brackets)?

src/CppUTest/SimpleString.cpp
```diff
@@ -478,6 +478,11 @@ SimpleString HexStringFrom(long value)
...
 }
 
+SimpleString HexStringFrom(int value)
+{
+    return StringFromFormat("%x", value);
+}
+
 SimpleString HexStringFrom(signed char value)
 {
     SimpleString result = StringFromFormat("%x", value);
```
arstrube: I don't know how this happened (mismerge?), but I think you want that instead:
```c++
SimpleString HexStringFrom(signed char value)
{
    SimpleString result = StringFromFormat("%x", value);
    if(value < 0) {
        size_t size = result.size();
        result = result.subString(size-(CPPUTEST_CHAR_BIT/4));
    }
    return result;
}
```

arstrube:
I just had a closer look - I think the merge went wrong. Sorry about that; my idea was, for your changes to be applied first and then me rebasing my signed char stuff on top of it, that would have been easier.

src/CppUTest/SimpleString.cpp
arstrube: You won't notice the difference on TravisCI / Appveyor, but it will blow up the Cl2000 because it has no 8 bit width data. So char is 16 bit.

tests/SimpleStringTest.cpp
```diff
@@ -1028,3 +1037,59 @@ TEST(SimpleString, StringFromOrdinalNumberOthers)
...
+	STRCMP_EQUAL("(0x1)", BracketsFormattedHexStringFrom(value).asCharString());
+}
+
+TEST(SimpleString, BracketsFormattedHexStringFromForUnsignedLong)
+{
+	unsigned long value = 1;
+
+	STRCMP_EQUAL("(0x1)", BracketsFormattedHexStringFrom(value).asCharString());
+}
+
```
arstrube: signed char?

tests/TestUTestMacro.cpp
```diff
@@ -546,8 +546,8 @@ static void _failingTestMethodWithBYTES_EQUAL()
 TEST(UnitTestMacros, FailureWithBYTES_EQUAL)
 {
     fixture.runTestWithMethod(_failingTestMethodWithBYTES_EQUAL);
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("expected <97 0x61>");
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <98 0x62>");
+    CHECK_TEST_FAILS_PROPER_WITH_TEXT("expected <97 (0x61)>");
+    CHECK_TEST_FAILS_PROPER_WITH_TEXT("but was  <98 (0x62)>");
```
arstrube: Can't find the SIGNED_BYTES tests anymore? They should have required the bracket change as weill?

arstrube: Or perhaps not merged yet?

src/CppUTest/SimpleString.cpp
cuitoldfish: yes, seems mismerge happened, git didn't give me a conflict report, it just auto-merge and then mis-merged

tests/TestUTestMacro.cpp
cuitoldfish: @arstrube , Do you mean SIGNED_BYTES tests missed or I need to add brackets for SIGNED_BYTES too?  If I understand correctly, you decide to add brackets hex format for SIGNED_BYTE later?

src/CppUTest/SimpleString.cpp
cuitoldfish: @arstrube ,  I checked the merge and it went well. The above codes slices only display the difference, and it don't display the whole codes, the codes
```.cpp
SimpleString HexStringFrom(signed char value)
{
    SimpleString result = StringFromFormat("%x", value);
    if(value < 0) {
        size_t size = result.size();
        result = result.subString(size-(CPPUTEST_CHAR_BIT/4));
    }
    return result;
}
```
still there. 

You can fetch my branch and do a diff to see :)

cuitoldfish:
>I see the hex strings are giving you some grief... shouldn't it be possible to print them (and test them) >exactly as always, just with brackets around them?

Not sure if I understand your comments correctly :| , below is my grief:

When I add hex string format for long, I begin to remember the width of long is different on  different compilation system, so take -1 as an example, it will have two possible output "(0xffffffff)" and "(0xffffffffffffffff)".   
it is, for below test case, 
```.cpp
TEST(MockExpectedCallComposite, hasLongIntParameter)
{
    composite.withParameter("param", (long int) -1);
    STRCMP_EQUAL("name -> long int param: <-1>", call.callToString().asCharString());
}
```
no matter I use "name -> long int param: <-1 (0xffffffff)>" or "name -> long int param: <-1 (0xffffffffffffffff)>", I will always get CI failed.
I think this test should only focus on to check if mockExpectedCall will print correct param and name string according to withName and withParameter function, but not the width of the parameter, so I changed -1 to 1 to keep this test focus on single check point.

and I have been thought to write tests to check if brackets hex string will have correct output on different compilation system, and if those test complain something I then adjust some implementation, I plan to add later in this PR, but we have to discuss how to print brackets hex string for all test and failure firstly.

tests/TestUTestMacro.cpp
arstrube: Both :-).

arstrube: I wanted to add it after yours was merged, but since @basvodde merged mine first, it was already there (mismerge?) and now yours is the privilege of fixing it up :-)

arstrube: and adding the brakcets stuff to it after the fact. Not nice, but what can I do....

src/CppUTest/SimpleString.cpp
arstrube: ah okay. That's good. sorry in that case.

arstrube:
Yeah, not doing -1 seems a good idea. I think we have tests for that elsewhere, and they involve some ugly #ifdefing.....

tests/TestUTestMacro.cpp
cuitoldfish: at least from the log of cpputest/master, there isn't any brackets hex format for signed char :)

cuitoldfish:
I also add extra test for bracket hex format for long in different platform, I want to make sure the output is correct in different platform, will push it soon.

The only concern is I do not have different platform, only one 32-bit system on my hand, so I have to check if my tests pass on Travis CI

tests/TestUTestMacro.cpp
arstrube: of course not! I was going to put them in when your brakcets stuff was merged. But now mine is merged already so YOU must add the brackets for me, please. Everything needs brackets now that it has been decided. My stuff too. And then my tests should be changed so they are looking for brackets.

arstrube: @basvodde normally I can't wait for my stuff to be merged. But next time there's a situation like this one I'll be sure to put in some stupid compiler error so it can't be merged :-P

arstrube:
@cuitoldfish that's okay. I won't be able to test for the cl2000 compiler anymore, which is a pretty unusual one, and we do have the Dos build for 16 bit in general, so you should be okay.

As for platforms, the one to watch out for is Windows 64 bit (which is covered by Appveyor) because it doesn't have 64 bit longs, it has 32 bit. I think you will see existing tests which are #ifdefd for that reason. You should find all you need in existing code / tests somewhere, because these kinds of problems have occurred before.....

arstrube:
Mmm. Sorry for the trouble. For instance, the failure of the Dos build is a problem with testing. The output of CppUTest is correct - ints are 16 bit on that platform, so it prints (0xffff), not (0xffffffff). I guess you must be ready to tear out your hair by now :-(.

tests/TestUTestMacro.cpp
cuitoldfish: well, now I know what's your meaning. I see there is HexStringFrom for signed char, and I need to add bracketsFormattedHesStringFrom for signed char, and updated relative test failure and tests.

cuitoldfish:
yes, I find that macro `CPPUTEST_64BIT_32BIT_LONGS`, and already watch out this, thanks for the reminder, please review my codes change in SimpleStringTest. But because my slow network speed, I can't know the immediately result of CI.

cuitoldfish:
seems currently there is no macro definition to detect the different width for int?

arstrube:
I remember in some places I 'cheated' when I added the Dos build - meaning I changed values for testing that to <= 0xffff, so they would work on all platforms. Perhaps it would be good to add a macro in CppUTestConfig.h..... to do it properly....

basvodde:
Almost there, right?

cuitoldfish:
@basvodde , yes, the only barrier is the different width output for int, we need to detect different width for int on different platform, plan to add new macro.

cuitoldfish:
@arstrube could you give some advice if we add macro to detect 16 bit int on wcl compiler? I have limited knowledge about the dos environment.

cuitoldfish:
seems this method can not work
```.cpp
#if CPPUTEST_USE_STD_C_LIB
#if (INT_MIN == 0xffff) && (INT_MAX == 0x7fff)
#define CPPUTEST_INT_BIT 2
#else
#define CPPUTEST_INT_BIT 4
#endif
#else
#define CPPUTEST_INT_BIT 4
#endif
```

arstrube:
Perhaps ask @basvodde on this one? `int` can be anything, potentially, from 16 bit to 64 bit. So it feels a bit arbitrary to ignore the possibility of 64 bit, but then again, if it works....... as I said before, in the past I simply used suitable values such as 13556 but it didn't feel right to ignore negative values for testing......

cuitoldfish:
how about this:
```.cpp
+#if CPPUTEST_USE_STD_C_LIB
 +#if (INT_MAX == 0x7fff)
 +#define CPPUTEST_INT_BIT 2
 +#else
 +#define CPPUTEST_INT_BIT 4
 +#endif
 +#else
 +#define CPPUTEST_INT_BIT 4
 +#endif
```

at least this makes the CI happy, and it doesn't cover the 64bit integer case.

@basvodde I think the PR is ready to be merged, want to know if you have other comments

include/CppUTest/CppUTestConfig.h
```diff
@@ -201,6 +201,18 @@
 #define CPPUTEST_CHAR_BIT 8
 #endif
 
+/* Handling of systems with a different int-width (e.g. 16 bit).
+ */
+#if CPPUTEST_USE_STD_C_LIB
+#if (INT_MAX == 0x7fff)
+#define CPPUTEST_INT_BIT 2
```
basvodde: How about calling this CPPUTEST_INT_SIZE ? Bit is a bit confusing.

Can you also add a test that does sizeof(int) == CPPUTEST_INT_SIZE as a sanity check?

cuitoldfish:
I give that name because CHAR_BIT and INT_BIT are consistent with the naming rule in limits.h in ASNI C. So if we rename INT_BIT to INT_SIZE then it's best to also rename CHAR_BIT to CHAR_SIZE, then this begin to be inconsistent with the naming rule in limits.h.

I prefer to keep consistent with ANSI C since CppUTest is a C/C++ test framework, what's your opinion?

arstrube:
I prefer CPPUTEST_INT_BIT but, I just realize, it has to be 16 vs. 32 to be consistent! You can't give it a value of 2. There is no platform in the whole universe that has 2 bits per int... :P

arstrube:
Also note that on platforms with CHAR_BIT == 16, the size of a 16 bit integer will be 1.

arstrube:
So this is not a value to rely on.

arstrube:
`#define CPPUTEST_INT_BIT INT_BIT`

I doubt we need this, though, because I don't see a reason to do calculations with it (yet) .

Why not just define CPPUTEST_16BIT_INTS, and then `#if #else` the tests based on that, similar to CPPUTEST_64BIT_32BIT_LONGS?

arstrube:
@basvodde the sanity check (and other things) would fail on the cl2000, because sizeof(int) == 1 != 2

arstrube:
Tests would then be something like:
```c++
#ifdef CPPUTEST_16BIT_INTS
TEST(SimpleString, BracketsFormattedHexStringFromForInt)
 {
 	STRCMP_EQUAL("(0xffff)", BracketsFormattedHexStringFrom((int)-1).asCharString());
 }
#else
 TEST(SimpleString, BracketsFormattedHexStringFromForInt)
{
 	STRCMP_EQUAL("(0xffffffff)", BracketsFormattedHexStringFrom((int)-1).asCharString());
 }
#endif

**Comment by cuitoldfish** [comment]
@arstrube , I made a mistake to set 2 or 4 to INT_BIT, it should be 16 or 32. 

But in my macro definition, size of integer rely on the value of INT_MAX, not CHAR_BIT, so I think it's ok no matter what the value of CHAR_BIT is.

it's another possible solution to define macro similar to `CPPUTEST_64BIT_32BIT_LONGS`, but I do not know what compiler parameter (like _WIN64) to rely on to define CPPUTEST_16BIT_INTS, could you give your guidance?

**Comment by arstrube** [comment]
Personally, I would base it on the size, like you did, but you could also define it based on `__WATCOMC__`

**Comment by arstrube** [comment]
The size being more generic, this depends on the platform rather than the compiler, and wouldn't necessarily be Dos, either.

**Comment by cuitoldfish** [comment]
is `__WATCOMC__` or relying on INT_MAX in limits.h more general solution to detect 16 bits integer?

**Comment by arstrube** [comment]
I thought INT_MAX in limits.h is, but am late in answering, and perhaps it's not that important.

**Comment by cuitoldfish** [comment]
then we can go with current __WATCOMC__ and CPPUTEST_16BIT_INTS way to detect 16bit int.

Anymore comments? Now this PR has passed CI check and I think it's ready to be merged

**Comment by cuitoldfish** [comment]
@basvodde I think the sanity check for the integer size has been done:

```.cpp
#ifdef CPPUTEST_16BIT_INTS
TEST(SimpleString, BracketsFormattedHexStringFromForInt)
{
	int value = -1;

	STRCMP_EQUAL("(0xffff)", BracketsFormattedHexStringFrom(value).asCharString());
}
#else
TEST(SimpleString, BracketsFormattedHexStringFromForInt)
{
	int value = -1;
	STRCMP_EQUAL("(0xffffffff)", BracketsFormattedHexStringFrom(value).asCharString());
}
#endif
```
In each TEST, if size of integer goes exception, the TEST will fail.

include/CppUTest/CppUTestConfig.h
```diff
@@ -201,10 +201,15 @@
 #define CPPUTEST_CHAR_BIT 8
 #endif
 
+/* Handling of systems with a different int-width (e.g. 16 bit).
+ */
+#if defined(__WATCOMC__)
+#define CPPUTEST_16BIT_INTS
+#endif
```
basvodde: I'd prefer the previous solution and not having Watcom checks in this header (when we can avoid them)

include/CppUTest/CppUTestConfig.h
```diff
@@ -201,10 +201,15 @@
...
 /*
  * Support for "long long" type.
  *
- * Not supported when CPUTEST_LONG_LONG_DISABLED is set.
+ * No-=-  =-9----------------- =99t supported when CPUTEST_LONG_LONG_DISABLED is set.
```
basvodde: Whats this? :)

basvodde:
Ok, I'll merge it. Can you do a new pull request to fix the comment?

cuitoldfish:
off course, that seems a strange modification and sorry for that

PR #989, opened 2016-05-25
Overlooked this one in my npos-2 PR

arstrube:


basvodde:
Anything we can do with the randomly failing mingw? Can we put a retry in the build file?

arstrube:
I don't know. You can restart Appveyor. I close and reopen the PR to force it to restart until it passes. Not elegant :P

basvodde:
No, but it always fails at the same time, so we ought to be able to put that in a loop and try multiple times?

arstrube:
I guess so...

PR #992, opened 2016-05-25
C++14 support in CI build

basvodde:


offa:
Shouldn't `-std=c++14` be checked to?

arstrube:
Me thinks @offa is right. The C++14 build does not have `-std=c++14` set.

arstrube:
In my opinion, it would be better to not have the functions and compile with -fno-sized-deallocation until such time that we actual want or need it. This amounts to a simple, clear statement that we are not interested in this functionality right now. We don't then have to compile for C++14, either.

basvodde:
@offa No, it doesn't set -std=c++14 as it sets -std=c++1y, which I believe was the right flag for std C++14. But let me see if the other one works too.

basvodde:
@arstrube I think compiling C++14 is good to ensure compatibility and prefer not turning off the warning. If we do compile C++14, we can safely turn off the C++14 compatibility warning though as we know we are C++14 compatible.

arstrube:
Please correct me if I'm wrong, but I looked at the log of the last Travis run, and in the section where it said it was compiling C++14, g++ was called without -std switch, not with -std=C++14.

Strangely, in this latest log, it seems that AutoConfig always comes back saying that C++14 is not supported by the compiler (both Clang and Gcc) and there is no section for C++14 anymore?

arstrube:
It says no to c++1y as well, and strangely, to c++11 ???

basvodde:
In the travis job, it compiled c++1y but it doesn't recognize c++14. On my local computer, it recognizes both. The compilation seems to work, so I'll merge this.

arstrube:
did I look at the wrong job, perhaps? Which one is the one that compiled?

basvodde:
The first one, just search for C++14. I outputted the language variant at the end of the configure script now.

arstrube:
ahh, I see.

offa:
>In the travis job, it compiled c++1y but it doesn't recognize c++14. On my local computer, it recognizes both. The compilation seems to work, so I'll merge this.

Older compiler don't know the `-std=c++14` since C++14 wasn't released at that time.

PR #997, opened 2016-06-03
Fix handling of -Wno-c++14-compat

arstrube:
This only takes care of automake. It has not been tested against Gcc 5.3.x.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -133,7 +133,7 @@ CXXFLAGS="$saved_cxxflags"
 # Flag -Wno-c++14-compat
 CXXFLAGS="-Werror -Wno-c++14-compat"
 AC_MSG_CHECKING([whether CXX supports -Wno-c++14-compat])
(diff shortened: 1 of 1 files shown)


arstrube:
This only takes care of automake. It has not been tested against Gcc 5.3.x.

basvodde:
Is this really needed?

arstrube:
Definitely on Cygwin 2.5 with Gcc 5.3. But I am not in a position anymore to verify that this fix works.

By the way, please have a look at the actual changes. Do you have any explanation why this should be in   CPPUTEST_CPP11BLACKLIST rather than CPPUTEST_CXXWARNINGFLAGS? Does that make sense at all (regardless of the problem this is supposed to fix)?

basvodde:
Yes, as the warning *should* only happen when compiling with C++11 support on, or so I thought.

PR #999, opened 2016-06-08
Add CHECK_EQUAL_C_LONGLONG to C TestHarness

uecasm:
And the other int flavours for good measure.

This will have a trivial conflict with #998 if that's merged first but any good merge tool will resolve it automatically.  (Or remind me after merging that and I'll rebase this.)


uecasm:
The AppVeyor failure just looks like the CI being broken, not a code fault.

basvodde:
Thanks

Issue #1004, opened 2016-06-15
CppUTestExt tests failing on multiple architectures


rhertzog:
Hello, I uploaded version 3.8 to Debian and unfortunately the package did not build on all architectures since the test suite is failing on many of them:
https://buildd.debian.org/status/package.php?p=cpputest

The errors are not the same on all architectures. Those I care the most about are armel and s390x:
https://buildd.debian.org/status/fetch.php?pkg=cpputest&arch=s390x&ver=3.8-2&stamp=1465996058
https://buildd.debian.org/status/fetch.php?pkg=cpputest&arch=armel&ver=3.8-2&stamp=1465996758

On s390x you get this:
```
tests/CppUTestExt/MockSupport_cTest.cpp:194: error: Failure in TEST(MockSupport_c, outputParametersOfType)
	expected <2 0x2>
	but was  <1 0x1>
```
On armel you get this:
```
tests/CppUTestExt/IEEE754PluginTest.cpp:109: error: Failure in TEST(FE__with_Plugin, should_fail_only_once_when_all_flags_are_set)
	expected <1 0x1>
	but was  <0 0x0>

....
tests/CppUTestExt/IEEE754PluginTest.cpp:72: error: Failure in TEST(FE__with_Plugin, should_fail____when__FE_INEXACT____is_set_and_enabled)
./include/CppUTest/TestTestingFixture.h:123: error:
	actual <.
OK (1 tests, 1 ran, 5 checks, 0 ignored, 0 filtered out, 0 ms)

>
	did not contain  <IEEE754_CHECK_CLEAR(FE_INEXACT) failed>

.
tests/CppUTestExt/IEEE754PluginTest.cpp:65: error: Failure in TEST(FE__with_Plugin, should_fail____when__FE_UNDERFLOW__is_set)
./include/CppUTest/TestTestingFixture.h:123: error:
	actual <.
OK (1 tests, 1 ran, 5 checks, 0 ignored, 0 filtered out, 0 ms)

>
	did not contain  <IEEE754_CHECK_CLEAR(FE_UNDERFLOW) failed>

.
tests/CppUTestExt/IEEE754PluginTest.cpp:58: error: Failure in TEST(FE__with_Plugin, should_fail____when__FE_OVERFLOW___is_set)
./include/CppUTest/TestTestingFixture.h:123: error:
	actual <.
OK (1 tests, 1 ran, 5 checks, 0 ignored, 0 filtered out, 0 ms)

>
[...]

PR #1005, opened 2016-06-16
Fix for Debian Armel failure (#1004)

arstrube:


arstrube:
Superceeded by #1006.

PR #1007, opened 2016-06-19
Readme updated

offa:
Some updates for the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,7 @@ The failure of one of these macros causes the current test to immediately exit
 
 Customize CHECK_EQUAL to work with your types that support operator==()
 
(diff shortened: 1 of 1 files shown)


basvodde:
Then you could make it even simpler by using the default main macro that is provided by CppUTest :)

offa:
@basvodde I didn't know, there's a default main macro. This should be added to the documentation too.

arstrube:
@offa he sneaked it in recently without anybody noticing :-)

offa:
That's true!

basvodde:
Uhm, well, I didn't consider it a major feature. Just a convenience ;)

PR #1008, opened 2016-06-22
Bug fix for wrong output parameters being copied

jgonzalezdr:
It's been a long time since my last contributions, but I've finally managed to get some free time and I've spend it in finishing the modifications that I started to enable expected calls to match multiple calls. I'm still merging all the changes of the main branch into my outdated code (almost a year full of commits!), but I've detected a bug that I've just fixed, so I'm submitting this standalone PR for it.

Here goes the bug description and fixing notes:

The present code can copy the output parameters of an expected call, but finally fulfill a different one.

This happens for example if an expected call for a function that returns output parameters and ignores other parameters is declared before an expected call for the same function, that also returns expected parameters, but that doesn't ignore parameters.

The solution has been to just prioritize output parameter finalization of expected calls that don't ignore parameters over calls that can ignore them (in the same way that call fulfillment works), instead of just picking the first one.

BTW, some optimizations have been also applied on the affected code, to avoid traversing the expected calls list twice (first to check if there is a call that matches some criteria, then again to get that call), and therefore some methods that are not necessary any more have been deleted.

### DIFF
--- a/include/CppUTestExt/MockExpectedCallsList.h
+++ b/include/CppUTestExt/MockExpectedCallsList.h
@@ -43,8 +43,6 @@ public:
     virtual int amountOfExpectationsFor(const SimpleString& name) const;
     virtual int amountOfUnfulfilledExpectations() const;
     virtual bool hasUnfulfilledExpectations() const;
-    virtual bool hasFulfilledExpectations() const;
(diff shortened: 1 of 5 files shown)


arstrube:
The appveyor error is a random failure.

PR #1009, opened 2016-06-28
Disable EXPECT_N_LEAKS() if memleak detection disabled

bkylerussell:
EXPECT_N_LEAKS() can't be supported when CppUTest is configured with --disable-memory-leak-detection, since there's nothing to make sure we hit the expected number of leaks.

In one sense, it would be nice for tests that use EXPECT_N_LEAKS() to go ahead and continue running with whatever capabilities CppUTest is configured with.  This change would allow that behavior.

But on the other hand, I guess someone that really expected leaks might want to see the failure to know that CppUTest wasn't configured like they expected.

Is the intended behavior for tests using EXPECT_N_LEAKS() to fail when CppUTest is configured with --disable-memory-leak-detection, or was this just a miss? Thoughts?

### DIFF
--- a/src/CppUTest/MemoryLeakWarningPlugin.cpp
+++ b/src/CppUTest/MemoryLeakWarningPlugin.cpp
@@ -570,8 +570,15 @@ void MemoryLeakWarningPlugin::postTestAction(UtestShell& test, TestResult& resul
     int leaks = memLeakDetector_->totalMemoryLeaks(mem_leak_period_checking);
 
     if (!ignoreAllWarnings_ && expectedLeaks_ != leaks && failureCount_ == result.getFailureCount()) {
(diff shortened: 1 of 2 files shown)


basvodde:
The failure is not intended, but it would sort-of make sense to fail, or at least print a warning.

However, the fix isn't ok. You'd need to work on the check, and if possible, do it runtime and not compile time. Also there is no test provided. So, as is, I won't accept this pull request.

bkylerussell:
Ok, thanks.  I really just wanted to get feedback on whether or not a failure was intended.  I can fix this up if you think printing a warning would be acceptable while allowing other tests to continue running.

bkylerussell:
I think I've fixed this up per the previous discussion so that a warning is printed and the change is accompanied by a test.  Checks seem to be happy now.

Issue #967, opened 2016-05-13
Get test case location in source via CLI


itavero:
We are working on an initial version of a CppUTest Test Runner for Visual Studio.
So far, we are able to detect which tests are available from the compiled binary using the `-ln` flag.
We can run selected tests and process the JUnit XML output for displaying the results, errors and execution time.

The next step would be to be able to jump directly to the test case in the source code from the Test Explorer in Visual Studio.
I tried using the information from the JUnit XML output, so we could at least jump to failed assertions.
However, Visual Studio does not process any changes to the test case after discovery (I can only publish new test results, which basically is OK / Failed with an additional string... rather limited).
So, we need this information during initial test case discovery, for which we currently use the `-ln` flag.

As far as I can tell, there currently isn't a way to get this information via the command-line interface.
Perhaps I could decompile the binary or do some other smart analysis, but that all feels like "hacking" to me.

I'd suggest adding a more elaborate interface (perhaps also with XML or JSON output to allow for easier parsing) that exposes the following information (XML-ish example):
````xml
<testrunner>
  <group name="SomeTestGroup">
    <testcase name="FirstTest" source="C:\Path\To\Source.c" startsOnLine="38" />
    <testcase name="SomeIgnoredTest" source="C:\Path\To\Source.c" startsOnLine="52" ignored="true" />
  </group>
</testrunner>
````

If anyone has suggestions for additional information that would be nice to expose, I'd love to hear them.

PS: This is currently being developed for internal use at one of our clients. We are looking into open-sourcing the Visual Studio integration. :wink:

arstrube:
I wonder whether this is really necessary? Have a look at this test runner for Eclipse. Perhaps it can give you some ideas:

git@ssh.github.com:tcmak/CppUTestEclipseJunoTestRunner.git

arstrube:
As you will see from the sources, the usual way is to parse the console output. One has to look at the cdt testrunner in order to see the rest of the 'magic'. All I know is that the current Cli contains everything needed to make it work, in particular, the possibility to run any combination of tests.

I don't have any deep knowledge of this, but I believe it initially has no knowledge of which test groups and test cases are there. It derives this from the console output, as well as the source path to failed tests, so you can jump to the source of failed tests. I don't think it lets you jump to the source of successful tests (haven't used it for some time).

So it runs the entire test suite the first time, builds a graphical representation of test groups and tests, and lets you run tests selectively thereafter.

itavero:
[...]

Issue #903, opened 2016-03-17
-ojunit output xml file format doesnot have any info regarding test source file name


bkjoshi76:
usecase : integrating cpputest output xml ( junit) format to the C++ sonar community plugin

c++ sonar community needs to have test src file name for displaying test execution status on the sonar dashboard.

Any option available to get test src  file info in xml file (-ojunit) format?

arstrube:
There is -k `packagename` -- might that help?

itavero:
I do not believe that the JUnit / xUnit XML format allows you to include such data.
See:
* https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
* http://reflex.gforge.inria.fr/xunit.html#xunitReport

If this data is needed, I think it will need to be in a new output format (so not in `-ojunit`).

itavero:
I also need information about the location of test cases within the sources, however I need to retrieve it *without* running the tests.
See issue #967.

itavero:
This XSD does seem to include the attributes you are looking for:
https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd

itavero:
@bkjoshi76 While working on a solution for #967, I've added a `file` and `line` attribute to each testcase that indicates the position of each test (first line).
I'd like to know if this information will be enough for you to get a better integration within Sonar, or do you need more information to be included in the XML?

My code will currently provide you with an XML file similar to this example:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<testsuite errors="0" failures="0" hostname="localhost" name="groupName" tests="2" time="0.046" timestamp="2016-05-20T08:07:42">
<properties>
</properties>
<testcase classname="groupName" name="firstTestName" assertions="4" time="0.031" file="/path/to/testsource.cpp" line="23">
</testcase>
<testcase classname="groupName" name="secondTestName" assertions="6" time="0.015" file="/path/to/testsource.cpp" line="34">
</testcase>
<system-out></system-out>
<system-err></system-err>
</testsuite>
```

arstrube:
@itavero I like that the number of assertions is in there. I think it is an important metric - I have seen "tests" with NO assertions :-/.....

bkjoshi76:
Hi,

This is nice solution what i am looking.

I think its good for sonar integration

Regards

Bhaskar joshi

itavero:
[...]

PR #1010, opened 2016-06-29
Extended JUnit output

### LINKED ISSUE #120: JUnit Style xml output file including the number of CHECKS
Since the amount of checks is an important metrics, propose to give out this metric in the XML output file.

itavero:
Added the following to the JUnit XML output:
* Location of test (path to file and line number)
* Number of checks (assertions) in a test case

Initial added this while working on #967, but my daytime job is taking up a lot of my time and these changes have been requested in other issues too, hence the PR.

This should fix #120 and #903.


itavero:
I'd love to hear your opinions, @basvodde & @arstrube...

tests/JUnitOutputTest.cpp
```diff
@@ -499,7 +520,7 @@ TEST(JUnitOutputTest, testFailureWithDifferentFileAndLine)
 {
     testCaseRunner->start()
             .withGroup("testGroupWithFailingTest")
-                .withTest("FailingTestName").thatFails("Test failed", "importantFile", 999)
+            .withTest("FailingTestName", "sourcefile.c", 123).thatFails("Test failed", "importantFile", 999)
```
basvodde: This change isn't too good. It is breaking the builder. The way you would probably want to get this is:

.withTest("FailingTestName").in("sourcefile.c").lineNumber(123)

That way you do not need to pollute all the tests with the "sourcefile.c" information, which is irrelevant for most of the tests.

Making builders work like that is fun :)

itavero: I'll look into it. Never really wrote a builder like this. :wink:

itavero:
@basvodde Besides the changes I made to the unit tests (`tests/JUnitOutputTest.cpp`), do the other changes look good to you?

itavero:
For some reason the Windows 64 MinGW build failed, but I do not really see why...

Andne:
Every so often the install command fails, never really been sure why, seems like there's a network glitch occasionally that prevents the download or something.

itavero:
@Andne, is there an easy way to restart the build without pushing a new commit?
And has an issue been created to make the build more reliable?

arstrube:
@itavero close and reopen the PR.

itavero:
You have got to be kidding me.. now the 32bit Windows MinGW build fails... :worried:

basvodde:
Nice.

PR #1018, opened 2016-07-07
Expected calls matching several actual calls

jgonzalezdr:
Basically the feature that I've added is the capability for a single expected call to match multiple actual calls, between a minimum and a maximum number of actual calls.

From the library user perspective, some new methods have been added to MockSupport:
- expectAtLeastOneCall
- expectAtLeastNCalls
- expectAtMostOneCall
- expectAtMostNCalls
- expectAnyCalls
- expectRangeOfCalls

The first improvement to the actual implementation is that expectNCalls doesn't internally generate N expected call objects, therefore memory usage is optimized and this can be fairly important when using the mock framework to perform integration or simulation tests in target platforms where memory resources are scarce.

The second improvement is the capability to declare "optional" expected calls, i.e. expected calls that can match actual calls according to some parameters and then return some specific value, but that we don't know and/or don't care when writing the test how many times they may be called. This is essential when using the mock framework to perform simulation tests, and that was actually the key motivation to implement that new feature, as I needed at work to simulate the behavior of a battery state of charge algorithm under different driving cycle scenarios, and the mocking framework was successfully used to simulate inputs and to check outputs.

The changes are rather extensive, but they have been performed as a series of incremental commits, and each commit is relatively simple and self-contained, solving small problems each time. Therefore, to review this PR, I recommend to inspect each commit in order.

In the following messages I'll comment a bit more about each commit, to make review a bit simpler.


jgonzalezdr:
For some reason, if you check the commits in this PR, some appear out of order, so be aware of this when reviewing them to avoid confusion.

You can always review them instead in my own branch, where they appear to be in the proper order:
https://github.com/jgonzalezdr/cpputest/commits/expected-calls-multimatch

jgonzalezdr:
Commit #1: [(dae0892) Preparation of mock library...](https://github.com/cpputest/cpputest/pull/1018/commits/dae0892ce56f30a723d51b1cfa3a58dfe1f771b0)

These modifications don't change or add any functionality, they just rename some methods and attributes, and add a couple of additional methods, in preparation for the actual implementation of multi-matching expected calls.

The key concepts changed are about **matching** actual calls, and **fulfilling** expected calls. 

Until now, an expected call could _match_ a single actual call, and when this _match_ was done, then the expected call was _fulfilled_, so _matching_ and _fulfilling_ was rather the same thing. 

From now on, these two concepts will be separated: During the processing an actual call, expected calls will be able to _match_ it under certain conditions (i.e. parameters and object match, the expected call hasn't reached the maximum number of actual calls that it can match, etc.). However, an expected call will be considered _fulfilled_ when it has _matched_ between its minimum and its maximum of actual calls.

This means that an expected call will be considered to be _fulfilled_ as soon as it reaches its minimum number of _matched_ actual calls, but that it will be able to _match_ actual calls until it reaches its maximum.

Additionally, notice that the state stored in expected calls relative to _"matching"_ will be only relevant while an actual call is being processed, but the state stored relative to _"fulfilling"_ will be relevant during the whole life cycle of the expected call.

jgonzalezdr:
Commit 2: [(ed56f39) Basic implementation of "multi-matching" expectedCalls](https://github.com/cpputest/cpputest/pull/1018/commits/ed56f3916328c7ce9cc6b3a96d25051fae5f4594)

These modifications implement the basic features, but they break for now strict ordered checks when using methods apart from expectOneCall. 

Additionally, the existing feature by which fulfilled expected calls are reported in call order (on failures) is also broken, because it has no sense now: As expected calls can match more than an actual call, and those could happen in any order, they are just reported in declaration order, and the actual calls order is not reported at all for now.

jgonzalezdr:
Commit 3: [(8ee564a) Added C language new calls for multi-matching expected calls](https://github.com/cpputest/cpputest/pull/1018/commits/8ee564aab45a024a2e24dba991947b2f29ad2c2d)

These modifications are rather trivial.

jgonzalezdr:
Commit 5: [(e94cd93) Added test that checks that non-fulfilled calls have higher matching...](https://github.com/cpputest/cpputest/pull/1018/commits/e94cd9377cd01ba5b3f0a9bf27b18eb56c31fd7c)
Commit 6: [(f42377e) When creating the list of potentially matching expected calls…](https://github.com/cpputest/cpputest/pull/1018/commits/f42377ef2147fc0709c8ae83af24d77557e86cab)

These modifications ensure that expected calls that aren't yet fulfilled (i.e. not having reached the minimum number of matching calls) have higher precedence over expected calls that are already fulfilled (because they've reached their minimum), to avoid the latter being greedy.

jgonzalezdr:
Commit 7: [(162d117) Renamed MockExpectedCallsDidntHappenFailure...](https://github.com/cpputest/cpputest/pull/1018/commits/162d117013d12855af7914b9143fd3fccf12bd68)

The rename was made to improve consistent nomenclature, as the new name better suits the actual failures that are handled by that class.

jgonzalezdr:
Commit 8: [(94c5c69) Re-enabled strict ordering checks when...](https://github.com/cpputest/cpputest/pull/1018/commits/94c5c69a72188f667df0fb00fde100e4af1a28ed)

Strict checking of mock call order is enabled again, but it works only for expected calls that have no optional calls (i.e. when the minimum and maximum number of expected calls are equal). If strict ordering is enabled, and an expected call with optional calls is declared, an failure is thrown.

The way to achieve this is by keeping track of the minimum and maximum expected call order, and marking the expected call as out of order if it's called outside of the expected range.

This solution is simple, but it makes the usage of optional calls impossible. I've considered other solutions, but they are way more complex and would require to fail as soon as an out of order call is detected (which is not the present behavior). Therefore, taking in account that probably nobody will find usage for strict ordered optional calls, I opted for the simplest solution.

jgonzalezdr:
Commit 9: [(84593a9) Fixed declaration of MockExpectedCallsListForTest::addFunction…](https://github.com/cpputest/cpputest/pull/1018/commits/84593a91c276cb4000f9d1904159951909867e24)
Commit 10: [(49f3b7d) Fixed declaration of call order related methods and attributes to use…](https://github.com/cpputest/cpputest/pull/1018/commits/49f3b7dc9692fd9537bbda1096a8671c6061e7b5)

Those modifications change the type of all parameters and attributes related to call order from _int_ to _unsigned int_, for consistency and internal coherency.

jgonzalezdr:
Commit 10: [(60828dc) Updated some unit tests and added a new one to improve code coverage.](https://github.com/cpputest/cpputest/pull/1018/commits/60828dcfb08376921b9f7761bc170b67b0239f47)

Trivial modifications to improve code coverage.

jgonzalezdr:
Commit 12: [(a12e83b) Modified MockSupport::expectNoCall() method implementation to use...](https://github.com/cpputest/cpputest/pull/1018/commits/a12e83b19250182091e1ea15ac9d70ae611da374)

The present implementation of expectNoCall() can be replaced to use expected calls that expect exactly 0 calls.

Additionally, this commit also fixes the reporting of extra unexpected calls, that was erroneously calculating the reported Nth additional call.

jgonzalezdr:
Commit 13: [(6b5db95) Removed superfluous constructor and method in MockExpectedCall…](https://github.com/cpputest/cpputest/pull/1018/commits/6b5db9522a22d63e899c9cc9cf375d76df25d7ba)

Trivial modification to remove unnecessary code.

jgonzalezdr:
Commit 14: [(412ca5e) Fixed the mock scope not being properly reported when failing because…](https://github.com/cpputest/cpputest/pull/1018/commits/412ca5e5024fd7787a729820aa531da2419f2dc1)

Fixes a bug when reporting a failure when using optional calls with strict ordering checks, and adds a unit test to check that.

jgonzalezdr:
Commit 15: [(0f4623a) Added a class to store a queue of actual calls...](https://github.com/cpputest/cpputest/pull/1018/commits/0f4623a70065e11e8c9e1615ec73655823947b97)

Adds the class that will store the list of actual calls, in the form of a queue to allow limiting its size.

jgonzalezdr:
Commit 16: [(0f4623a) Modified MockCheckedActualCall to store a reference to its associated…](https://github.com/cpputest/cpputest/pull/1018/commits/0f4623a70065e11e8c9e1615ec73655823947b97)

This modification is needed because when the log of actual calls is implemented, it will have to be passed around in order to have it available when a failure is to be thrown, and just because the MockSupport object will store it (just as it's now storing the list of expected calls) it's a bit more efficient to pass the mock object to the actual call rather than both lists.

jgonzalezdr:
Commit 17: [(bc33181) Added functionality to MockCheckedActualCall and MockActualCallsQueue…](https://github.com/cpputest/cpputest/pull/1018/commits/bc331818188cbb2a5f49ec21ecb0c5156132fc60)

Here the functionality to generate strings with the description of actual calls (including call order) has been added, which will be used later in the failure reports.

jgonzalezdr:
Commit 18: [(044a693) Added actual calls reporting to MockFailure, and preliminary…](https://github.com/cpputest/cpputest/pull/1018/commits/044a69359b4ccb172f140d5b1286b6fa97eda050)

With these modifications the failure classes can now report the actual calls in call order.

Take in account that at this point, the actual calls queue is always passed empty in production code.

jgonzalezdr:
Commit 19: [(b3a20a4) Fixed compilation for Dos platform.](https://github.com/cpputest/cpputest/pull/1018/commits/b3a20a45aeda63320790ec849508d51219ecdfbb)

Trivial modification to fix compilation in travis ci.

jgonzalezdr:
Commit 20: [(eb3aae1) Implemented a log of actual calls in MockSupport…](https://github.com/cpputest/cpputest/pull/1018/commits/eb3aae13db1d5e473d4bf0babc6cbab8e2c623f8)

This finally implements the functionality to keep a log of actual calls in each mock object, effectively re-introducing the capability to print the actual calls in call order when a failure happens.

jgonzalezdr:
Commit 21: [(fb5653f) Added functionality to MockSupport to limit the maximum size…](https://github.com/cpputest/cpputest/pull/1018/commits/fb5653fbfe4c7c5c290fa1f40d1c2cddd238d670)
Commit 22: [(3999548) Added C language new call for setMaxCallLogSize.](https://github.com/cpputest/cpputest/pull/1018/commits/3999548cfc66781083983cce68dbb8c2a21b8fc7)

This commits implement the functionality to limit the size of the actual calls log. 

This feature is essential when performing integration or simulation tests that perform lots of actual calls in embedded target platforms with scarce memory, but it's also very useful in other cases to avoid saturating the failure report. In fact, I'm wondering if it would be convenient to have a default limit value like 20 or so, because in the vast majority of cases that should provide enough information to the user, and in the rest of cases he can just change that limit as needed for individual tests.

jgonzalezdr:
Commit 23: [(faf07f9) Added unit test for MockActualCallsList::hasFinalizedMatchingExpectat…](https://github.com/cpputest/cpputest/pull/1018/commits/faf07f9a2305c1e9bdafcbc29fd7c909d610251f)
Commit 24: [(e912650) Improved unit test for MockSupport::setMaxCallLogSize().](https://github.com/cpputest/cpputest/pull/1018/commits/e91265075f9ff10f8d6e6e9ad2d2d53c86c13723)

Just added some additional trivial unit tests to improve coverage.

basvodde:
Wow, what happened to small pull requests :) This thing is huge.

First question. In which scenario would you *want* to use expectAtLeastOneCall (or any of them) ?

jgonzalezdr:
I know the PR is huge, the problem is that it's an all or nothing thing, because the different "pieces" are useless alone and moreover, they would break some capabilities of the current implementation.

Regarding the family of "expect call methods", I'll admit that expectOneCall, expectNCalls, expectAnyCalls and expectNoCalls are the most useful ones, especially with simple white-box unit tests, but the other ones are useful when performing TDD or for more complex integration tests written according to a behavior specification (black-box testing) and not according to the actual implementation (white-box testing).

I'll put an example based on a real-world case:

We have a _Configuration_ class that manages the configuration for a device stored in EEPROM, and it has a method "_bool IsFeatureXActivated()_" that returns a boolean indicating whether a certain feature is activated. We have a class _DigitalOutputs_ with a method "void ActivateOutputY()". We have a class _SoundManager_ with a method "_void PlaySound(enum...)_".

Then we have a function "_void OnEventZ()_" to be tested which behavior is specified as, among other things, "if feature X is activated, then the output X shall be activated and a long beep sound shall be played".

We could write a test like that:
```
TEST(TestGroup, OnEventZ_FeatureXActivated)
{
    mock().expectOneCall("Configuration::IsFeatureXActivated").andReturnValue(true);
    mock().expectOneCall("DigitalOutputs::ActivateOutputY");
    mock().expectOneCall("SoundManager::PlaySound").withParameter("type", LONG_BEEP);
    ...

    OnEventZ();

    mock().checkExpectations();
}
```
This test seems fine, but it will fail if the implementation does something like:
```
if (Configuration::IsFeatureXActivated()) DigitalOutputs::ActivateOutputY();
...
if (Configuration::IsFeatureXActivated()) SoundManager::PlaySound(LONG_BEEP);
```
It fails because conceptually it's OK to assume that the "output" calls shall be called just once, but unfortunately the "input" calls can be invoked more than once without breaking the specs. Therefore, a more appropriate test would be:
```
TEST(TestGroup, OnEventZ_FeatureXActivated)
{
    mock().expectAtLeastOneCall("Configuration::IsFeatureXActivated").andReturnValue(true);
    mock().expectOneCall("DigitalOutputs::ActivateOutputY");
    mock().expectOneCall("SoundManager::PlaySound").withParameter("type", LONG_BEEP);
    ...

    OnEventZ();

    mock().checkExpectations();
}
```
This way, the test won't break each time that the implementation is changed without actually breaking the specs.

Keep in mind that this is a simplified example, but my team and I had to struggle with the maintenance of lots of tests for complex functionalities developed in "agile ways" (i.e. the customer changing each five seconds some details of the specs), that were breaking all the time during development because of being defined "too strictly". That's also one of the key factors that induced me to implement these new mocking methods.

uecasm:
A similar case where this sort of desire crops up is where you're doing "loose" AAA tests -- you have lots of little test cases that each want a little bit of setup and one main thing that you're asserting on -- eg. you care that calling `Frobnicate(42)` results in calling `TurnBlargHandle(8)` at some point but you don't care what else it does along the way.  You might provide some dummy return values for things that `Frobnicate` is allowed to call but is not required to, and don't care that it calls twenty other methods to get there, as long as it does make that one call you care about.

It all depends on what you're wanting to do with the mocks, whether they're exact-implementation-replay or something more loose.  Sometimes people call the latter kind "stubs" or "fakes" but they're still mocks at heart.  Generally the closer you tie the test to the actual implementation, the more brittle it is and the harder it is to refactor things without breaking a large number of tests, often in trivial ways.

jgonzalezdr:
Totally agree with uecasm, 

I usually call that "aspect testing": Sometimes integration tests have to be performed on a fairly complex module made of several pieces (each tested with their own unit tests), and we need to have the confidence (and evidence!) that the pieces work well together, but then the resulting module is too complex to test it thoroughly as a whole. Then, in a similar way than when system or hardware tests are performed, the module is tested based on different individual "aspects" of its behavior.

A real-world example is testing the behavior of a task defined in a Time-Triggered RTOS. That task is defined as a class that has a method that shall be called exactly once every 1 milliseconds. To test the behavior of the task over time, a battery of tests is made which call the task method in a loop, stimulating some inputs and checking some outputs, while the rest of the inputs are just mocked so they don't interfere with the "aspect" that we're testing in each particular test by returning specific values (which can't be achieved using _ignoreOtherCalls()_).

Usually, the most useful method when doing this kind of tests is _expectAnyCalls()_, but the other ones are also useful: for example, _expectAtMostNCalls()_ can be used to simulate an input that changes its value over time when we don't know exactly how many times the call that reads the input will be called. In my particular example, the task amongst other things has to check that a digital input signal changes its state each 100 milliseconds, and that if it doesn't change it shall enter a specific state of its internal FSM at most 50 milliseconds after the missing edge. 

Let's take a look at a test that checks that for two proper cycles the task doesn't trigger any state change, then the related signal becomes stuck and the task finally triggers the state change:
```
TEST(TaskA, missingEdgeDetection)
{
    mock().expectNCalls(100, "GPIO::GetPortA").andReturnValue(0x48):
    mock().expectNCalls(100, "GPIO::GetPortA").andReturnValue(0x40):
    mock().expectNCalls(100, "GPIO::GetPortA").andReturnValue(0x48):
    mock().expectNCalls(100, "GPIO::GetPortA").andReturnValue(0x40):
    mock().expectAtMostNCalls(150, "GPIO::GetPortA").andReturnValue(0x48):
    ...
    mock().expectAnyCalls("GPIO::GetPortB").andReturnValue(0xF0): // Unrelated
    mock().expectAnyCalls("GPIO::GetPortC").andReturnValue(0xA5): // Unrelated
    ...

    while (taskA->GetState() != TASKA_STATE_FAILURE) {
        taksA->Activate();
    }

    mock().checkExpectations();
}
```
A deep white-box analysis (or trial and error) would permit replacing the optional calls with an exact number of calls, but this wouldn't make the tests more robust and reliable, because a non-important change (in this context) like modifying the filtering algorithms for inputs would break this test, when that should only break the tests that check specifically aspects related to filtering.

This kind of integration tests are common and necessary when developing critical systems (like in automotive or railway). It's true that for systems with critical functional safety requirements higher than SIL-2 or ASIL-B a certified unit testing tool is almost mandatory, but for less critical systems CppUTest is just perfect, and these new methods will make that kind of tests easier and more robust.

basvodde:
I'm never going to be able to get through this pull request as one. Would you mind trying to split it up in smaller steps.

My suggestions for first steps:

- Pull out commits that clean the code without adding functionality (small refacatorings)
- Start with doing the ExpectNCalls so that it would not create a Call object for each (which has caused some performance problems and we have an issue open for that)

basvodde:
Related to the AtLeast kind-of calls. I'm still a bit worried about these, adding these might cause people to use them... what I mean is that they start writing less strict assertions and weaker unit tests. I think I'll probably eventually end up integrating them, but want a short discussion here first. Please be patient (although we can start with smaller pull requests related to the expectNCalls, as mentioned above)

jgonzalezdr:
Don't worry @basvodde, I already knew such a big PR would require a lot of review and possibly submitting it as smaller pieces. Doing like you say will provide smaller PRs that don't brake almost anything.

jgonzalezdr:
@basvodde: I've just submitted the first small PR (#1020), when this is reviewed, I'll submit another one that implements the N calls with a single expected call object.

jgonzalezdr:
@basvodde: Regarding your concerns related to less strict assertions and weaker unit tests, I think that the new methods are as good or evil as ignoreOtherCalls() or ignoreOtherParameters(). In fact, other mocking frameworks like Google Mock, Mockito, EasyMock or jUnit all have this kind of features, because they're very useful.

From my experience, writing proper unit tests does not depend on the unit testing framework or the lack of some "dangerous" features, but only on proper judgement, and because that last thing sometimes seems too scarce, performing code reviews on the unit tests, not only on the production code.

I've seen programmers under deadline pressure write totally deceiving tests that just to increase coverage, but that don't actually check that the behavior is right:
```
TEST(ClassX, cheatingTest)
{
    mock().expectOneCall("functionA").ignoreOtherParameters().andReturnValue(true):
    mock().expectNCalls(100, "functionB");
    mock().ignoreOtherCalls();

    functionUnderTest(1); // This function returns a value, but it's not checked
    functionUnderTest(10); // This function returns a value, but it's not checked
    functionUnderTest(50); // This function returns a value, but it's not checked

    mock().clear();
}
```

cuitoldfish:
@jgonzalezdr , I have elaborated on this mock scenario for a long time because my colleague @maxilai submitted almost same PR #973 .

I realize that you both have one common pain point,  getting extra maintenance effort when dealing with 
**expectOneCall().andReturnValue()**

but not with 
**expectOnceCall().withParameter()**

This bring out to one idea I'm thinking for a long time, are we combine the two mock functionality "what mock class should action" and "what mock class should check" into the **expectOneCall().andReturnValue()** expression?
GMock has a good statement 

>ON_CALL defines what happens when a mock method is called, but doesn't imply any expectation on the method being called. EXPECT_CALL not only defines the behavior, but also sets an expectation that the method will be called with the given arguments, for the given number of times

If we have one separate expression like **mock().andReturnValue()**, then if you just only want to mock some behavior, you can suppress the mock check. Will this handle most pain case in @jgonzalezdr situation?

From the utility point of view,  comparison between CppUTest with GTest, it's hard to say who is better, GTest has many utility and  keep as flexible as possible, and CppUTest keep as simple enough as possible. The more flexible is, the harder to learn and use.

Anyway, if we accept this PR strategy, i have following concern
1. we have a new "ExpectAtMostNCall", then how to easily distinguish "ExpectAtMostNCall" and "ExpectNCall".  
2. should we keep what CppUTest's user's current understanding about the usage and the test failure output even if we have new features? 
3. how do we think of the compatibility with older version which have already generated many many test case?

jgonzalezdr:
Replying to @cuitoldfish comments:

1. If I'm understanding well, you may be proposing to rename _expectedNCalls_ to _expectExactlyNCalls_ or something like that, to avoid confusions. I didn't do that on purpose, to avoid breaking the existing API and backwards compatibility. Furthermore, I think that each function name has enough semantical information about its purpose, and in any case users with doubts can always RTFM :-D. I don't think that programmers that know the difference between "==" and "<=" will have any problems with the expectXXXCalls methods.

2. The output in case of failure is very similar to the previous one, and pretty self-explanatory, so it shouldn't be too hard to analyze and understand by the average programmer:

New output:
```
	Mock Failure: Expected call WAS NOT fulfilled.
	EXPECTED calls that WERE NOT fulfilled:
		foo::bar -> all parameters ignored (expected 3 calls, called 1 time)
	EXPECTED calls that WERE fulfilled:
		foo -> no parameters (expected 2 calls, called 2 times)
		foo -> int bar: <1 (0x1)> (expected 1 call, called 1 time)
	ACTUAL calls that were expected (in call order):
		(1) foo -> no parameters
		(2) foo -> int bar: <1 (0x1)>
		(3) foo -> no parameters
		(4) foo::bar -> all parameters ignored
```
Old output
```
	Mock Failure: Expected call did not happen.
		foo::bar -> all parameters ignored
		foo::bar -> all parameters ignored
		foo -> no parameters
		foo -> int bar: <1 (0x1)>
		foo -> no parameters
		foo::bar -> all parameters ignored
```

3. These modifications are backward compatible, therefore existing tests should work as expected.

cuitoldfish:
ok, agree. And how do you think my previous part   

If we have one separate expression like mock().andReturnValue(), then if you just only want to mock some behavior, you can suppress the mock check. Will this handle most pain case in your situation?

cuitoldfish:
I mean, if we add another function to let user specify what mock class will action but not to check, such as **mock().andReturnValue()**, then your example test case will look like
```.cpp
TEST(TestGroup, OnEventZ_FeatureXActivated)
{
    mock("Configuration::IsFeatureXActivated").andReturnValue(true);
    mock().expectOneCall("DigitalOutputs::ActivateOutputY");
    mock().expectOneCall("SoundManager::PlaySound").withParameter("type", LONG_BEEP);
    ...

    OnEventZ();

    mock().checkExpectations();
}
```
then no matter how many times **isFeatureActivated()** are invoked, test case won't be hurted.

cuitoldfish:
or we could directly use the mock().setData,  I forget this one :(
```.cpp
TEST(TestGroup, OnEventZ_FeatureXActivated)
{
    mock().setData("Configuration::IsFeatureXActivated::returnValue", true);
    mock().expectOneCall("DigitalOutputs::ActivateOutputY");
    mock().expectOneCall("SoundManager::PlaySound").withParameter("type", LONG_BEEP);
    ...

    OnEventZ();

    mock().checkExpectations();
}
```

```.cpp
bool MockConfiguration::IsFeatureXActivated()
{
    return mock().getData("Configuration::IsFeatureXActivated::returnValue").getIntValue();
}
```
If I understand your example correctly, could this solution eliminate your test maintenance effort?

yes, your above example is a simple one, but I think, from unit test point of view, although the "input call" from outside should not be checked so strictly,  the "output" call to the outside should be checked strictly in most case, isn't it?

jgonzalezdr:
I find too many drawbacks to your proposal, and also to the one in #PR973:

1. The _ignoreAdditionalCalls()_ method proposed in #PR973 is too vague. 

 I can't see any proper test case where it's useful, because it's a "blank check". In the example that I proposed, it would also allow calling several times the other functions, which is *not* the expected behavior. 

 Moreover, it wouldn't allow precise control of returned values. 

 In fact, having the new functions, I would totally forbid the usage of _ignoreOtherCalls()_ to my teams, for the same reasons: One thing is taking in account that an actual call has no side-effects and is safe and according to the specs to expect more than just once, making the tests less prone to breaking unnecessarily but remaining as strict and robust as it should be; but a way different thing is allowing blindly anything, without control of returned values, etc.

2. Implementing this _ignoreAdditionalCalls()_ method in the expected call instead of the mock support, would be a more elegant and controlled solution.

 The resulting calls could be something like `expectNCalls(5, "whatever").withParam("param", 5).ignoreAdditionalCalls().withReturnValue(false)`. 

 This would be equivalent to `expectAtLeastNCalls(5, "whatever").withParam("param", 5).withReturnValue(false)`.

 But anyway, I still prefer my solution, as I think that it's more concise and meaningful.

3. Regarding the proposal to to something like GMock ON_CALL, I really think that GMock's solution is confusing and not the way to go. 

 Just search "GMock ON_CALL" on google, and there's full of questions of users that don't understand the difference between EXPECT_CALL and ON_CALL.

 From my point of view, all the actual calls happening in a test should be expected, regardless of whether I expect any calls, at least one, or at most 50. I think they are expected because writing down these expectations comes from a process of analysis and judgement to determine how the test best checks properly the behavior of a functionality. In other words, unexpected things should not be happening in tests.

 I understand the intention behind ON_CALL, or generalizing this concept, defining "default" behaviors for functions: It's just indicating that these calls are not really relevant for the actual "aspect" that is being tested in the actual test. But I don't think that the testing framework should differentiate both; if you want to do that you can simply place a comment in the test indicating that fact.

4. I consider an _andReturnValue()_ method in the mock support to be awful semantically. Maybe renamed to _defaultReturnValue()_... but anyway, it's nonsense to me: What is the point that any unexpected call should return the same value? What happens if I set it to return an integer, and the call shall return a pointer?

5. Your first proposal to narrow down the affected functions using `mock("Configuration::IsFeatureXActivated").andReturnValue(true)` is, in my opinion, not "user friendly", abuses a bit on the concept of "mock scopes", and it would require to define the mocked functions once and again for different scenarios.

 In my projects, the preferred solution is to mock classes just once, and these mocks are shared between all tests. Therefore, all the mocked functions are implemented just once, in the likes of `mock().actualCall("functionName")...`. If you've got tenths or hundreds of classes, and thousands of tests, it's essential to share the mocks to avoid redundant work.

6. The second proposal using getData/setData has the same drawbacks than your first proposal, as I just explained before, but additionally it would be much more complex to implement.

cuitoldfish:
thanks for the clarification, I think right now I have gotten the key point in your example and motivation for this new feature. I think it is the more balanced design for some use scenario which need more advanced mock control.

PR #1020, opened 2016-07-11
Expected calls matching several actual calls - (1) Preparation

jgonzalezdr:
As discussed in PR #1018, I submit here some small refactoring changes in preparation for the next modifications.

These modifications don't change or add any functionality, they just rename some methods and attributes, and add a couple of additional methods, in preparation for the actual implementation of multi-matching expected calls.

The key concepts changed are about **matching** actual calls, and **fulfilling** expected calls. 

Until now, an expected call could _match_ a single actual call, and when this _match_ was done, then the expected call was _fulfilled_, so _matching_ and _fulfilling_ was rather the same thing. 

From now on, these two concepts will be separated: During the processing an actual call, expected calls will be able to _match_ it under certain conditions (i.e. parameters and object match, the expected call hasn't reached the maximum number of actual calls that it can match, etc.). However, an expected call will be considered _fulfilled_ when it has _matched_ its minimum and its maximum number of actual calls.

This means that an expected call will be considered to be _fulfilled_ as soon as it reaches its minimum number of _matched_ actual calls, but that it will be able to _match_ actual calls until it reaches its maximum.

Additionally, expected calls have another associated concept: **finalized**. Calls that don't ignore parameters are automatically considered to be _finalized_ when they have matched its object and all its parameters (if they have any). However, calls that can ignore parameters are not _finalized_ until the actual call is finalized (i.e. checkExpectations() has been called on the actual call). I think that this clarifies a bit the "...WithoutIgnoredParameters" and "...WithIgnoredParameters" methods in expected calls list, because this nomenclature was a bit confusing (does that mean that we look for calls that can ignore parameters, or that have actually ignored already some?).

Notice that the state stored in expected calls relative to _"matching"_ will be only relevant while an actual call is being processed, but the state stored relative to _"fulfilling"_ will be relevant during the whole life cycle of the expected call.

### DIFF
--- a/tests/CppUTestExt/ExpectedFunctionsListTest.cpp
+++ b/tests/CppUTestExt/ExpectedFunctionsListTest.cpp
@@ -63,7 +63,7 @@ TEST_GROUP(MockExpectedCallsList)
 TEST(MockExpectedCallsList, emptyList)
 {
     CHECK(! list->hasUnfulfilledExpectations());
-    CHECK(list->size() == list->amountOfUnfulfilledExpectations());
+    CHECK(! list->hasFinalizedMatchingExpectations());
     LONGS_EQUAL(0, list->size());
 }
 
@@ -74,7 +74,7 @@ TEST(MockExpectedCallsList, addingCalls)
     LONGS_EQUAL(2, list->size());
 }
 
-TEST(MockExpectedCallsList, listWithFulfilledExpectationHasNoUnfillfilledOnes)
+TEST(MockExpectedCallsList, listWithFulfilledExpectationHasNoUnfulfilledOnes)
 {
     call1->callWasMade(1);
     call2->callWasMade(2);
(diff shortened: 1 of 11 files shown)


tests/CppUTestExt/ExpectedFunctionsListTest.cpp
```diff
@@ -266,17 +266,28 @@ TEST(MockExpectedCallsList, callToStringForFulfilledFunctions)
...
 TEST(MockExpectedCallsList, toStringOnEmptyList)
 {
     STRCMP_EQUAL("<none>", list->unfulfilledCallsToString().asCharString());
 }
+
+TEST(MockExpectedCallsList, hasFinalizedMatchingExpectations)
+{
+    call1->ignoreOtherParameters();
+    list->addExpectedCall(call1);
+    CHECK(! list->hasFinalizedMatchingExpectations());
```
basvodde: These are 2 tests in one?

basvodde:
It seems good. Only one comment to split a test in two. THen I'll merge this as a good start!

jgonzalezdr:
I've split the test and added some additional ones.

PR #1023, opened 2016-07-14
Expected calls matching several actual calls - (3) Implementation of "multi-matching" expectedCalls.

jgonzalezdr:
This PR adds at least the initial implementation of multi-matching expected calls, but only for a fixed number of expected calls.

The basic idea is simple: An actual call is not considered to be made until _checkExpectations()_ is called on it (i.e. another actual call has been created or _checkExpectations()_ is called on the mock support), therefore _callWasMade()_ is not called on the matching expected call until there is only a single one selected (i.e. the one which fulfills the actual call). When this occurs, the corresponding count of matched actual calls on the expected call is increased.

The drawback of this is that actual calls order is not registered properly, so for the meantime
failure messages just show the expectations that were fulfilled in
declaration order, as it has no sense anymore to order them by call
order. This will be corrected in subsequent PRs.


jgonzalezdr:
The decrease in coverage is just due to mathematics: Having deleted the composite calls, there are fewer lines of code, so the already uncovered ones gain a bit of weight.

include/CppUTestExt/MockCheckedExpectedCall.h
```diff
@@ -36,10 +36,12 @@ class MockCheckedExpectedCall : public MockExpectedCall
 
 public:
     MockCheckedExpectedCall();
+    MockCheckedExpectedCall(unsigned int numCalls);
     virtual ~MockCheckedExpectedCall();
 
     virtual MockExpectedCall& withName(const SimpleString& name) _override;
-    virtual MockExpectedCall& withCallOrder(unsigned int callOrder) _override;
+    virtual MockExpectedCall& withCallOrder(unsigned int callOrder) _override { return withCallOrder(callOrder, callOrder); }
```
basvodde: Can you avoid inline implementations, as they have caused problems in the past (on some platform)

basvodde:
This PR is rather huge. But, based on the comments, won't this cause a problem by not failing the test in the actualCall as soon as it is possible to fail it?

include/CppUTestExt/MockCheckedExpectedCall.h
jgonzalezdr: It's just temporary, in subsequent PRs I clean this up, removing that superfluous call.

But I can fix that anyway for the meantime if you want.

jgonzalezdr:
The effect of delaying the callWasMade has no incidence on failures, because out of order calls actually don't make the tests fail immediately when a call is made out of order, that behavior hasn't changed at all: an out of order failure is triggered only when calling checkExpectations() on the mock support, and only if there isn't any other failure (like expected calls not fulfilled).

In fact, with these changes, the internal state of expected calls is more coherent: In the current implementation, when strict ordering is enabled, all the unfulfilled expected calls that match the actual call function name are marked as "out of order" if they don't match the order (i.e. normally all the calls except one), even if they are later discarded and not actually "called" at all. If the test doesn't fail it's not relevant, because finally all the calls will match and be marked as "in order"; and if the test doesn't fail it's also irrelevant, because this information is not displayed to the user.

With the new implementation, the expected call is considered "out of order" only if it's actually called out of order, in other cases it's considered to be "in order".

jgonzalezdr:
You may find it easier to review this PR if you review the individual commits:

- The first one implements the usage of a single expected call object for multiple calls, but sets aside for the meanwhile strict ordered tests proper implementation.
- The second one deletes the now unnecessary composite call class.
- The third one fixes the strict ordered tests.
- The fourth one adds some additional test to compensate some coverage loss due to indirect testing made by the composite call tests.

basvodde:
Could you next time make the PRs still smaller :)

jgonzalezdr:
Any way, from the users perspective, the behaviour of the framework is the same than before, except for the information displayed when a test fails.

jgonzalezdr:
I would have loved to push a smaller PR, but sincerely I didn't manage to reduce this one without passing by an intermediate state where the framework is actually broken :-(.

include/CppUTestExt/MockCheckedExpectedCall.h
cuitoldfish: @basvodde would you like to give more information about the unexpected behavior for inline function :)? I have also met unexpected behavior on one TI's platform but can't dig out the reason

jgonzalezdr:
@basvodde: Do yo need any extra information that may help you review this PR? May I help in any way?

basvodde:
For now, just need more time. Been trying to have a short vacation :)

jgonzalezdr:
Then enjoy your vacation! ;-)

include/CppUTestExt/MockCheckedExpectedCall.h
basvodde: I actually do not know. I just know that @arstrube removed all inline functions over time :)

jgonzalezdr: Maybe @arstrube could give his opinion on this. Shall I fix that in this PR, or is it OK to leave it for a short while, taking in account that the next 2 PRs that I've planned are cleanup ones, where actually [that superfluous method is removed](https://github.com/jgonzalezdr/cpputest/commit/1855c48c08d1a50248d467b8900d9938474ea307)?

jgonzalezdr: I'm not sure if this is the reason why @arstrube removed inline implementations, but it might be just because for testing inline methods can't be mocked up. 

In my projects I only accept inline implementations when these are only trivial calls to other overloaded methods for the same function (i.e. just adding default values or simple transformations of parameters). In these cases, it's fine doing it like that because it doesn't affect testability.

arstrube: It's been a long time, and it had to do with the TI Cl2000 compiler. I can't recall what the problem was and, unfortunately, have no longer access to that compiler for testing. The only way to find out would be to look at old commit comments..... I can't completely rule out that it was unexpected runtime behavior, although I am fairly certain it wasn't.

arstrube: @jgonzalezdr I think it's okay to leave it until the method is removed. There were worse problems for the Ti Cl2000 in the past and the only person who noticed them was me :)

arstrube: @jgonzalezdr ...after many moons.

matthewpiatt:
Any update on this stale PR? Still very interested in this major performance enhancement.

basvodde:
Hi all.

Sorry for the delay on this one. I finally spend a couple hours checking this (and checking an alternative implementation). I guess the above implementation is pretty clean. Though the mocking code needs to be cleaned up a bit more in the future.

I'll merge this. Hope everyone can make smaller commits in the future. Sorry for being the bottleneck on this important change.

Thanks @jgonzalezdr !!

matthewpiatt:
Woohoo! Better late than never. Thanks @basvodde

matthewpiatt:
Woohoo! Better late than never. Thanks Bas!

offa:
Great news! Thanks @jgonzalezdr  and @basvodde.

jgonzalezdr:
Thank you @basvodde, I didn't see the merge until today! :-)

I'll do PRs as small as possible for pending work from #1018 so you can review them faster. ;-)

PR #1025, opened 2016-07-18
Cmake C++ Flags fix

offa:
Fix for cmake, which didn't add the C++ compiler flags.

There's currently a build failure of [_cmake clang_](https://travis-ci.org/offa/cpputest/jobs/145564921) – anyone an idea what's the problem there?


arstrube:
no idea :/

jgonzalezdr:
I've sent you a [PR](https://github.com/offa/cpputest/pull/2) on your own branch that fixes the problem.

This happens in old versions of cmake, because internally in check_cxx_source_compiles, which is indirectly called by check_cxx_compiler_flag, something like `if("${VAR}" MATCHES "^${VAR}$")` is performed in the passed variable, and that really messes up things a bit if the variable name (e.g. WARNING_CXX_FLAG_flag_which_is_being_tested) contains an special regex character like "+". 

In newer versions of cmake it's been replaced by `if(NOT DEFINED ${VAR})`, so this problem wouldn't happen, but anyway, the workaround that I propose is necessary to maintain compatibility with older cmake versions.

offa:
> This happens in old versions of cmake, 

I see, this explains why it fails for CI (Cmake 2.8) but works for me (Cmake 3.x).

@jgonzalezdr thank you very much for your help! I've merged your PR.

PR #1028, opened 2016-08-01
Fix example build issue about gcc compatibility

gerryqd:
This adds handling of '-Wno-c++14-compat', and fixes
the build issue if gcc5 is used.

The fix is verified in Cygwin 2.5 with Gcc 5.3.


build/MakefileWorker.mk
```diff
@@ -314,7 +314,7 @@ ifeq ($(CPPUTEST_USE_GCOV), Y)
 	CPPUTEST_CFLAGS += -fprofile-arcs -ftest-coverage
 endif
 
-CPPUTEST_CXXFLAGS += $(CPPUTEST_WARNINGFLAGS) $(CPPUTEST_CXX_WARNINGFLAGS)
+CPPUTEST_CXXFLAGS += $(CPPUTEST_WARNINGFLAGS) $(CPPUTEST_CXX_WARNINGFLAGS)  -Wno-c++14-compat
```
basvodde: This one would need to be in the CPPUTEST_CXX_WARNINGFLAGS

gerryqd: You are correct! I prefer not change in this way. Better to close it and create another pull request.

PR #1040, opened 2016-09-14
Fixing the gtest build and providing support for gtest18

basvodde:
As the title says


offa:
The travis.yml needs such an update too.

PR #1055, opened 2016-11-11
Add CppUTestExt to IAR

dangu:
This patch adds the possibility to compile the CppUTestExt project in the IAR environment.

The IAR EWARM version is now 7.7.


basvodde:
The diff looks a bit weird. Did you have a look at it?

dangu:
Well, I guess some of the weirdness comes from the removed project files? (I ignored the *.ewt and *.dep-files that are automatically generated from the IAR software)

Then there are a lot of diffs in the actual project files as I let the IAR EWARM upgrade them to version 7.7. These files are not really human readable so it doesn't make much sense to try and diff them.

Apart from that I don't see any problems?

PR #1056, opened 2016-11-11
Fix path to open-watcom installer

Andne:
It appears that the path on the server for open-watcom changed recently.  The installer is now in the /install directory instead of the /pub directory.  I have updated the path in the script and my builds on travis-ci are working again.


Andne:
The second commit is to use HTTP instead of FTP.  For whatever reason the watcom FTP server runs at about 10% the speed of the HTTP server.  As a result, the build completes significantly faster when using the HTTP server.

Andne:
@arstrube It seems like CPPU6.EXE may not actually be running.  What is the file-size limit for the executables?  I kind of doubt that's the problem, would expect the linker to error out then, but it is definitely the largest one.  Is something else happening here maybe?

arstrube:
Hi Andy, first of all, many thanks for taking care of this. 

Sadly, my current project has taking me away from C and unit testing and put me into the world of build management and IBM ClearCase. That's why I haven't shown up here recently.

Your observation is right; CPPU6.EXE did not run. It would be nice if this kind of problem would fail the test. I don't know off-hand why it doesn't.

The linker will link a much bigger exe than this without complaint, however, the resulting EXE cannot run. The size limit isn't exactly 640k, either. It looks like in order to fix this, we will have to introduce CPPU8.EXE, because all of the existing exe's are fairly large and it wouldn't be a good idea to try and move stuff around between them.

I haven't worked on CppUTest for a while; hence I'm not sure that my development environment still works as it should!

Andne:
Looks like I was able to re-arrange the objects to allow all executables to run without adding another.  I did re-name some of them to indicate which ones test CppUTest vs. CppUTestExt though.  This is definitely something to keep an eye on though.  Maybe we should pick a known good size as the limit and fail the build if any of the files exceed that size limit?

platforms/Dos/Makefile
```diff
@@ -29,8 +29,9 @@ CXXFLAGS := $(CFLAGS) -xds
 
 .PHONY: all clean
 
-all: CPPU.LIB CPPUX.LIB CPPU1.EXE CPPU2.EXE CPPU3.EXE CPPU4.EXE \
-    CPPU5.EXE CPPU6.EXE CPPU7.EXE
+all: CPPU.LIB CPPUX.LIB \
+     CPPU1.EXE CPPU2.EXE CPPU3.EXE \
+     CPPUX1.EXE CPPUX2.EXE CPPUX3.EXE CPPUX4.EXE
```
arstrube: Hi Andy,

I would recommend not to do this, because it breaks the automatic running of the tests via batch file!

Andne: Oops, I only looked at the shell script, and it seemed to handle this fine.

I assume by batch file you mean ALLTESTS.BAT, to me it looks like it just runs any *.exe file in the current directory.

arstrube: Oh, okay. I guess I changed it at some point.

Still, I would suggest to make it two PRs - the first one for the essential fixes (like the PR I just submitted), and a second one for the name changes. It has been some time since I worked on this; initially I had namings like the ones you now suggest (and they make a lot of sense), then I changed it to simplify things. 

I am actually surprised that ALLTESTS.BAT runs just any .exe; this is not really good in my opinion!

arstrube: And so does the shell script - a well, so be it then.

arstrube:
As for the silent failure, I remember now why.

The exe fails inside DosBox, and I could not find any way to propagate this failure to outside of the DosBox! If you have any idea, that would be great.

arstrube:
In any case, thank you for picking this up and working on it!

PR #1057, opened 2016-11-11
Adding strdup support

felipe-lavratti:
Hello,

Created a PR adding support to `strdup` and `strndup`, addressing issue #3.

I decided __not__ to add a `PlatformSpecificStrlen` abstraction pointer for a few reasons:
  1. it would enforce platforms to include `strings.h` sometimes unnecessary,  
  2. it would be hard to platform code autodetect `strdup`
  3. testing platform code is pain, I kept platform changes to its minimum in this PR.

This way, CppUTest has its own `strlen` implementation that relay on the C standard that all strings are null terminated, if any platform decides to override this behaviour, then it shall not enable `CPPUTEST_HAVE_STRDUP` on config file.

Automake will enable `CPPUTEST_HAVE_STRDUP` automatically if the original `strdup` function is available in the platform.

It's the first time I dig into CppUTest internals and did liked what I saw, very good written code, I hope I kept it in its standards. 

Anything you may need, please feel free to ask.
Felipe


felipe-lavratti:
Hold this PR until I fix the CI checks.

felipe-lavratti:
About coverage (https://coveralls.io/builds/8783613)
It has decreased a little since coverage runs with CMake, where strdup macros are disabled. How do you recommend I approach it?

felipe-lavratti:
Fixed CI checks;
Added CMake checks for `strdup` and left enabled by default to increase code coverage.

basvodde:
Seems to be a conflict now. Usually a bit smaller commits would be prefered. Unfortunately, I tend to take way too long merging large PRs. (sorry people)

felipe-lavratti:
Hi @basvodde, it's a simple conflict, we have both fixed the path of the `travis_ci_build.sh` script, I'm resolving conflicts and pushing right now. Sorry about that.

basvodde:
Added some comments, but merged it anyway. Looks ok.

Thanks.

configure.ac
```diff
@@ -60,7 +60,7 @@ AC_TYPE_LONG_LONG_INT
 
 # Checks for library functions.
 AC_FUNC_FORK
-AC_CHECK_FUNCS([gettimeofday memset strstr])
+AC_CHECK_FUNCS([gettimeofday memset strstr strdup])
```
felipe-lavratti: It depends the level of coverage you want. 
strdup and strndup are both available when the user doesn't mess up with feature_test_macros and restrict itself only with the -std=gnuXX flag to enable gnu extensions. In the worst case the user namespace may end up with one extra symbol in these very specific cases. 
The reason I did not check for both strdup and strndup on configure is to avoid extra macros in code.

What do you recommend me to approach this?

include/CppUTest/MemoryLeakDetectorMallocMacros.h
```diff
@@ -34,18 +36,11 @@ extern void crash_on_allocation_number(unsigned number);
...
- * size_t length = 1 + strlen(str);
- * char* result = (char*) cpputest_malloc_location(length, file, line);
- * memcpy(result, str, length);
- * return result;
- *
- */
+#ifdef CPPUTEST_HAVE_STRDUP
+#define strdup(str) cpputest_strdup_location(str, __FILE__, __LINE__)
+#define strndup(str, n) cpputest_strndup_location(str, n, __FILE__, __LINE__)
+#define CPPUTEST_USE_STRDUP_MACROS 1
```
felipe-lavratti: The CPPUTEST_USE_STRDUP_MACROS definition was only a matter of stardization in the header and tests, since CPPUTEST_USE_MALLOC_MACROS definition is created below, so I felt it was better to stick with this style. 
In my opinion all these test macros could be refactored/simplified after we get CMake to generate a complete Config.h as Autotools do.

Again, I'd be glad to change it to match your criteria.

felipe-lavratti:
Glad to be of any use :)

PR #1058, opened 2016-11-11
My suggestion for fixing the Wcl problems

arstrube:
I suggest to make as few changes as needed.

In the future, if any of the EXE's doesn't run, some of its tests can be moved to CPPU8.EXE


basvodde:
Now this has conflicts :(

arstrube:
Hi Bas, the conflicts are expected. This was an alternative to #1056.

PR #1059, opened 2016-11-13
Match tests folder to src and include folders

Andne:
Nothing functional changed in this one, I'm just fixing a layout quirk that I've wondered about for a while.  The src and include folders have separate subfolders for CppUTest and CppUTestExt while the tests folder has the CppUTest files in it directly and the CppUTestExt files in a subfolder.  I've re-arranged this to put all of the CppUTest test files in tests/CppUTest instead and updated the various project files to point to them there.

As far as I can tell, I've updated all of the main project files, or at least everything that runs on Travis-CI and Appveyor.  There may be some other project under platforms that need updating yet, but I can't do those since I don't have the ability to run those projects.

Thoughts?


basvodde:
Seems pretty ok. Leaving some time to comment as the change is pretty big.

arstrube:
Much appreciated. Been wishing for this a long time!

PR #1069, opened 2016-12-07
Added mac builds on travis

felipe-lavratti:
Now it is right.
Build will fail in two mac builds, without the IEEE fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,44 @@ matrix:
       env: BUILD=cmake_coverage
     - compiler: gcc
       env: BUILD=test_report
+    - compiler: clang
+      os: osx
+      env: BUILD=autotools
+      addons:
+        apt:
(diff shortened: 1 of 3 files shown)


felipe-lavratti:
Yep! Build .11 and .15 are failing on travis, as expected. 

@Bas, would you like me to merge master and resubmit the PR?

Also, how do you fill about the complete duplication of the builds that were running on Linux, to Mac, now. The cost is that it takes a lot of time to Travis run a single Mac build, even more time it takes with all those 10. 

Best, 
Fanl

basvodde:
Looks good and I'll merge it.

I don't think all of the builds are needed on Mac. If we can run the main (long) one then I think we can find most of the mistakes. At least, no need to do CMake and Autotools builds (keep the autotools as it gives most often problems :P)

PR #1083, opened 2017-02-02
Install copier correctly

NikBrg:
These changes fix a bug in CppUMock where copiers are not installed correctly (especially when used with the MockSupportPlugin)


basvodde:
Thanks!

Issue #1070, opened 2016-12-13
Travis CI Builds are way too long


offa:
The **Travis CI builds** are **_too long_**, after a commit it takes **> 2 hours** – not minutes – to get the final result. There are builds taking almost 30 minutes and a total build count of 20 is huge.

------------------------
### Here's an idea:

Some days ago, travis added support for *Cron Job Builds*<sup>1)</sup> (per day / week / month). This allows following splitting:

- **Fast Builds** *(each commit)* — Some fast builds (~ 5 min total) for each commit that *(a)* build the project, *(b)* execute the tests and *(c)* do some checking steps like Coveralls
- **Slow Builds** *(Cron)* — all the extensive other builds, eg. gtest builds, autotools, rare compilers etc.

This will give fast feedback for commits / PR's while maintaining the exhaustive build suite<sup>2)</sup>.

------------------------------
<sup>1)</sup> I use them in some other projects, really nice feature.  
<sup>2)</sup> This also allows some kind of *nightly* builds or version. That's full tested and created by the cron.

basvodde:
Yeah, I guess that makes sense.

Andne:
I think some of the build count comes from the recent additions of Mac OSX.  It's probably not needed to run all of the same build tools as we do on the other environments there, mostly just need to check the compiler.  I think that was mentioned on a different pull request when Mac OSX got added.

That said, it appears that cmake is faster than autotools, so does it make sense to keep those then?  I don't know enough off the top of my head what the `check_all` target does under autotools, so maybe there's more going on there than needs too for the quick build.  It looks like the coverage build runs using the cmake toolchain, so maybe just run that configuration (cmake_coverage) for each compiler and platform only?  That would be 4 builds that run for each commit, serially probably about 5-10 minutes, with parallel builds (seems to be the default behavior) less.

Some questions, do the cron job builds only run if the repository has changed, or do they always run?  What do they run against (my first guess would be master)? Do we have a way to force a full build if the PR is changing stuff that is normally only checked during the cron job build?

offa:
> Some questions, do the cron job builds only run if the repository has changed, or do they always run? What do they run against (my first guess would be master)? Do we have a way to force a full build if the PR is changing stuff that is normally only checked during the cron job build?

[...]

PR #1085, opened 2017-03-10
Reduced Travis CI runtime

offa:
This addresses the very long runtimes of Travis CI builds (#1070). The number of builds is reduced and each build is kept within some minutes.

Currently the total time has been decreased from *2 1/2 hours* to *5 minutes*.

### CI Builds

- [ ] Clang and GCC
 - [ ] C++11 on / off
 - [ ] CMake
 - [ ] Autotools
- [ ] WCL
- [ ] OS X (clang only)
- [ ] GTest
- [ ] Coverage 

### Cron Builds

In addition to the faster CI builds, long running steps are moved into *Cron Builds*.

- [ ] Full check (`make full_check`; autotools only)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ sudo: false
 matrix:
     include:
     - compiler: clang
-      env: BUILD=autotools
-    - compiler: clang
       env: BUILD=cmake CPP11=OFF
     - compiler: clang
       env: BUILD=cmake CPP11=ON
@@ -16,8 +14,19 @@ matrix:
       env: BUILD=cmake CPP11=OFF
     - compiler: gcc
       env: BUILD=cmake CPP11=ON
+
     - compiler: gcc
       env: BUILD=autotools
+      addons:
+        apt:
+          packages:
+            - valgrind
+    - compiler: clang
+      env: BUILD=autotools
+      addons:
+        apt:
+          packages:
+            - valgrind
 
     - compiler: gcc
       env: BUILD=cmake_coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ matrix:
           packages:
             - dosbox
 
+    - compiler: clang
+      os: osx
+      env: BUILD=cmake
+
 
 before_script:
     - export CPPUTEST_BUILD_DIR=$TRAVIS_BUILD_DIR/cpputest_build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: cpp
 
+dist: trusty
 sudo: false
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,18 +7,17 @@ matrix:
     include:
     - compiler: clang
       env: BUILD=autotools
+    - compiler: clang
+      env: BUILD=cmake CPP11=OFF
+    - compiler: clang
+      env: BUILD=cmake CPP11=ON
+
+    - compiler: gcc
+      env: BUILD=cmake CPP11=OFF
+    - compiler: gcc
+      env: BUILD=cmake CPP11=ON
     - compiler: gcc
       env: BUILD=autotools
-
-    - compiler: clang
-      env: BUILD=cmake CPP11=OFF
-    - compiler: clang
-      env: BUILD=cmake CPP11=ON
-
-    - compiler: gcc
-      env: BUILD=cmake CPP11=OFF
-    - compiler: gcc
-      env: BUILD=cmake CPP11=ON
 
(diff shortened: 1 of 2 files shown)


offa:
OS X builds take badly long to spawn. Sometimes > 10 minutes.

offa:
I have configured a cron build for my fork (daily); let's see if it works.

offa:
Looks like the `check_special_situations` target on autotools isn't supposed to run on linux build!?

basvodde:
Could you also run the OSX at least on Chron. We do find problems with that build.

.travis.yml
```diff
@@ -1,96 +1,60 @@
 language: cpp
+
+dist: trusty
```
basvodde: Do we want to put the dist here? I guess that means we need to remember updating it all the time.

.travis.yml
```diff
@@ -1,96 +1,60 @@
...
-- os: linux
-- rvm: '1.9.3'
-- secure: |-
-    P05xUfJVw5YM4hF7hzQLjyMzDD4Q1/fyWP9Uk5aK5VrSWNY99EuxldXI5QK/
-    vA1NkcW49tQW1wQvBlRtdlLNOmUfDP/oiJFXPwNn4dqwOIOEet2P7JO/5hnH
-    MNHlZmGu2WpoZREhOFBfsIhK0IP8mloqLDq2XemBdga/LWygrLU=
-- secure: |-
-    Y/8iNkf6uEbE3qltnM+7mGlCvFWzyttwwRGgVGw1m9xOiUJcobvOImQRU8XZ
-    91dgO+Fz0A3mljqs1sK1OPjpXmFGE1jP/NlotMw0WlDOuSIDjQ4ubwdTNGAw
-    NY53R9ygbIjEmqxHAJm9mOZqxW2hNaoI7TcX6oX248/hLibyx8M=
```
basvodde: Can you keep this still, this is for the deploying (although that doesn't work at the moment)

offa: Sure, here are still not all entries in the travis file. I'm adding things step-by-step.

Commit to re-add this follows …

.travis.yml
offa: Travis supports two distros, trusty is the more recent / up-to-date. There's no updated needed, as travis ships the updates.

offa:
> Could you also run the OSX at least on Chron. We do find problems with that build.

Hmm, have to figure out how to add this as cron-only.

.travis.yml
offa: There are two `secure` entries – should I add both? Is the `rvm` / `os` part before needed too?

.travis.yml
basvodde: Sure, but when there will be a new distro, then making trusty explicit here would not automatically move us to the new one. If we remove it, it would use the latest one always (I assume)

offa: No, it will use the default one that's defined by travis. Atm. that's 12.04 – incl. completely outdated software.

Once there's a new environment, Trusty is still available (maybe it becomes the new default?)

basvodde:
Waiting for MaxOSX Cron job ?

.travis.yml
basvodde: I assume it won't take years... but ok. I'm worried that we are now using the new distro, but because we made it explicit, over time we'll actually use the old distro.

But, I guess it's ok, as long as we keep updating this file.

offa: Once there's a change necessary, I'm here to help! 😄

offa:
I have to find out, how to add a OS X build to Cron's only. If it's added as a usual job, it'll be included the normal builds as well (at least it needs to spawn) – this will increase the feedback time to hours (= bad).

Unfortunately there's no _"add-this-build-if-cron-only"_ setting yet.

.travis.yml
offa: If there's some change in the current behaviour of travis, we still can drop the line.

offa:
An idea to handle this limitation (just an idea yet …):

1. Build the ci jobs as usual
1. If successful merge it into a "nightly" branch
1. Build cron's from the "nightly" branch (full builds, checks, environments)
1. Extension possible: If successful here, push nightly tag and pack a release to that version

basvodde:
Hmm.. ok, though I think they'll fix the spawn time in the future. So, we'd better keep it out for now and not complicate it too much, right? I'll merge this. Thanks!

offa:
Yes, that's the best solution at the moment I guess.

@basvodde If not done yet, you have to add a cron job at travis ([doc](https://docs.travis-ci.com/user/cron-jobs/)).

PR #1087, opened 2017-03-24
Add nothrow delete operator variant overload.

Xiche:
Fixes the failure case caused by this simple test program:

```
#include <iostream>
#include <algorithm>
#include <vector>

#include <CppUTest/CommandLineTestRunner.h>
#include <CppUTest/TestHarness.h>

TEST_GROUP(HelloWorld)
{
};

TEST(HelloWorld, OnlyTest)
{
    std::vector<int> v{0, 0, 3, 0, 2, 4, 5, 0, 7};
    std::stable_partition(v.begin(), v.end(), [](int n){return n>0;});
    for (int n : v) {
        std::cout << n << ' ';
    }
    std::cout << '\n';
}

int main(int argc, char** argv)
{
    return CommandLineTestRunner::RunAllTests(argc, argv);
}
```

### DIFF
--- a/src/CppUTest/MemoryLeakWarningPlugin.cpp
+++ b/src/CppUTest/MemoryLeakWarningPlugin.cpp
@@ -356,9 +356,19 @@ void* operator new(size_t size, const std::nothrow_t&) UT_NOTHROW
     return operator_new_nothrow_fptr(size);
 }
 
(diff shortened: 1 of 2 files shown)


basvodde:
Thanks. I'm a bit worried about compiler support, but I guess we'll see. So far it passes all jobs.

Issue #8, opened 2012-08-08
Improve macro for checking conditions


basvodde:
Currently, the macro CHECK(cond) is provided for checking arbitrary conditions.
A typical use is CHECK(actual>=minimum). Unfortunately, the output of this macro is somewhat limited.
For the example above, the failure message is "CHECK(actual>=minimum) failed."

To get more insight into the problem, I found the following macro very helpful:
# define CHECK_RELATION(first, relop, second)\

CHECK_LOCATION((first) relop (second), (StringFrom(#first #relop #second ", ") \
+StringFrom(first)+StringFrom(#relop)+StringFrom(second)).asCharString(), **FILE**, **LINE**)

Example: CHECK_RELATION(actual, >=, minimum)
Output on failure: "CHECK(actual>=minimum, 0.5>=0.8) failed"

In conclusion, I suggest to add the above macro (or something a bit more elegant) to utestmacros.h.

Also close in SourceForge at:
https://sourceforge.net/tracker/?func=detail&aid=3365842&group_id=199196&atid=968521

PR #1096, opened 2017-05-30
define CHECK_RELATION() macro

dlindelof:
Hi,

I've taken a shot at implementing the `CHECK_RELATION()` macro described in #8.

On failure, the following kind of message will be shown in the output:

```
/Users/dlindelof/Work/cpputest/tests/CppUTest/TestUTestMacro.cpp:400: error:
	Message: evaluates to 0.5 >= 0.8
	CHECK(actual >= minimum) failed
```

Please let me know if this is something you still think could be useful.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -143,11 +143,17 @@
   } }
 
 #define CHECK_COMPARE(first, relop, second)\
+  CHECK_COMPARE_TEXT(first, relop, second, NULL)
+
+#define CHECK_COMPARE_TEXT(first, relop, second, text)\
+  CHECK_COMPARE_LOCATION(first, relop, second, text, __FILE__, __LINE__)
+
+#define CHECK_COMPARE_LOCATION(first, relop, second, text, file, line)\
  { SimpleString conditionString;\
    conditionString += StringFrom(first); conditionString += " ";\
    conditionString += #relop; conditionString += " ";\
    conditionString += StringFrom(second);\
-   UtestShell::getCurrent()->assertCompare((first) relop (second), "CHECK_COMPARE", conditionString.asCharString(), NULL, __FILE__, __LINE__);\
+   UtestShell::getCurrent()->assertCompare((first) relop (second), "CHECK_COMPARE", conditionString.asCharString(), text, __FILE__, __LINE__);\
  }
 
 //This check checks for char* string equality using strcmp.
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -142,6 +142,14 @@
     UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, NULL, file, line); \
   } }
 
+#define CHECK_RELATION(first, relop, second)\
+ { SimpleString conditionString = "evaluates to ";\
+   conditionString += StringFrom(first); conditionString += " ";\
+   conditionString += #relop; conditionString += " ";\
+   conditionString += StringFrom(second);\
+   CHECK_TRUE_LOCATION((first) relop (second), "CHECK", #first " " #relop " " #second, conditionString.asCharString(), __FILE__, __LINE__)\
+ }
+
 //This check checks for char* string equality using strcmp.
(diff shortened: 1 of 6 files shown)


dlindelof:
Hi,

Did you get a chance to have a look at this pull request? Anything you think I should have done differently?

David

basvodde:
I did. I've been wondering whether we want it or not... 

Would love to hear from some others before merging it.

dlindelof:
Ok no problem.

offa:
I know some cases where this could really help. Especially checks where the actual value doesn't matter as long as it is above / below a certain limit. At the moment I'm using `CHECK_TRUE()` / `CHECK_FALSE()` here; if the test is ok, no problem; but on failure it's difficult to get to the root cause quickly, as you don't have much more information to start than 'it failed'.

jgonzalezdr:
As offa, I've also had in many cases to use CHECK_TRUE / CHECK_FALSE to perform comparisons, which is needed many times when doing integration tests or blackbox testing of some complex units, therefore I think it's a good improvement.

However, I don't like the macro name. When I first looked at the PR name, I thought that this macro was to check relations between classes (parent of, subclass of, etc.) or something more strange. I kindly suggest to use something more obvious, like CHECK_COMPARE.

Anyway, good work @dlindelof!!!

include/CppUTest/UtestMacros.h
```diff
@@ -142,6 +142,14 @@
     UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, NULL, file, line); \
   } }
 
+#define CHECK_RELATION(first, relop, second)\
+ { SimpleString conditionString = "evaluates to ";\
+   conditionString += StringFrom(first); conditionString += " ";\
+   conditionString += #relop; conditionString += " ";\
+   conditionString += StringFrom(second);\
+   CHECK_TRUE_LOCATION((first) relop (second), "CHECK", #first " " #relop " " #second, conditionString.asCharString(), __FILE__, __LINE__)\
```
jgonzalezdr: The second parameter to CHECK_TRUE_LOCATION should be the same than the macro ("CHECK_RELATION").

include/CppUTest/UtestMacros.h
```diff
@@ -142,6 +142,14 @@
     UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, NULL, file, line); \
   } }
 
+#define CHECK_RELATION(first, relop, second)\
+ { SimpleString conditionString = "evaluates to ";\
```
jgonzalezdr: The message printed with this string is a bit too scarce... Maybe "The actual comparison check evaluates to \"" + ... + "\"" would be more informative to the user. 

Anyway, using the text message to handle this additional information is not good for me, the text message should be reserved for use by the user. It would be better if you created a new subclass of TextFailure and a new method assertComparison() in UTestShell to pass the actual parameter evaluation, such that it's printed in the failure string and not in the user message.

You should then also add a xxx_TEXT macro that takes a user message.

basvodde:
I agree, CHECK_COMPARE might be better, CHECK_RELATION is a bit strange. Or any other suggestions for macro name?

Also @dlindelof, could you respond to the comments made?

I'll integrate this when we're comfortable with it. Much appreciated.

dlindelof:
Yes I'm on it. I've changed the macro name and the second parameter to `CHECK_TRUE_LOCATION`. But I'm not sure what @jgonzalezdr meant about the use of the text message, or about subclassing `Tex(s?)tFailure`. I'll need to dig into the code again. I think I should have something by beginning of next week.

basvodde:
Looking forward to it.

dlindelof:
Hello,

I've applied your suggestions, thanks for pointing me in the right direction. Please have a look and let me know if this is fine now. If yes, I'll also update the documentation.

basvodde:
Thanks. Merged.

dlindelof:
Thanks for merging. I'd like to update the documentation, but I cannot find it. Can you show me where it is?

And would this close issue #8 ?

basvodde:
The site is at:
https://github.com/cpputest/cpputest.github.io

I closed #8. That was an old one :)

PR #1097, opened 2017-05-31
IAR: Moving test source files to fix error introduced with 02513e80353…

dangu:
Pull request #1059 broke the IAR build as the test source files were moved. Now the IAR project reflect these changes.

Also increasing the virtual flash size to make room for more tests in the IAR environment.


basvodde:
Thanks!

PR #1103, opened 2017-06-17
allow building CppUTest as a sub-project of a cross-platform application (fixes #1102)

### LINKED ISSUE #1102: CMake build files aren't engineered to allow building CppUTest as a sub-project of a cross-platform application
The CMake files in this repository aren't immediately ready for cross-platform builds. The top level CMakeLists.txt discovers platform-specific build options based on the build host; but these may not apply to the target platform. A project needing to embed CppUTest for a different target (other than the host platform) cannot therefore simply do a `add_subdirectory(<path/to/cpputest>)`. Since the meat of the build is defined within src/CppUTest/CMakeLists.txt, a reasonable alternative is for the enclosing project to do `add_subdirectory(<path/to/cpputest/src/CppUTest>)`.

I propose the following version for src/CppUTest/CMakeLists.txt to ease embedding CppUTest into an external project:

```cmake
add_library(CppUTest
        CommandLineArguments.cpp
        MemoryLeakWarningPlugin.cpp
        TestHarness_c.cpp
        TestRegistry.cpp
        CommandLineTestRunner.cpp
        SimpleString.cpp
        TestMemoryAllocator.cpp
        TestResult.cpp
        JUnitTestOutput.cpp
        TeamCityTestOutput.cpp
        TestFailure.cpp
        TestOutput.cpp
        MemoryLeakDetector.cpp
        TestFilter.cpp
        TestPlugin.cpp
        TestTestingFixture.cpp
        SimpleMutex.cpp
        Utest.cpp
)

if (CPP_PLATFORM)
    target_sources(CppUTest
        PRIVATE
            ../Platforms/${CPP_PLATFORM}/UtestPlatform.cpp
    )
endif(CPP_PLATFORM)

target_include_directories(CppUTest
    PUBLIC
        ../../include
)

if (WIN32)
    target_link_libraries(CppUTest winmm.lib)
endif (WIN32)
install(FILES ${CppUTest_headers} DESTINATION include/CppUTest)
install(TARGETS CppUTest
    EXPORT CppUTestTargets
    RUNTIME DESTINATION bin
    LIBRARY DESTINATION lib
    ARCHIVE DESTINATION lib)
```

I've made the following changes:

* add a section to declare the include folder as a PUBLIC property. This means that any project depending on this CMakeLists.txt would automatically inherit the include folder.

```cmake
target_include_directories(CppUTest
    PUBLIC
        ../../include
)
```

* use a relative path for referring to include. This removes the need for an external definition of the variable CppUTestRootDirectory.

* remove `STATIC` from ```add_library(CppUTest STATIC ${CppUTest_src} ${CppUTest_headers})```. There is no need to be specific about building a STATIC library. The enclosing project could choose alternatives to static linking.

* remove the unconditional addition of `../Platforms/${CPP_PLATFORM}/UtestPlatform.cpp` to the sources that get compiled into the CppUTest library. It may be that the enclosing project wants to provide its own Platform.cpp without re-using one of the standard Platforms or dropping in a new one under `../Platforms/` . Platform.cpp could be project-specific, and therefore a user should have the flexibility to add it to the build externally.

* `add_library(CppUTest ...)` should not need to mention the headers as dependencies.

Finally, I request that there should be a README containing an example CMakeLists.txt showing how CppUTest could be embedded into a project.

The CMake files in this project could also benefit from modern cmake idioms. Refer to https://github.com/onqtam/awesome-cmake#resources and https://www.youtube.com/watch?v=bsXLMQ6WgIk

rgrover:
I have made the following changes:

* add a section to declare the include folder as a PUBLIC property. This means that any project depending on this CMakeLists.txt would automatically inherit the include folder.

```cmake
target_include_directories(CppUTest
    PUBLIC
        ../../include
)
```

* use a relative path for referring to include. This removes the need for an external definition of the variable CppUTestRootDirectory.

* remove `STATIC` from ```add_library(CppUTest STATIC ${CppUTest_src} ${CppUTest_headers})```. There is no need to be specific about building a STATIC library. The enclosing project could choose alternatives to static linking.

* remove the unconditional addition of `../Platforms/${CPP_PLATFORM}/UtestPlatform.cpp` to the sources that get compiled into the CppUTest library. It may be that the enclosing project wants to provide its own Platform.cpp without re-using one of the standard Platforms or dropping in a new one under `../Platforms/` . Platform.cpp could be project-specific, and therefore a user should have the flexibility to add it to the build externally.

* `add_library(CppUTest ...)` should not need to mention the headers as dependencies.

* Finally, I provide a README containing an example CMakeLists.txt showing how CppUTest could be embedded into a project.

### DIFF
--- a/src/CppUTest/CMakeIntegration-README.md
+++ b/src/CppUTest/CMakeIntegration-README.md
@@ -1,0 +1,24 @@
+It is possible to integrate CppUTest as a sub-module of an enclosing CMake
+project. This may be useful if CppUTest is being built for a target platform
+other than that of the development host. The following is an example how an
+external project can refer to this CMakeLists.txt to build CppUTest as a
+library and include it as a target dependency.
+
+```cmake
+cmake_minimum_required(VERSION 3.7)
+project(trying_CppUtest)
+
+SET(CppUTestRootDirectory /path/to/cpputest)
+
+# Either set CPP_PLATFORM to one of the provided platforms under
+# ${CppUTestRootDirectory}/src/Platforms/, or provide a project-specific
+# platform.cpp (as shown below)
+add_subdirectory(${CppUTestRootDirectory}/src/CppUTest CppUTest)
+target_sources(CppUTest
+	PRIVATE
+		${PROJECT_SOURCE_DIR}/UtestPlatform.cpp
+)
+
+add_executable(trying_CppUtest main.cpp)
+target_link_libraries(trying_CppUtest CppUTest)
+```
--- a/src/CppUTest/CMakeLists.txt
+++ b/src/CppUTest/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(CppUTest_src
+add_library(CppUTest
         CommandLineArguments.cpp
         MemoryLeakWarningPlugin.cpp
         TestHarness_c.cpp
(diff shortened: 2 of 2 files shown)


basvodde:
Thanks.

PR #1107, opened 2017-07-06
Reset dotCount_ to align progress indicators for repeat runs

tylerjsmith:
If you run a test set with "-r#", the progress indicator dots don't line up from
one run to another. It makes it difficult to look for differences when the output
is different every time even though the tests run the same.


basvodde:
In either case, would you mind adding a test also?

Issue #1079, opened 2017-01-26
Cannot compile latest release nor latest commit as x64 Windows executable


Adnn:
When compiling under Windows, warnings are treated as error. A previous PR, merged in 6b06b13, fixed the warnings for Windows x32.

Yet, it is still impossible to compile as a 64 bits Windows executable, because of several C4312 errors (on 64bits Windows, converting an unsigned int to a pointer raise a warning, because the pointer is of greater size...)

There is also a warning C4723 (potential divide by 0) in IEEE754PluginTest_c.c.

andrew-elder:
Any updates on this issue?

basvodde:
Not yet, I think.

Andne:
I have some code on my machine to fix the divide by zero, but unfortunately it's causing MinGW to crash on appveyor, so I've been trying to figure that out.

andrew-elder:
Ok, I have reverted my fix, applied latest from master (thanks @Andne) and the issue remains  fixed. I think this issue can be closed out.

PR #1121, opened 2017-10-08
Fix float warning

Andne:
Visual C++ detects the divide by zero at compile time unless the variables are volatile.  This fixes one of the warnings reported in #1079.

Changing the build image was needed in order to keep the MinGW32 build working. It's not a permanent fix, but I'm not sure what will be yet.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 version: 3.7.0-ci{build}
 
-image: Visual Studio 2015
+image: Previous Visual Studio 2015
 
 cache:
   - C:\ProgramData\chocolatey\bin -> appveyor.yml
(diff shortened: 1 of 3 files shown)


Andne:
Travis is failing because it can't download OpenWatcom.
Here's an older build with the same change that worked: https://travis-ci.org/Andne/cpputest/builds/281660313

basvodde:
Anything that needs to do on the Watcom download?

Andne:
I'm not sure, it looks like a network error to me, does their ftp server go offline sometimes for maintenance?

There appears to be a fork on GitHub called open-watcom V2 (http://open-watcom.github.io/open-watcom/) that may be a more reliable source.  That would be one possible option.

GitHub support LFS now, so maybe if someone can get it downloaded we could keep the zip file in the repository using that?  Then it wouldn't have to download anymore.  Not ideal I suppose, I think I like the idea of using the V2 fork if that still accomplishes the same goal.

basvodde:
Yeah, zip is not good, but it would be good to have it more reliable than now...

Issue #783, opened 2015-10-05
Two samples fail to build on 64bit compiler (MSVC)


ZahlGraf:
Hi

There are two samples that does not build due to 64bit issues:
SimpleStringTest and TestUTestMacro

In both files an 32 bit or 16 bit value is casted to a pointer. MSVC2015 is complaining about this, because of the different bit size of the variables. You can workaround this error, when you first cast to (size_t) and then to the pointer type. 

This is a patch

``` diff
 tests/SimpleStringTest.cpp | 2 +-
 tests/TestUTestMacro.cpp   | 6 +++---
 2 files changed, 4 insertions(+), 4 deletions(-)

diff --git a/tests/SimpleStringTest.cpp b/tests/SimpleStringTest.cpp
index 9e087fe..c1a2649 100644
--- a/tests/SimpleStringTest.cpp
+++ b/tests/SimpleStringTest.cpp
@@ -614,7 +614,7 @@ TEST(SimpleString, _64BitAddressPrintsCorrectly)

 IGNORE_TEST(SimpleString, _64BitAddressPrintsCorrectly)
 {
-    char* p = (char*) 0xffffffff;
+    char* p = (char*)(size_t)(0xffffffff);
     SimpleString expected("0x123456789");
     SimpleString actual = StringFrom((void*)&p[0x2345678A]);
     STRCMP_EQUAL(expected.asCharString(), actual.asCharString());
diff --git a/tests/TestUTestMacro.cpp b/tests/TestUTestMacro.cpp
index 0929946..8e58a26 100644
--- a/tests/TestUTestMacro.cpp
+++ b/tests/TestUTestMacro.cpp
@@ -485,13 +485,13 @@ TEST(UnitTestMacros, FailurePrintHexOutputForPointers)

 TEST(UnitTestMacros, POINTERS_EQUALBehavesAsProperMacro)
 {
-    if (false) POINTERS_EQUAL(0, (void*) 0xbeefbeef)
-    else POINTERS_EQUAL((void*)0xdeadbeef, (void*)0xdeadbeef)
+    if (false) POINTERS_EQUAL(0, (void*)(size_t)0xbeefbeef)
+    else POINTERS_EQUAL((void*)(size_t)0xdeadbeef, (void*)(size_t)0xdeadbeef)
 }

 IGNORE_TEST(UnitTestMacros, POINTERS_EQUALWorksInAnIgnoredTest)
 {
[...]

PR #1122, opened 2017-10-12
MSVC2015 64-bit compile fix for pointers, see #783

andrew-elder:


andrew-elder:
Now I see this fix may be addressed here https://github.com/cpputest/cpputest/pull/847 ?

andrew-elder:
Is travis build broken? What happened to it?

basvodde:
Yup, the fix doesn't compile on clang....

andrew-elder:
Is there a better approach to fix the MSVC2015 64-bit compile?

Andne:
I have a solution for the SimpleStringTest.cpp warning, but it requires that I assume that `CPPUTEST_USE_LONG_LONG` will always be enabled on 64-bit platforms with 32-bit longs (Windows mainly I think).

@basvodde Are you ok with me making this assumption?  Seems like it ought to be valid since if you have 64-bit pointers and 32-bit longs you need some way to create a 64-bit integer.  This also fixes a test that has two forms currently.

Another solution I'm finding is to avoid any possibility of sign-extension occurring by making sure the most significant bit is 0.  Changing the constants to be less than 0x80000000 seems to work pretty well.

andrew-elder:
Thanks for adding to this discussion @Andne. It did seem that making sure CPPUTEST_USE_LONG_LONG is defined would be another solution to the problem.

basvodde:
@Andne At this moment, I can't think of a reason why we can't assume it.

andrew-elder:
Closing this pull request because the same problem is solved more cleanly in https://github.com/cpputest/cpputest/pull/1124.

PR #1123, opened 2017-10-14
Fix MinGW 32-bit crashes on appveyor, some minor CMake cleanup too

### LINKED ISSUE #1120: MinGW 32-bit tests crash on Appveyor
After the last update in Appveyor, the 32-bit MinGW builds are unable to runs tests.  I have been able to reproduce the issue locally by using CMake 3.9.1 (I was previously using CMake 3.4.3).

It appears that something changed between CMake 3.4.3 and CMake 3.9.1 regarding link libraries, the new version is linking in gcc_eh while the old version removed it from the list.  I've managed to manually tweak the library list and removing gcc_eh from it does appears to allow the executables to work again.

Unfortunately, it appears that the change that caused the addition of gcc_eh is upstream in CMake itself (https://github.com/Kitware/CMake/commit/d59e3509850329096a7d6dfbc037d721b70034c9).  I also don't understand why linking in gcc_eh is such a problem.

Andne:
Fixes #1120 

I finally found a more permanent fix than using the old image on appveyor.  By forcing static linkage I'm able to get the right set of libraries to be used and avoid any funny conflicts in the exception handling.  Now all of the appveyor builds are working properly again and should continue to as they update software more.

At present I've only applied my fix when running under MinGW, since the version of CMake on travis-ci is so old, I'm unable to tell if any changes need to be made there as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,13 +20,20 @@ if(HAVE_STRDUP)
   add_definitions(-DCPPUTEST_HAVE_STRDUP=1)
 endif(HAVE_STRDUP)
 
-# Apply workaround for MinGW timespec redefinition (pthread.h / time.h)
 if (MINGW)
+    # Apply workaround for MinGW timespec redefinition (pthread.h / time.h)
     include(CheckStructHasMember)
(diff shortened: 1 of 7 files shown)


basvodde:
Thanks!

PR #1124, opened 2017-10-16
Fix warnings on x64, add to appveyor

Andne:
This is my take at what was being done in #1122.  Most of the fixes are just unsetting the most-significant bit/byte of the constant so that sign-extension doesn't need to be performed.  The SimpleString tests were fixed by assuming that `long long` always exists on 64-bit, so therefore a test that had two different forms depending on `sizeof(long)` didn't need to have the two forms anymore.

I've also added the x64 build to Appveyor.  It only seems to work on v140 (VS2015) for some reason, the v100 (VS2010) build gets an error about missing winmm.lib.  I suspect there may be an issue with the install on Appveyor, I do not get that error locally.  Also, the warnings that are being fixed don't appear on v100 anyways.

@andrew-elder Take a look at this, does it fix things for you?

### DIFF
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -151,6 +151,7 @@
       <WarningLevel>Level3</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <DebugInformationFormat>OldStyle</DebugInformationFormat>
+      <TreatWarningAsError>true</TreatWarningAsError>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -185,6 +186,7 @@
       <WarningLevel>Level3</WarningLevel>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <DebugInformationFormat>OldStyle</DebugInformationFormat>
+      <TreatWarningAsError>true</TreatWarningAsError>
     </ClCompile>
     <ResourceCompile>
       <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
--- a/CppUTest.vcxproj
+++ b/CppUTest.vcxproj
@@ -5,9 +5,17 @@
       <Configuration>Debug</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
     <ProjectConfiguration Include="Release|Win32">
       <Configuration>Release</Configuration>
       <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
     </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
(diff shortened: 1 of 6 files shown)


andrew-elder:
@Andne - yes this seems to work for me! Who does the merge?

basvodde:
The removal of the first bit comparison feels like quite a hack to me. I'm uncomfortable with that and do not fully understand why it would work. Before I understand it, I won't merge it :)

Andne:
For some reason Visual C++ uses sign extension when converting from 32-bit values to 64-bit pointers.  They warn about the operation here: https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(C4312)&rd=true

I don't know why they are performing sign-extension on a *known* unsigned value, known given that the warning is "'type cast': conversion from 'unsigned int' to 'void *' of greater size.  Even adding the suffix `U` to force the value to be unsigned doesn't help.  At least they recognize that sign-extension doesn't change a value when the upper-most bit is a 0 I guess.  The only solutions appear to be to use a number that won't change value during sign-extension or use a constant that is already the right size.  Unfortunately the second requires declaring the values as `unsigned long long` (`ULL`) which won't be portable to 32-bit systems.  Since Visual C++ uses a 32-bit `long` value, it doesn't work to change the literal to explicitly be `unsigned long`, just changes the warning message slightly.

andrew-elder:
So, are we at a "stalemate" then?

basvodde:
@Andne What would the Pragma solution look like?

Andne:
This could be done either function-level or file-level.  Technically it can be line-by-line where a slightly different pragma that doesn't require the push/pop could be used, but I think that would get too messy.  I prefer function-level, otherwise you lose visibility too easier and it increases the chance of not getting a legitimate warning.

```
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable:4312)
#endif

TEST(UnitTestMacros, POINTERS_EQUALBehavesAsProperMacro)
{
    if (false) POINTERS_EQUAL(0, (void*) 0xbeefbeef)
    else POINTERS_EQUAL((void*)0xdeadbeef, (void*)0xdeadbeef)
}

#ifdef _MSC_VER
#pragma warning(pop)
#endif
```

basvodde:
Can we add the pragma in the config header and only for 64 bit?

That way the file isn't polluted with all the pragmas that are irrelevant for other platforms.

andrew-elder:
We also know that using (void *)(size_t)0xdeadbeef works for MSVC 64-bit. What do you think about creating a macro to wrap the const pointer declaration?

Andne:
In theory yes, however that would suppress the warning throughout the code base which I don't think is particularly desirable.  The warning can be ignored in the spots that are triggering it because we know that the pointers are never dereferenced (otherwise creating them from random values wouldn't work at all).  Suppressing it everywhere potentially would hide a spot where the warning truly indicates a problem, probably resulting in a program crash.

basvodde:
@andrew-elder Yeah, I'd prefer a macro to_void_pointer(0xbeefbeef) or so then, that might not be a bad idea.

Andne:
I changed my fix to use a macro instead - `to_void_pointer` and `to_func_pointer`.  The macros are local to the test file that was causing the error and triggered by a 64bit machine with 32 bit longs.

Instead of double-casting, I used the preprocessor concatenation operator (`##`) to declare the literal values as `unsigned long long`, which also prevents the warning from being reported.  When the build is not 64-bit with 32-bit longs, the values are declared as `unsigned long`, which should be the pointer size on most machines I think.

I also enabled the warnings as errors option now that all of the warnings are cleaned up.  Hopefully stuff won't sneak back in with that turned on and it's active for other platforms already as far as I can tell.

andrew-elder:
I like the change. I'm going to assume that lowercase #define macros don't break Cpputest coding conventions... @Andne - thank you for working on this.

andrew-elder:
@basvodde - are the changes proposed here ok? Are you planning to merge it?

basvodde:
Yup, this is better. We might want to put the macro somewhere else in the future, but lets keep it here for now.

Issue #1126, opened 2017-10-27
Clang 5.0 Static Analyzer warnings


vronchetti:
8 warnings generated by Clang 5.0 scan-build:

[cpputest_clang_5_0_scan_build.zip](https://github.com/cpputest/cpputest/files/1422150/cpputest_clang_5_0_scan_build.zip)

I suspect most need only to be suppressed, e.g.:

| tests/CppUTest/MemoryOperatorOverloadTest.cpp
-- | --
Warning: | line 52, column 5

<img width="542" alt="screen shot 2017-10-27 at 9 03 14 am" src="https://user-images.githubusercontent.com/8479744/32105243-c8d78248-baf5-11e7-95f4-f24cdfc2dc06.png">

How to proceed?

vronchetti:
Is it reasonable to expect no static analyzer warnings from cpputest?  Thank You!

basvodde:
Reasonable. And the analyzer is very right. However, it is testing exactly what it warns for, so I do not know how to remove the warning.

basvodde:
I guess most of these been fixed...

basvodde:
@vronchetti Double checked them. It is not reasonable to expect no static analyzer warnings as some of these warnings are there on purpose. E.g. a method that purposely crashes the application. The other ones seem all fixed. Closing this.

PR #1128, opened 2017-10-31
Clang Static Analyzer - "uninitialized argument value" warning fix

andrew-elder:
Another partial fix for issue #1126.


andrew-elder:
Comments on this one?

basvodde:
I've been wondering about this fix and whether it would cause portability problems. So far, all builds seem to pass so I'll integrate it. Thanks!

PR #1133, opened 2017-12-03
Fix build cpp u test with gcc instruction

PaulBussmann:
Please merge .gitignore and build instruction

### DIFF
--- a/README_InstallCppUTest.txt
+++ b/README_InstallCppUTest.txt
@@ -8,7 +8,9 @@
 2. Build CppUTest and examples
 
 2a. For unix/gcc (including cygwin)
- > cd <someDirectory>/CppUTest/cpputest_build
+ > cd <someDirectory>/CppUTest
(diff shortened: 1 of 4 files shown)


README_InstallCppUTest.txt
```diff
@@ -8,7 +8,9 @@
 2. Build CppUTest and examples
 
 2a. For unix/gcc (including cygwin)
- > cd <someDirectory>/CppUTest/cpputest_build
+ > cd <someDirectory>/CppUTest
+ > ./autogen.sh
+ > cd build
```
basvodde: This shouldn't be there as the packaged version of Cpputest comes with generated configure script. It would need to be changed so that is makes a distinction between the package and the git clone.

basvodde:
Added cpputest_build instead. Rest of changes no need, so made a separate PR for that. Closing this one.

PR #1136, opened 2018-02-18
Simplified MockSupport::expectNoCall() method implementation.

jgonzalezdr:
MockSupport::expectNoCall() now uses a multi-matching expected call with zero expected calls, thus avoiding the usage of a dedicated list to keep track of "unexpectations" and therefore simplifying the architecture of the MockSupport class.


jgonzalezdr:
@basvodde: I know you're very busy, but this PR is pretty simple and I think it's a good improvement to simplify internal mechanics for expectNoCall(), can you have a look at it? Do you need any help or explanation from my side?

basvodde:
Ok. Sorry for delay. Thanks for the contribution.

PR #1137, opened 2018-02-20
Fix clang 5.0.0 errors

jgonzalezdr:
This PR fixes the "errors" (warnings really) raised by Clang 5.0.0 and later, which make the cpputest builds to fail.

The PR is long and boring, but the changes are trivial, so it should not be difficult to review.

I've added also Visual Studio 2012 and 2013 to CI builds, as I've found while fixing clang errors that VS2010's implementation of C++11 is not "complete" (it doesn't allow override specs in destructors), so as builds are pretty fast I thought it would be better to ensure that it builds properly with these toolsets too.

### DIFF
--- a/src/CppUTest/TestPlugin.cpp
+++ b/src/CppUTest/TestPlugin.cpp
@@ -90,7 +90,7 @@ TestPlugin* TestPlugin::getNext()
 }
 TestPlugin* TestPlugin::removePluginByName(const SimpleString& name)
 {
-    TestPlugin* removed = 0;
+    TestPlugin* removed = NULLPTR;
     if (next_ && next_->getName() == name) {
         removed = next_;
         next_ = next_->next_;
@@ -150,7 +150,7 @@ void SetPointerPlugin::postTestAction(UtestShell& /*test*/, TestResult& /*result
 //////// NullPlugin
 
 NullTestPlugin::NullTestPlugin() :
-    TestPlugin(0)
+    TestPlugin(NULLPTR)
 {
 }
 
--- a/src/CppUTest/TestPlugin.cpp
+++ b/src/CppUTest/TestPlugin.cpp
@@ -62,10 +62,10 @@ void TestPlugin::runAllPostTestAction(UtestShell& test, TestResult& result)
 
 bool TestPlugin::parseAllArguments(int ac, char** av, int index)
 {
-    return parseAllArguments(ac, const_cast<const char**> (av), index);
+    return parseAllArguments(ac, const_cast<const char *const *> (av), index);
 }
 
-bool TestPlugin::parseAllArguments(int ac, const char** av, int index)
+bool TestPlugin::parseAllArguments(int ac, const char *const *av, int index)
 {
     if (parseArguments(ac, av, index)) return true;
     if (next_) return next_->parseAllArguments(ac, av, index);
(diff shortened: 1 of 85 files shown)


jgonzalezdr:
The appveyor fail is due to the usual error when installing MinGW with chocolatey, but otherwise it would have passed all checks (see https://github.com/jgonzalezdr/cpputest/commits/fixClangErrors2).

src/CppUTestExt/MockSupport.cpp
```diff
@@ -449,7 +449,7 @@ MockSupport* MockSupport::getMockSupportScope(const SimpleString& name)
 
     if (hasData(mockingSupportName)) {
         STRCMP_EQUAL("MockSupport", getData(mockingSupportName).getType().asCharString());
-        return (MockSupport*) getData(mockingSupportName).getObjectPointer();
+        return static_cast<MockSupport*>( const_cast<void*>( getData(mockingSupportName).getObjectPointer() ) );
```
basvodde: I don't think we are using 'new' style casts in the CppUtest codebase. Could you check this?

src/CppUTestExt/MockSupport.cpp
```diff
@@ -461,8 +461,8 @@ MockSupport* MockSupport::getMockSupportScope(const SimpleString& name)
 MockSupport* MockSupport::getMockSupport(MockNamedValueListNode* node)
 {
     if (node->getType() == "MockSupport" && node->getName().contains(MOCK_SUPPORT_SCOPE_PREFIX))
-        return  (MockSupport*) node->item()->getObjectPointer();
-    return NULL;
+        return static_cast<MockSupport*>( const_cast<void*>( node->item()->getObjectPointer() ) );
```
basvodde: Another one

src/CppUTestExt/MockSupport.cpp
jgonzalezdr: I've put the static cast with the "new" style just for homogeneity with the const cast. I think that const casts can only be performed with the "new" style, but I'll check anyway if clang accepts as const cast a cast that maintains the type but removes only the const expression.

jgonzalezdr: Confirmed, doing a 2-step conversion is not accepted as a const cast by clang:
image

Maybe it's the moment to being to introduce "new" style casts... :sweat_smile:

jgonzalezdr: As the necessity to perform a const cast is in fact due to the lack of differentiation between const and non-const objects in MockSupport auxiliary data model, the const cast could be totally unnecessary if PR #1141 is merged. In this case I could revert the changes due to const casts, as Clang should not moan any more.

basvodde: Merged. Can we now remove the new style casts again?

jgonzalezdr:
@basvodde: Removed the const_cast<> stuff, please check it now.

src/CppUTest/TestPlugin.cpp
```diff
@@ -62,10 +62,10 @@ void TestPlugin::runAllPostTestAction(UtestShell& test, TestResult& result)
 
 bool TestPlugin::parseAllArguments(int ac, char** av, int index)
 {
-    return parseAllArguments(ac, const_cast<const char**> (av), index);
+    return parseAllArguments(ac, const_cast<const char *const *> (av), index);
```
basvodde: Still have one. But it seems there was already a const_cast here, so I was wrong!

PR #1138, opened 2018-02-20
Fixed compilation using CMake with Clang + MinGW bintools in Windows

jgonzalezdr:
The workaround to fix CMake + MinGW compilation breaks CMake + Clang + MinGW bintools compilation, so it's fixed to apply only if the compiler is not Clang (for some dark reason CMake flags Clang as a MinGW compiler).

Besides that I've added a couple of toolchain files for CMake to simplify using this toolchain combination, and added it to CI builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,10 @@ environment:
     PlatformToolset: 5.3.0
   - Platform: MinGW64
     PlatformToolset: 5.3.0
+  - Platform: MinGWClang64
+    PlatformToolset: 5.3.0
+  - Platform: MinGWClang32
+    PlatformToolset: 5.3.0
   - Platform: Win32
     PlatformToolset: v90
   - Platform: Win32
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -30,7 +34,7 @@ environment:
     PlatformToolset: v140
 
 install:
-- ps: if ($env:Platform -like 'MinGW*') { choco install mingw --version $env:PlatformToolset $( if ($env:Platform -like '*32') { Write-Output --forcex86 } ) }
+- ps: if ($env:Platform -like 'MinGW*') { choco install mingw --version $env:PlatformToolset $( if ($env:Platform -like '*32') { Write-Output --forcex86 --params /exception:dwarf } ) }
```
Andne: Is there a reason you explicitly set the exception model?

jgonzalezdr: Yes, because clang uses this exception model (DWARF), so when linking with MinGW's linker it will fail complaining that some functions are missing. So, as far as I know, the Clang + MinGW toolchain combination works only with MinGW DWARF exception model toolchain.

Issue #1019, opened 2016-07-07
Add support to ignore object in mock expectation


MainRo:
There are cases where it would be convenient to ignore the object of a mock when checking expectations. This could be exposed in a similar ways than ignoreOtherParameters, via a new "ignoreObject" method in MockExpectedCall. 

I can work on this feature. Tell me if doing it this way would be ok or not.

Romain.

basvodde:
Would it be ignoreOtherObjects?

MainRo:
Since there can be only one object I think "ignoreObject" is a more appropriate name. But if you prefer "ignoreOtherObjects" for consistency with the existing methods, it is fine.

basvodde:
Your right. ignoreObject.

jgonzalezdr:
This would be a very useful addition. I really just needed this feature once, and I worked around it by using conditional compilation to actually call onObject() or not depending on whether it had to be ignored or not. But having this ignoreObject() method would be much more convenient.

MainRo:
ok I will work on that and submit a PR.

lemtronix:
Great idea @MainRo!  This is definitely something our team would use.  If a developer could help guide me in the right direction, I'd try implementing it.

jgonzalezdr:
I've had also the need of being able to ignore the objects passed to the actual call, so I have implemented that in #1140.

Instead of adding a new member function to the expected call to indicate that the specific object shall be ignored, I've just made the framework to ignore objects unless an explicit call to `onObject()` is performed passing a specific object.

The rationale to avoid a new `ignoreObject` function similar to the `ignoreOtherParameters` function is that with parameters it is necessary to explicitly indicate that parameters are ignored instead of silently ignoring them, in order to differentiate calls to overloaded functions with different parameters (otherwise there would be no way to for example indicate that we expect a call to the overloaded version with 3 parameters instead of a version with 2 parameters). However, it has no sense that a free function and a member function have the same name, therefore if `onObject` is called on an actual call, there is no doubt that it is a member function, and then it's just a matter of checking if the call is expected on a specific object or not.

basvodde:
Assume this is done now, closing the issue.

PR #1140, opened 2018-02-21
Made expectation objects optional.

jgonzalezdr:
With this modifications, if onObject() is called on a expected call, it is considered that the expectation matches only the specific object passed; but if it's not called, it's considered that the expectation matches any object.

Therefore, when onObject() is called on an actual call, both the expectations that target the passed object and the expectations that do not target a specific object will match.

This is a solution for #1019.


basvodde:
Hrm, the pull request has conflicts now with the previous one. Could you resolve these?

jgonzalezdr:
@basvodde: Conflicts fixed.

Issue #34, opened 2012-08-24
Using the "template trick" for placement new


basvodde:
Using this implementation for placement new:
http://nvwa.cvs.sourceforge.net/viewvc/nvwa/nvwa/debug_new.h?view=markup

Would allow for placement new to be overridden without a compiler warning (with debug info). It feels a bit awkward though.

It might be interesting to keep it as an option.

A longer discussion related to this can be found here:
https://groups.google.com/forum/?fromgroups=#!topic/cpputest/afbEk_2ey2Q

bryceschober:
FYI, the updated source link for the NVWA debug_new.h is:
https://github.com/adah1972/nvwa/blob/master/nvwa/debug_new.h

Did anyone propose a version using this trick that could be recreated in a pull request?

basvodde:
It would make it more complex as it would need to be optional.

I am interested in the stack trace functionality they seem to have?

jgonzalezdr:
@basvodde Stack trace functionality is based on GCC's libc backtrace funcionality or in Win32 on CaptureStackBackTrace. Nwva also uses GCC builtin functions to get the return address of functions. Return address to symbol mapping is done in nwva by invoking external tools (e.g. addr2line).

I experimented around some time ago with GCC's backtrace functionality to avoid using the new macro in MemoryLeakDetectorNewMacros.h that breaks STL, and it was possible to just override new/delete operators (not using file and line preprocessor info), get the return address from the stack, and even get the function name without invoking external tools. Performance however was not very good.

It seems also feasible to do it using Win32' CaptureStackBackTrace, and another option is using libunwind. All major "host" platforms could be covered, but it seems there is no "universal" solution.

If there is interest, I can push my code.

basvodde:
I would be interested in the ability to see stack traces on memory leaks and probably optional on mock failures. On test failures, it is of no use :)

bryceschober:
My interest is in a pragmatic opt-in solution for unit tests covering
sources that use placement new.

PR #1146, opened 2018-03-19
Capture function info from call stack for memory leak detection.

jgonzalezdr:
When the new operator override macros are not used (e.g. to avoid STL
errors), the function that called the new operator is traced by
inspecting the call stack.

Currently only works for GCC when the backtrace function is supported
(i.e. no support for MinGW).

### DIFF
--- a/include/CppUTest/MemoryLeakDetector.h
+++ b/include/CppUTest/MemoryLeakDetector.h
@@ -117,18 +117,18 @@ private:
 struct MemoryLeakDetectorNode
 {
     MemoryLeakDetectorNode() :
-        size_(0), number_(0), memory_(NULLPTR), file_(NULLPTR), line_(0), addr_(NULLPTR), allocator_(NULLPTR), period_(mem_leak_period_enabled), next_(NULLPTR)
+        size_(0), number_(0), memory_(NULLPTR), file_(NULLPTR), line_(0), caller_(NULLPTR), allocator_(NULLPTR), period_(mem_leak_period_enabled), next_(NULLPTR)
     {
     }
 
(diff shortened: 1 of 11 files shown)


jgonzalezdr:
This is a preliminary PR just to discuss this feature, as started in #34. The next step I'll do is to call addr2line to obtain file and line info when printing report.

Afterwards, many different improvement can be done: Add support for Win32 equivalent functionality, maybe clang support also, partially import addr2line functionality to avoid executing external calls (and break dependency with addr2line being installed / accessible), etc.

Also, not related to mem leaks, I'll also explore adding callstack printing when a failure is raised.

basvodde:
Thanks for the initial PR. This will definitively not be integrated as-is. It'll require a lot of work still.

As a starting point though. Could you make a pull request for the rfind in SimpleString. But please rename it properly, fix the implementation to be more in the style of the rest of CppUTest, and add proper test to it. It is a good chance for practices TDD as it is such an isolated piece of code.

basvodde:
Also, the eventual implementation will need to be so that there are much much much less #if in the code.

jgonzalezdr:
@basvodde This is just a proof of concept, I wanted to keep it simple so you an others can test it a bit, see if it is useful, see quickly what needs to be changed, etc. A final version will definitely need some "preparation" PRs to make some structural changes, will need to add some optimization to keep the memory footprint of memory allocation info nodes the same than now, etc., but will of course be a bit more complex (even if tidier) and a bit harder to review.

Regarding the rfind functions, I've just cherry picked that quickly from a dangling branch I've got here for another feature I'm playing with, it's got its UT, so I can push that quickly. However, what do you mean by "fix the implementation to be more in the style of the rest of CppUTest"? Just code style?

basvodde:
Ok. I'll ignore this for now then :)

PR #1149, opened 2018-03-22
cpputest: kill example test with memset

jbrandeb:
This breaks badly if building with
```
./autogen.sh
./configure --prefix=/home/jesse/cpputest
make
make install
cd examples
make CPPUTEST_HOME=/home/jesse/cpputest
```

### DIFF
--- a/examples/AllTests/HelloTest.cpp
+++ b/examples/AllTests/HelloTest.cpp
@@ -27,6 +27,7 @@
 
 #include "hello.h"
 
+#include <cstring>
 #include <stdio.h>
 #include <stdarg.h>
 #include "CppUTest/TestHarness.h"
@@ -49,6 +50,7 @@ void setup()
 }
 void teardown()
 {
+    memset(buffer, 0, sizeof(*buffer));
     delete buffer;
 }
 };
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -11,7 +11,8 @@ endif
 
 #--- Inputs ----#
 COMPONENT_NAME = CppUTestExamples
-CPPUTEST_HOME = ..
+CPPUTEST_BUILD = ..
+CPPUTEST_HOME := ..
 
 CPPUTEST_USE_EXTENSIONS = Y
 CPP_PLATFORM = Gcc
@@ -31,6 +32,6 @@ INCLUDE_DIRS =\
   ApplicationLib\
   $(CPPUTEST_HOME)/include\
 
-include $(CPPUTEST_HOME)/build/MakefileWorker.mk
+include $(CPPUTEST_BUILD)/build/MakefileWorker.mk
 
 


examples/AllTests/HelloTest.cpp
```diff
@@ -49,6 +50,7 @@ void setup()
 }
 void teardown()
 {
+    memset(buffer, 0, sizeof(*buffer));
```
basvodde: Why the additional memset?

jbrandeb:
It was just there because I needed something from string.h to prove my
point.  It isn't actually doing anything useful, probably should have put
in a comment.  The real root cause is that make install doesn't result in
working directories when you have a file that includes string.h.

basvodde:
Ok. So, this shouldn't be merged?

jbrandeb:
This is an example failing test. Not really sure how to fix, but yes I would not merge until you or someone else can get this test to start passing. 

This was introduced with a commit in the last couple of years but apparently no one uses or tests make install. 

--
Jesse Brandeburg

basvodde:
I do make install all the time and it is done in most distributions, so I'm a bit surprised that it would be broken. Lemme quickly check

basvodde:
Tried your example. The make install works fine. The make in the examples fails, but it fails because the MakefileWorker.mk is not installed.

jbrandeb:
I know you said you tried my example.  What branch did you try it on?  Did you do the exact steps I provided above?  I'm going to submit this comment and then go try a fresh pull.

Here is a more specific explanation from one of my coworkers:

Using the LeakDetector header files from CppUTest which has been installed on the system (i.e. CPPUTEST_HOME=/usr) is problematic for applications which include <string.h>.
Using the LeakDetector header files from the CppUTest source directory (i.e. CPPUTEST_HOME=~/git/cpputest) works fine.

In the “installed” case, MemoryLeakDetectorMallocMacros.h includes CppUTestConfig.h, which includes CppUTestGeneratedConfig.h, which has the auto-generated info from “configure”.
In the “source” case, the same thing happens as above, except CppUTestGeneratedConfig.h is not the “real” file, it includes another “generated/CppUTestGeneratedConfig.h”, only if HAVE_CONFIG_H is defined.

Since the tdd_tests do not define “HAVE_CONFIG_H”, the GeneratedConfig.h file is never included.
The result is that “CPPUTEST_HAVE_STRDUP” is not defined, and so the cpputest_strdup_location() functions are never declared.

jbrandeb:
Ok, I get the same MakefileWorker.mk fault now.  So at least I see that issue. Is the problem actually because our Makefile is *not* including MakefileWorker.mk?

jbrandeb:
Ah, you have to apply my patch to the examples directory on your fresh pull, before you will see a failure because the examples directory files don't by default include string.h which is what triggers the problem.

My patch also works around the MakefileWorker.mk problem, so that should have been a giveaway to me that you didn't test with my patch.  Sorry I missed that during the discussion.

jbrandeb:
Not sure if you saw the attached issue, the result of compiling the examples directory after you clone with this patch applied is:
[jbrandeb@jfsjbrandeb002 examples]$ make CPPUTEST_HOME=/home/jbrandeb/cpputest2
compiling HelloTest.cpp
/usr/include/string.h:166:37: error: expected identifier before string constant
 extern char *strdup (const char *__s)
                                     ^
/usr/include/string.h:166:37: error: expected ‘,’ or ‘...’ before string constant
In file included from <command-line>:0:0:
/home/jbrandeb/cpputest2/include/CppUTest/MemoryLeakDetectorMallocMacros.h:62:21: error: conflicting declaration of C function ‘char* cpputest_strdup_location(const char*, int)’
 #define strdup(str) cpputest_strdup_location(str, __FILE__, __LINE__)
                     ^
/home/jbrandeb/cpputest2/include/CppUTest/MemoryLeakDetectorMallocMacros.h:55:14: note: previous declaration ‘char* cpputest_strdup_location(const char*, const char*, int)’
 extern char* cpputest_strdup_location(const char* str, const char* file, int line);
              ^~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/string.h:174:55: error: expected identifier before string constant
 extern char *strndup (const char *__string, size_t __n)
                                                       ^
/usr/include/string.h:174:55: error: expected ‘,’ or ‘...’ before string constant
/home/jbrandeb/cpputest2/include/CppUTest/MemoryLeakDetectorMallocMacros.h:63:25: error: conflicting declaration of C function ‘char* cpputest_strndup_location(const char*, size_t, int)’
 #define strndup(str, n) cpputest_strndup_location(str, n, __FILE__, __LINE__)
                         ^
/home/jbrandeb/cpputest2/include/CppUTest/MemoryLeakDetectorMallocMacros.h:56:14: note: previous declaration ‘char* cpputest_strndup_location(const char*, size_t, const char*, int)’
 extern char* cpputest_strndup_location(const char* str, size_t n, const char* file, int line);
              ^~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [../build/MakefileWorker.mk:504: objs/AllTests/HelloTest.o] Error 1

jbrandeb:
Closing this example fail because the bug was fixed by #1192

PR #1150, opened 2018-03-22
CMakeLists: use standard install directory variables

bkylerussell:
Use the variables provided by cmake for install directories to
standardize across build systems.

The hard-coded /lib appended to exec_prefix does not necessarily match
automake's $(libdir) (especially in multilib scenarios), so depending on
whether you're using cmake or automake, you could actually end up with
different pkg-config files given the right configuration, in addition to
files being installed in different directories.


basvodde:
Thanks

PR #1154, opened 2018-04-06
IAR: Add support to redirect file writes to stdout.

CFlix:
Ad the possibility to output strings that are normally written to a file to stdout. This makes it possible to generate JUnit XML output with the IAR simulator or even real HW. All FPuts are redirected to stdout and prepended with "FILEx" where x is an incrementing number. This makes it possible to separate the output later to different files.
Small caveat: All FPuts must be terminated with a newline character and should not contain newline characters except for the terminating one.

### DIFF
--- a/src/Platforms/Iar/UtestPlatform.cpp
+++ b/src/Platforms/Iar/UtestPlatform.cpp
@@ -125,15 +125,18 @@ int (*PlatformSpecificVSNprintf)(char *str, size_t size, const char* format, va_
 
 static PlatformSpecificFile PlatformSpecificFOpenImplementation(const char* filename, const char* flag)
 {
+    static int fileNo = 0;   // Count the files so different files can be differenciated.
     (void)filename;
     (void)flag;
-    return 0;
+    fileNo++;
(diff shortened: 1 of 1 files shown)


basvodde:
I'm not sure about this commit and would like to have some other people look at it perhaps @arstrube 

Also, the code contains a significant amount of comments. Could you look at making it in such a way that these comments are not needed?

CFlix:
Yes I can do that. The intention was to clarify the behavior because it is different than the standard file function behavior. Is there a different place where to document that? Maybe a comment above the function or no comment at all what do you prefer?

CFlix:
Regarding the functionality of the commit itself: I can provide a powershell script that splits the output to different files that can be further processed (in my case the Jenkins JUnit report plugin). If I find the time I will provide a merge request with an example in the `examples/` directory.

basvodde:
Related to comment. You can probably name your variable to reflect this?

Powershell script: ack!

ChunBoo:
Can try to enable this option in IAR
image

Issue #1158, opened 2018-05-23
SimpleString ambiguous calls with NULL


jacob-keller:
When using a newer C++ compiler, we're seeing an issue with using CHECK_EQUAL and a NULL pointer value.

Specifically, I have a check like this:

CHECK_EQUAL(NULL, actual);

When we compile using clang 4, we have no issues, but when updating to clang 6, we see the following compilation issue:

```C++
test.cpp:89:2: error: call to 'StringFrom' is ambiguous
        CHECK_EQUAL(NULL, actual);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/CppUTest/UtestMacros.h:127:3: note: expanded from macro 'CHECK_EQUAL'
  CHECK_EQUAL_LOCATION(expected, actual, NULL, __FILE__, __LINE__)
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/CppUTest/UtestMacros.h:138:52: note: expanded from macro 'CHECK_EQUAL_LOCATION'
      UtestShell::getCurrent()->assertEquals(true, StringFrom(expected).asCharString(), StringFrom(actual).asCharString(), text, file, line); \
                                                   ^~~~~~~~~~
/usr/local/include/CppUTest/SimpleString.h:138:14: note: candidate function
SimpleString StringFrom(const void* value);
             ^
/usr/local/include/CppUTest/SimpleString.h:139:14: note: candidate function
SimpleString StringFrom(void (*value)());
             ^
/usr/local/include/CppUTest/SimpleString.h:141:14: note: candidate function
SimpleString StringFrom(const char *value);
             ^
/usr/local/include/CppUTest/SimpleString.h:137:14: note: candidate function
SimpleString StringFrom(bool value);
             ^
/usr/local/include/CppUTest/SimpleString.h:157:14: note: candidate function
SimpleString StringFrom(const SimpleString& other);
             ^
/usr/local/include/CppUTest/SimpleString.h:171:14: note: candidate function
SimpleString StringFrom(const std::string& other);
```

This appears, I believe, because of how StringFrom in the SimpleString.h header doesn't have a strict equivalent StringFrom(nullptr_t), so it's ambiguous which function it could actually be used with.

I believe when building using clang4, that NULL is not considered a nullptr type but just zero.

[...]

PR #1159, opened 2018-05-29
add StringFrom for std::nullptr_t values

jacob-keller:
In C++11, the nullptr_t type was introduced, and in some projects NULL
is defined as nullptr. In this case, the C++ compiler will produce
warnings when trying to convert NULL to a string, because it is
ambiguous which conversion is used.

  test.cpp:89:2: error: call to 'StringFrom' is ambiguous
        CHECK_EQUAL(NULL, actual);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /usr/local/include/CppUTest/UtestMacros.h:127:3: note: expanded from macro 'CHECK_EQUAL'
    CHECK_EQUAL_LOCATION(expected, actual, NULL, __FILE__, __LINE__)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /usr/local/include/CppUTest/UtestMacros.h:138:52: note: expanded from macro 'CHECK_EQUAL_LOCATION'
        UtestShell::getCurrent()->assertEquals(true, StringFrom(expected).asCharString(), StringFrom(actual).asCharString(), text, file, line); \
                                                     ^~~~~~~~~~
  /usr/local/include/CppUTest/SimpleString.h:138:14: note: candidate function
  SimpleString StringFrom(const void* value);
               ^
  /usr/local/include/CppUTest/SimpleString.h:139:14: note: candidate function
  SimpleString StringFrom(void (*value)());
               ^
  /usr/local/include/CppUTest/SimpleString.h:141:14: note: candidate function
  SimpleString StringFrom(const char *value);
               ^
  /usr/local/include/CppUTest/SimpleString.h:137:14: note: candidate function
  SimpleString StringFrom(bool value);
               ^
  /usr/local/include/CppUTest/SimpleString.h:157:14: note: candidate function
  SimpleString StringFrom(const SimpleString& other);
               ^
  /usr/local/include/CppUTest/SimpleString.h:171:14: note: candidate function
  SimpleString StringFrom(const std::string& other);

These warnings can be resolved by providing a StringFrom which takes the
nullptr_t as the argument, avoiding the need for conversions at all.

According to the C++ standard, the define __cplusplus is set to a value greater
than 199711L when C++ 11 is supported. Thus, we'll check this when implementing
the function, in order to allow building on older C++ standards.

Additionally, avoid a compiler warning due to the expected unused parameter in
the implementation of StringFrom.

### DIFF
--- a/include/CppUTest/SimpleString.h
+++ b/include/CppUTest/SimpleString.h
@@ -174,7 +174,9 @@ SimpleString BracketsFormattedHexStringFrom(cpputest_ulonglong value);
 SimpleString BracketsFormattedHexStringFrom(signed char value);
 SimpleString BracketsFormattedHexString(SimpleString hexString);
 
-
+#if __cplusplus > 199711L
+SimpleString StringFrom(const std::nullptr_t value);
+#endif
 
 #if CPPUTEST_USE_STD_CPP_LIB
 
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -540,6 +540,12 @@ SimpleString BracketsFormattedHexString(SimpleString hexString)
     return SimpleString("(0x") + hexString + ")" ;
 }
 
+#if __cplusplus > 199711L
+SimpleString StringFrom(const std::nullptr_t __attribute__((unused)) value)
+{
+    return "(null)";
+}
+#endif
 
 #ifdef CPPUTEST_USE_LONG_LONG
 
--- a/tests/CppUTest/SimpleStringTest.cpp
+++ b/tests/CppUTest/SimpleStringTest.cpp
@@ -577,8 +577,6 @@ TEST(SimpleString, nullptr_type)
 {
     SimpleString s(StringFrom(nullptr));
     STRCMP_EQUAL("(null)", s.asCharString());
-
-    CHECK_EQUAL(nullptr, (nullptr_t)0);
 }
 #endif
 
--- a/tests/CppUTest/SimpleStringTest.cpp
+++ b/tests/CppUTest/SimpleStringTest.cpp
@@ -572,6 +572,16 @@ TEST(SimpleString, Sizes)
     STRCMP_EQUAL("10", StringFrom((int) size).asCharString());
 }
 
+#if __cplusplus > 199711L
+TEST(SimpleString, nullptr_type)
+{
+    SimpleString s(StringFrom(nullptr));
+    STRCMP_EQUAL("(null)", s.asCharString());
+
+    CHECK_EQUAL(nullptr, (nullptr_t)0);
+}
+#endif
+
 TEST(SimpleString, HexStrings)
 {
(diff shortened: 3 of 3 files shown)


jacob-keller:
This should fix https://github.com/cpputest/cpputest/issues/1158

jacob-keller:
The appveyor tests appear to be failing due to MinGW installations missing?? I don't think that's related to this patch.

jacob-keller:
```
/home/travis/build/cpputest/cpputest/tests/CppUTest/SimpleStringTest.cpp:581:34: error: zero as null pointer constant [-Werror,-Wzero-as-null-pointer-constant]
    CHECK_EQUAL(nullptr, (void *)0);
                                 ^
                                 nullptr
/home/travis/build/cpputest/cpputest/include/CppUTest/UtestMacros.h:127:34: note: expanded from macro 'CHECK_EQUAL'
  CHECK_EQUAL_LOCATION(expected, actual, NULLPTR, __FILE__, __LINE__)
                                 ^
```

I was trying to see if nullptr and 0 would compare properly in the new test case, but it turns out this causes warnings. Any suggestions?

basvodde:
Perhaps cast it to nullptr_t ?

jacob-keller:
> Perhaps cast it to nullptr_t 

Maybe. I was trying to avoid that to see whether we actually treat nullptr_t and 0 as the same when checking null pointers..

jacob-keller:
Ok, this should be complete. I've got checks for the C++ version, and a test case to ensure that stringfrom works, and that CHECK_EQUAL works.

basvodde:
The travis build is still failing though ?

jacob-keller:
Ok, so it doesn't like type casting 0 to a nullptr...

jacob-keller:
`
C:\cygwin\bin\bash.exe : configure.ac:11: installing './compile'
At line:1 char:1
+ C:\cygwin\bin\bash.exe --login -c 'cd /cygdrive/c/projects/cpputest/c ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (configure.ac:11: installing './compile':String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
configure.ac:11: installing './config.guess'
configure.ac:11: installing './config.sub'
configure.ac:4: installing './install-sh'
configure.ac:4: installing './missing'
Makefile.am: installing './INSTALL'
Makefile.am: installing './depcomp'
parallel-tests: installing './test-driver'
Executing <../configure> in </cygdrive/c/projects/cpputest/cpputest_build>
checking for a BSD-compatible install... /usr/b
`

Not sure what the problem is here?

jacob-keller:
I don't understand the failure on cygwin...

basvodde:
Seems to be caused by network. I restarted, now it is good.

PR #1164, opened 2018-07-01
Fix CppUTest for OSX / Xcode9 / clang-900.x or any newer version

SaeedYasin:
Add -Wno-reserved-id-macro and -Wno-keyword-macro to clang builds. Fixes CppUTest on Xcode 7+, with clang-700+, where warnings were added about using reserved namespaces like __, and for macros that shadow keywords. Also, make it future proof.

Remove warning, cast from 'const void *' to 'Event *' drops const qualifier in examples/AllTests/EventDispatcherTest.cpp file.


basvodde:
Doubly solved by renaming the macro.

PR #1166, opened 2018-07-13
Print test names and test group names escaped when using teamcity

sensslen:
When using Teamcity output, we encountered ill formatted output due to missing escapes in testnames. This Patch always outputs "user defined messages" escaped


basvodde:
Would you mind adding a test for it?

sensslen:
A valid request - added Tests for escaped prining of Test Group and Test name

PR #1172, opened 2018-09-05
TestDoubles (unopionated CppUMock)

redengin:
see README_TestDoubles.md.

This implementation wraps the current CppUMock behavior using a more expressive grammar (tail recursion by type).

The CppUTestExt/Test_TestDouble.cpp demonstrates that this grammar can be used for call matching, parameter matching, output matching, and return values.

The README_TestDoubles.md identifies where the TestDouble design doesn't match the semantics of CppUMock.  I'd like to coordinate on how to go forward (i.e. make CppUMock less opinionated, or fork into a separate semantic).

FIXMEs:
* TestDouble classes expect C++ move semantics.  Instead, explicit move semantics can be used to reduce the copying of SimpleString.
* Expectations should be expressed as a set.  The CppUMock semantics enforce sequence of expectations when matching the actual (based on the assumption of matching all actual).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,9 +88,18 @@ CppUTestTests
 Makefile
 CMakeFiles/
 CMakeCache.txt
-*.cmake
 Testing/
 generated/
+CTestTestfile.cmake
+install/
+CppUTestConfig.cmake
+CppUTestConfigVersion.cmake
+CppUTestTargets.cmake
+cmake_install.cmake
+src/CppUTest/cmake_install.cmake
+src/CppUTestExt/cmake_install.cmake
+tests/CppUTest/cmake_install.cmake
+tests/CppUTestExt/cmake_install.cmake
 
 # Dart files
 DartConfiguration.tcl
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ CMakeCache.txt
 *.cmake
 Testing/
 generated/
+
+# Dart files
+DartConfiguration.tcl
--- a/.gitignore
+++ b/.gitignore
@@ -51,9 +51,6 @@ _build_
 _build
 builds/*gcc*
 /.deps
-CppUTestExtTests
-CppUTestTests
-Makefile
 config.h
 config.h.in
 config.status
@@ -86,6 +83,9 @@ cpputest_*.xml
 *.ewt
 
 #CMake files
+CppUTestExtTests
+CppUTestTests
+Makefile
 CMakeFiles/
 CMakeCache.txt
 *.cmake
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,7 @@ builds/*gcc*
 /.deps
 CppUTestExtTests
 CppUTestTests
-/Makefile
+Makefile
 config.h
 config.h.in
 config.status
@@ -84,3 +84,10 @@ cpputest_*.xml
 #IAR automatically generated files
 *.dep
 *.ewt
+
+#CMake files
+CMakeFiles/
+CMakeCache.txt
+*.cmake
+Testing/
+generated/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "plantuml.server": "http://www.plantuml.com/plantuml",
+  "plantuml.render": "PlantUMLServer",
+}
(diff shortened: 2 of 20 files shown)


.gitignore
```diff
@@ -84,3 +81,16 @@ cpputest_*.xml
...
 *.dep
 *.ewt
+
+#CMake files
+CppUTestExtTests
+CppUTestTests
+Makefile
+CMakeFiles/
+CMakeCache.txt
+*.cmake
```
offa: `.cmake` are used within this project.

basvodde:
Thanks for the pull request. This will need quite a lot of work before I will integrate this within the code. However, I read the tests and I couldn't really understand what you are trying to achieve over the current CppUMock?

redengin:
@basvodde the current implementation just wraps CppUMock so the only value added is a more expressive grammar.  The TestDouble grammar separates expectations from actual calls to eliminate the common syntax error I've seen where developers mistakenly use mock().expectOneCall() inside the Mock (instead of mock().actualCall()).

My objective of this pull request is to make you aware of the TestDouble design (see README_TestDoubles.md).  This pull request demonstrates that the TestDouble grammar will behave equivalently to CppUMock.

To realize the TestDouble design, I'd like to coordinate with you to determine whether CppUMock should be the underlying implementation for TestDouble or if TestDouble should implement its own expectation framework based on the CppUMock code base.  The primary difference is that the TestDouble framework only opines on the cardinality of expectations if the expectation explicitly describes .times() (otherwise the expectation services all actual calls).  The other change is to allow an expectation to inject a function (.use(<model>)) to handle behavior rather than require the mocked method to implement model behavior.

basvodde:
Ok. Not quite sure why it is more expressive than the current Mock interface. Implementation wise, it seems to use templates a lot, which are avoided within CppUTest. That will likely make it hard to integrate and will definitively prevent it from being a standard interface.

.gitignore
redengin: I fixed the .gitignore to only ignore the generated files.

redengin:
@coveralls @basvodde  @jwgrenning 
is there a way to run CI checks locally (so I don't have to push commits to test)?

basvodde:
The ci runs mostly make test_all

redengin:
@basvodde I'm interested in why templates are avoided.  If necessary I'm happy to explicitly define each invocation based on your guidance - (i.e. I'm guessing you don't want a macro function that implements template functionality).

redengin:
@basvodde 
The CI contexts I want to run locally are:
* continuous-integration/travis-ci/pr
* continuous-integration/appveyor/pr
* coverage/coveralls

redengin:
To get continuous-integration/travis-ci/pr I'm not sure how to initialize a pointer.  Clang doesn't like 0 as a reference to nullptr, but nullptr is only provided by C++11 and therefore can't be used per CppUTest compile options.

using a static_cast<void*>(0) to fake nullptr

basvodde:
re templates. We avoid them as we want to be as safe as possible for old and embedded compilers. Hence, templates (with a few exceptions which are put under #ifdef) are not used.

basvodde:
Don't think you can run the CI local. Sorry about that.

redengin:
@basvodde  I've finally passed CI, what next? (this is only a prototype, it's not ready to merge to master)

The current implementation of TestDouble is a prototype to demonstrate that the grammar of using TestDouble is largely equivalent to CppUMock grammar.  How can we share ideas about the grammar (i.e. call semantics)?

As there is already a large body of users of CppUMock, I don't expect that TestDouble would replace CppUMock, but rather the two can leverage a common Expectation Framework (e.g. a generalized store of expectations which both implementations can query).

You've identified that their are additional coding standards (e.g. not using C++ templates).  I'd like to pick your brain on coding standards and document them for others.

redengin:
README_TestDoubles.md lays out how CppUTest can support familiar API's (mockito, gtest/gmock).  I'm fine with dropping this pull request.  My next pull request will be replacing the common functionality between CppUMock and TestDoubles with a new base class.

redengin:
working on new implementation

basvodde:
Ok. I'm still not sure what you are trying to achieve and am not very likely to merge it, but I'm curious and open to be convinced differently.

redengin:
The design is described in README_TestDoubles.md.  Let me know what you think.

PR #1176, opened 2018-09-11
Add _TEXT assertions to C-Interface

youduda:
Adds for all C-Interface assertions a <assertion>_TEXT macro variant using the existing functionality.


basvodde:
Thanks.

PR #1177, opened 2018-09-18
add OrderedTest C-Interface

youduda:
Adds C-Interface macro variant for OrderedTests
Adds tests for OrderedTests with the same level and C-Interface


basvodde:
Thanks!

PR #1181, opened 2018-11-05
Shuffle test execution order with -s

rivuonna:
Shuffling is off by default.

-s enables shuffling with random seed derived from current time.

The seed is printed before executing tests. A specific test execution
order can be reproduced by setting the seed explicitly with -s <seed>.

Seed values 0 and 1 are reserved values and cannot be set.

If -v (verbose) is enabled, the seed is also printed after test
execution. This might be useful in case tests produce console output
which could cause the seed printed at the beginning to be lost.

Test order is reshuffled between repeats (-r).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,10 @@ missing
 test-driver
 platforms/iar/settings
 cpputest_*.xml
+generated/
+*.deps/
+*.dirstamp
+*.trs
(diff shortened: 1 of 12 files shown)


rivuonna:
This pull request is obsolete. New PR: #1182

Issue #865, opened 2016-02-12
Allow running tests in a shuffled/randomized order


WilcoBT:
We previously used gtest and then the --gtest_repeat=[XXX] with  --gtest_shuffle to run the tests a few times in different orders to exclude ordering issues. I do see the repeat parameter on the website, but not the shuffle.
Are there any plans to add support for this function or estimations how to add it and how much trouble this would be?

basvodde:
No, because if you run random and get random failures, then you'll need to know the random pattern in order to debug it. If it is purely random, it is only leading to guarenteed frustration :P

arstrube:
Not necessarily. If you run the tests with -v, it is fairly easy to narrow down a problem (if it does occur, which most of the time it shouldn't).

Andne:
The catch there is that if the tests are random, multiple runs with -v may not actually trigger the failure, especially if it's only when two specific tests run one after the other.  At that point, debugging becomes quite complicated since you have no guarantee that the error will occur on any given run.

Here's a thought - what if as part of running the tests on random, you could output to the test order to a file and then read in that file in order to execute the tests in the same order again?

arstrube:
It can already be done:
```
whatevertest.exe -v >log.txt
```
The output (preferably a limited subset) can be fed into a debug run, if needed:
```
whatevertest.exe TEST(group, test5) TEST(group, test1) .....
```
An entirely different question is how much work the "randomness" would be to implement....

WilcoBT:
Well, about the reproducability, I got the impression from the google variant that the randomness was based on a seed that was put in the log file(I at least see it on my console), so if you have an issue you could rerun with that seed to reproduce.

mcergun:
any progress on this aspect?

WilcoBT:
Not yet, we first have a release to finish.

But we are wondering if our fix will be meeting the requirements for backporting here or will be added to the changes that we have on the local checkout.

rivuonna:
Check out PR #1181

rivuonna:
Check out #1182

PR #1182, opened 2018-11-05
Shuffle test execution order with -s

rivuonna:
Shuffling is off by default.

-s enables shuffling with random seed derived from current time.

The seed is printed before executing tests. A specific test execution
order can be reproduced by setting the seed explicitly with -s <seed>.

Seed values 0 and 1 are reserved values and cannot be set.

If -v (verbose) is enabled, the seed is also printed after test
execution. This might be useful in case tests produce console output
which could cause the seed printed at the beginning to be lost.

Test order is reshuffled between repeats (-r).

### DIFF
--- a/include/CppUTest/CommandLineArguments.h
+++ b/include/CppUTest/CommandLineArguments.h
@@ -81,7 +81,7 @@ private:
 
     SimpleString getParameterField(int ac, const char *const *av, int& i, const SimpleString& parameterName);
     void SetRepeatCount(int ac, const char *const *av, int& index);
-    void SetShuffle(int ac, const char *const *av, int& index);
+    bool SetShuffle(int ac, const char *const *av, int& index);
     void AddGroupFilter(int ac, const char *const *av, int& index);
     void AddStrictGroupFilter(int ac, const char *const *av, int& index);
     void AddExcludeGroupFilter(int ac, const char *const *av, int& index);
--- a/include/CppUTest/CommandLineArguments.h
+++ b/include/CppUTest/CommandLineArguments.h
@@ -47,6 +47,7 @@ public:
     bool isListingTestGroupAndCaseNames() const;
     bool isRunIgnored() const;
     int getRepeatCount() const;
+    unsigned int getShuffle() const;
     const TestFilter* getGroupFilters() const;
     const TestFilter* getNameFilters() const;
     bool isJUnitOutput() const;
@@ -72,6 +73,7 @@ private:
     bool listTestGroupAndCaseNames_;
     bool runIgnored_;
     int repeat_;
+    unsigned int shuffle_; // 0: shuffling disabled, 1: random seed, other values: specific random seed
     TestFilter* groupFilters_;
     TestFilter* nameFilters_;
     OutputType outputType_;
@@ -79,6 +81,7 @@ private:
 
     SimpleString getParameterField(int ac, const char *const *av, int& i, const SimpleString& parameterName);
     void SetRepeatCount(int ac, const char *const *av, int& index);
+    void SetShuffle(int ac, const char *const *av, int& index);
     void AddGroupFilter(int ac, const char *const *av, int& index);
     void AddStrictGroupFilter(int ac, const char *const *av, int& index);
     void AddExcludeGroupFilter(int ac, const char *const *av, int& index);
--- a/include/CppUTest/CommandLineArguments.h
+++ b/include/CppUTest/CommandLineArguments.h
@@ -73,7 +73,7 @@ private:
     bool listTestGroupAndCaseNames_;
     bool runIgnored_;
     int repeat_;
-    unsigned int shuffle_; // 0: shuffling disabled, 1: random seed, other values: specific random seed
+    unsigned int shuffle_;
     TestFilter* groupFilters_;
     TestFilter* nameFilters_;
     OutputType outputType_;
--- a/include/CppUTest/SimpleString.h
+++ b/include/CppUTest/SimpleString.h
@@ -93,7 +93,6 @@ public:
     static void setStringAllocator(TestMemoryAllocator* allocator);
 
     static int AtoI(const char*str);
-    static unsigned AtoU(const char*str, bool& wasAbleToParseADigit);
     static unsigned AtoU(const char*str);
     static int StrCmp(const char* s1, const char* s2);
     static size_t StrLen(const char*);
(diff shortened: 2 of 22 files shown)


include/CppUTest/CommandLineArguments.h
```diff
@@ -72,13 +73,15 @@ class CommandLineArguments
     bool listTestGroupAndCaseNames_;
     bool runIgnored_;
     int repeat_;
+    unsigned int shuffle_; // 0: shuffling disabled, 1: random seed, other values: specific random seed
```
basvodde: Can you use enum values of #defines for these rather than adding a comment?

include/CppUTest/StandardCLibrary.h
```diff
@@ -15,6 +15,8 @@
 #ifdef __cplusplus
  #if CPPUTEST_USE_STD_CPP_LIB
   #include <cstdlib>
+  #include <vector>
+  #include <algorithm>
```
basvodde: We don't really want to depend that much on the STD C++ library. Is there a way around that?

src/CppUTest/CommandLineArguments.cpp
```diff
@@ -149,6 +155,21 @@ void CommandLineArguments::SetRepeatCount(int ac, const char *const *av, int& i)
 
 }
 
+void CommandLineArguments::SetShuffle(int ac, const char * const *av, int& i)
+{
+    shuffle_ = 1;
```
basvodde: Is this line needed?

tests/CppUTest/CommandLineArgumentsTest.cpp
```diff
@@ -102,12 +111,74 @@ TEST(CommandLineArguments, repeatSetDifferentParameter)
...
+    int argc = 3;
+    const char* argv[] = { "tests.exe", "-s", "777"};
+    CHECK(newArgumentParser(argc, argv));
+    CHECK_EQUAL(args->getShuffle(), 777);
+}
+
+TEST(CommandLineArguments, shuffleBeforeDoesNotDisturbOtherSwitchCase1)
+{
+    const char* argv[] = { "tests.exe", "-s", "-sg", "group" };
+    shuffleDisturbTest(4, argv, 1);
```
basvodde: These tests don't test anything...

basvodde:
The code will need to be cleaned up quite a bit. But the most worrying part of the dependency to std C++ vector, which isn't what we want to have in the CppuTest core.

rivuonna:
FYI, I will look into resolving the review comments later when I have free time.

include/CppUTest/StandardCLibrary.h
rivuonna: Fixed using defines

include/CppUTest/CommandLineArguments.h
rivuonna: Fixed using defines

include/CppUTest/StandardCLibrary.h
rivuonna: Fixed

src/CppUTest/CommandLineArguments.cpp
rivuonna: Yes - it's the code path that enables shuffling when command line switch includes plain "-s" (unit test: CommandLineArguments :: shuffleEnabled )

tests/CppUTest/CommandLineArgumentsTest.cpp
rivuonna: In fact they do, although I trimmed the number of tests. `CommandLineArguments::parse` parses command line switches in order - if one parser misbehaves, it can break the following command line switches.

E.g., comment out `if (wasAbleToParseDigit)` in `CommandLineArguments::SetShuffle` --> switches following `-s` will not work, and `shuffleBeforeDoesNotDisturbOtherSwitch` unit test will catch.

rivuonna:
> 
> 
> The code will need to be cleaned up quite a bit. But the most worrying part of the dependency to std C++ vector, which isn't what we want to have in the CppuTest core.

@basvodde I finally took the time to address the review comments and fix the issues. Would you please re-review?

include/CppUTest/SimpleString.h
```diff
@@ -93,6 +93,8 @@ class SimpleString
     static void setStringAllocator(TestMemoryAllocator* allocator);
 
     static int AtoI(const char*str);
+    static unsigned AtoU(const char*str, bool& wasAbleToParseADigit);
```
basvodde: Would you mind returning 0 on failure? I believe that works and then the output boolean is not needed. I know it works like this, just wouldn't really want to have output boolean parameters :)

basvodde:
Thanks. I re-reviewed it. It is an interesting implementation. I'll merge it, just would like to get rid of the output boolean. If you don't feel strongly about it, please remove it. If you do feel strongly about it then I'll leave it as is and merge it.

Thanks!

include/CppUTest/SimpleString.h
rivuonna: If 0 value is used to signal error, `shuffleDisabledCase3` stops working (i.e., `./testExecutable -s 0` does not run any tests). --> I would say the output boolean is mandatory, and it needs to be kept.

basvodde: But... 0 is an invalid seed in your code anyways... so -s 0 would be an invalid command already now.

rivuonna: I changed the behavior of the command line parser to treat `-s0` (and `-s 0`) as invalid parameters, and was able to remove the output boolean.

src/CppUTest/CommandLineArguments.cpp
```diff
@@ -149,6 +157,26 @@ void CommandLineArguments::SetRepeatCount(int ac, const char *const *av, int& i)
...
+    SimpleString shuffleParameter(av[i]);
+    if (shuffleParameter.size() > 2) {
+        shuffle_ = SimpleString::AtoU(av[i] + 2);
+        if (shuffle_ == SHUFFLE_DISABLED) {
+            return false;
+        }
+    } else if (i + 1 < ac) {
+        const unsigned parsed = SimpleString::AtoU(av[i + 1]);
+        if (parsed != 0)
+        {
```
basvodde: I'll merge it, but would recommend to try to keep your style consistent :)

basvodde:
Thanks for the contribution!

src/CppUTest/CommandLineArguments.cpp
rivuonna: Oops, missed that one. Thanks for merging!

rivuonna:
By the way, you can now resolve issue #865

PR #1184, opened 2018-11-14
Do not mark cpp source files as executable

rnestler:
These cpp source files were marked as executable for whatever reason.
Since there is no point in marking source code executable we can remove
the flag.


basvodde:
Thanks!

PR #1189, opened 2018-12-19
Added support for setting tolerance for doubles in mock

oyvindlr:
Added overloaded methods for withParameter and withDoubleParameter which takes a tolerance as well as the expected value for doubles

### DIFF
--- a/tests/CppUTestExt/MockNamedValueTest.cpp
+++ b/tests/CppUTestExt/MockNamedValueTest.cpp
@@ -27,7 +27,6 @@
 
 #include "CppUTest/TestHarness.h"
 #include "CppUTestExt/MockNamedValue.h"
-#include <cfloat>
 
 TEST_GROUP(ComparatorsAndCopiersRepository)
 {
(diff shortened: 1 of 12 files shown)


tests/CppUTestExt/MockNamedValueTest.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include "CppUTest/TestHarness.h"
 #include "CppUTestExt/MockNamedValue.h"
+#include <cfloat>
```
basvodde: Is it possible to get rid of the cfloat include here? We try to avoid direct Standard C includes to increase portability.

basvodde:
Thanks. Looks good. Just one minor issue that could be improved.

tests/CppUTestExt/MockNamedValueTest.cpp
oyvindlr: Absolutely. It's not used, I just forgot to remove it. Will fix once I'm back from holiday.

basvodde:
Thanks.

PR #1190, opened 2018-12-19
Ability to disable the guard bytes and poison of heap blocks to allow…

StephenHillAtViaivi:
… dynamic code analysis tools, such as Boundschecker, to detect read-overruns.

### DIFF
--- a/include/CppUTest/MemoryLeakDetector.h
+++ b/include/CppUTest/MemoryLeakDetector.h
@@ -216,7 +216,11 @@ public:
     void removeMemoryLeakInformationWithoutCheckingOrDeallocatingTheMemoryButDeallocatingTheAccountInformation(TestMemoryAllocator* allocator, void* memory, bool allocatNodesSeperately);
     enum
     {
+#ifdef CPPUTEST_DISABLE_MEM_CORRUPTION_CHECK
+        memory_corruption_buffer_size = 0
+#else
         memory_corruption_buffer_size = 3
+#endif
     };
 
     unsigned getCurrentAllocationNumber();
--- a/src/CppUTest/MemoryLeakDetector.cpp
+++ b/src/CppUTest/MemoryLeakDetector.cpp
@@ -32,6 +32,8 @@
 
 static const char* UNKNOWN = "<unknown>";
 
+static const char GuardBytes[] = {'B','A','S'};
+
 SimpleStringBuffer::SimpleStringBuffer() :
     positions_filled_(0), write_limit_(SIMPLE_STRING_BUFFER_LEN-1)
 {
@@ -481,7 +483,11 @@ SimpleMutex *MemoryLeakDetector::getMutex()
 
 static size_t calculateVoidPointerAlignedSize(size_t size)
 {
+#ifndef CPPUTEST_DISABLE_MEM_CORRUPTION_CHECK
     return (sizeof(void*) - (size % sizeof(void*))) + size;
+#else
+   return size;
+#endif
 }
 
 size_t MemoryLeakDetector::sizeOfMemoryWithCorruptionInfo(size_t size)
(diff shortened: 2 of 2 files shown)


StephenHillAtViaivi:
Hmmm, this changeset doesn't fix the problem with new() as the accounting information is added on the end (which is rather a shame).

basvodde:
What was the purpose?

StephenHillAtViaivi:
Currently CppuTest allocates extra trailing bytes on the end of any buffer dynamically allocated and fills them with a pattern. This pattern is checked when de-allocated, which is great for detecting write-overruns, but cannot detect read-overruns. There are many tools (Boundschecker, AppVerifier, Valgrind) which can detect read and write-overruns, but they are defeated by these trailing bytes. This change-set is designed to allow these trailing bytes to be disabled in the case where these tools are used. I will be updating this changeset with some fixes later.

basvodde:
Ah ok. That makes sense.

StephenHillAtViaivi:
Add fixes for plugins

src/CppUTest/MemoryLeakDetector.cpp
```diff
@@ -513,21 +519,25 @@ char* MemoryLeakDetector::reallocateMemoryAndLeakInformation(TestMemoryAllocator
 
 void MemoryLeakDetector::invalidateMemory(char* memory)
 {
+#ifndef CPPUTEST_DISABLE_HEAP_POISON
```
basvodde: Why is this @ifdef needed?

include/CppUTest/MemoryLeakDetector.h
```diff
@@ -216,7 +216,11 @@ class MemoryLeakDetector
     void removeMemoryLeakInformationWithoutCheckingOrDeallocatingTheMemoryButDeallocatingTheAccountInformation(TestMemoryAllocator* allocator, void* memory, bool allocatNodesSeperately);
     enum
     {
+#ifdef CPPUTEST_DISABLE_MEM_CORRUPTION_CHECK
+        memory_corruption_buffer_size = 0
+#else
         memory_corruption_buffer_size = 3
```
basvodde: I guess here you also want to do the sizeof(guardbyte) ?

basvodde: I guess by making GuardBytes empty when disabled, you do not require this variable at all and most of the #ifdef can be removed?

src/CppUTest/MemoryLeakDetector.cpp
```diff
@@ -481,7 +483,11 @@ SimpleMutex *MemoryLeakDetector::getMutex()
 
 static size_t calculateVoidPointerAlignedSize(size_t size)
 {
+#ifndef CPPUTEST_DISABLE_MEM_CORRUPTION_CHECK
```
basvodde: Why this one?

src/CppUTest/MemoryLeakDetector.cpp
StephenHillAtViaivi: Address Sanitizer, Boundschecker and Appverifier poison all buffers with a pattern, so this is just unnecessary. I have made it a different conditional as some dynamic code analysis tools might not provide this feature (e.g. Valgrind).

include/CppUTest/MemoryLeakDetector.h
StephenHillAtViaivi: Ah, I can't do this as the Microsoft tool-chain doesn't allow zero length arrays.

basvodde:
I'm uncomfortable with the #ifdef in the code. We are trying to remove #if and not add more. Still, I'll merge this as I understand the reason and cannot (yet) think of a better solution.

Although... just speculating... a different solution might be a command line option to turn it on/off. Most of it could be done dynamically. Unfortunately that means it would need to be off outside the tests so that it can be decided when parsing the parameters. Then, for global objects, the memory corruption check isn't done... that might lead to problems.. especially considering the nodes allocated differently...

Merged for now but requires more thought.

Issue #1148, opened 2018-03-22
autogen then configure then make install yields non-working tree


jbrandeb:
Thanks for CppUtest!  
When we are using a program that includes string.h or <cstring>, then compile fails with this sequence:
./autogen.sh
./configure --prefix=~/cpputest
make
make install

```
cd examples
make CPPUTEST_HOME=~/cpputest
compiling HelloTest.cpp
/usr/include/string.h:166:37: error: expected identifier before string constant
 extern char *strdup (const char *__s)
                                     ^
/usr/include/string.h:166:37: error: expected ‘,’ or ‘...’ before string constant
In file included from <command-line>:0:0:
/home/jbrandeb/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h:62:21: error: conflicting declaration of C function ‘char* cpputest_strdup_location(const char*, int)’
 #define strdup(str) cpputest_strdup_location(str, __FILE__, __LINE__)
                     ^
/home/jbrandeb/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h:55:14: note: previous declaration ‘char* cpputest_strdup_location(const char*, const char*, int)’
 extern char* cpputest_strdup_location(const char* str, const char* file, int line);
              ^~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/string.h:174:55: error: expected identifier before string constant
 extern char *strndup (const char *__string, size_t __n)
                                                       ^
/usr/include/string.h:174:55: error: expected ‘,’ or ‘...’ before string constant
/home/jbrandeb/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h:63:25: error: conflicting declaration of C function ‘char* cpputest_strndup_location(const char*, size_t, int)’
 #define strndup(str, n) cpputest_strndup_location(str, n, __FILE__, __LINE__)
                         ^
/home/jbrandeb/cpputest/include/CppUTest/MemoryLeakDetectorMallocMacros.h:56:14: note: previous declaration ‘char* cpputest_strndup_location(const char*, size_t, const char*, int)’
 extern char* cpputest_strndup_location(const char* str, size_t n, const char* file, int line);
              ^~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [../build/MakefileWorker.mk:504: objs/AllTests/HelloTest.o] Error 1
```

I'll issue a pull request with the code changed to generate the bug.

jbrandeb:
[...]

PR #1192, opened 2019-01-04
include string.h in StandardCLibrary.h

jacob-keller:
If memory leak detection for C code is enabled, the header file
MemoryLeakDetectorMallocMacros.h is included by the command line prior
to any of the raw source code for an object.

In this case, if CPPUTEST_HAVE_STRDUP is defined, the strdup function
will be redefined using a macro.

If the source code happens to include <string.h> or <cstring>, this will
happen *after* the memory leak detection header is included on the
command line.

Because of this, strdup will have already been redefined, and will then
result in compiler warnings similar to the following:

/usr/include/string.h:166:37: error: expected identifier before string constant
 extern char *strdup (const char *__s)
				     ^
/usr/include/string.h:166:37: error: expected ‘,’ or ‘...’ before string constant
In file included from <command-line>:
.../include/CppUTest/MemoryLeakDetectorMallocMacros.h:62:21: error: conflicting declaration of C function ‘char* cpputest_strdup_location(const char*, int)’
 #define strdup(str) cpputest_strdup_location(str, __FILE__, __LINE__)
		     ^~~~~~~~~~~~~~~~~~~~~~~~
.../include/CppUTest/MemoryLeakDetectorMallocMacros.h:55:14: note: previous declaration ‘char* cpputest_strdup_location(const char*, const char*, int)’
 extern char* cpputest_strdup_location(const char* str, const char* file, int line);
	      ^~~~~~~~~~~~~~~~~~~~~~~~

This happens because the "strdup (const char *__s) is converted by the
macro.

To fix this, <string.h> must be included prior to the setup of memory
leak detection.

Modify StandardCLibrary.h to include <string.h> if CPPUTEST_HAVE_STRING_H
is defined. This fixes the error and allows proper usage of the leak
detection strdup implementation.

### DIFF
--- a/include/CppUTest/StandardCLibrary.h
+++ b/include/CppUTest/StandardCLibrary.h
@@ -27,6 +27,11 @@
 /* Needed for some detection of long long and 64 bit */
 #include <limits.h>
 
+/* Needed to ensure that string.h is included prior to strdup redefinition */
+#ifdef CPPUTEST_HAVE_STRING_H
+#include <string.h>
+#endif
+
 #else
 
 #ifdef __KERNEL__


jacob-keller:
This fixes #1148 

I didn't bother to include the test case in #1149 since I don't really think a separate test case would be useful.

You can see this if memory leak detection is enabled, CPPUTEST_HAVE_STRDUP is defined, and <string.h> is included.

I'm not 100% sure if this is the best fix, but I can't think of another solution, since the memory leak detection headers are included by the command line, so we can't simply insert <cstring> prior to the strdup redefinition macros.

PR #1195, opened 2019-02-12
Add scoped enum support

mikanbako:
Scoped enumerations from C++11 are no implicit
conversions from their values to integral types.

Add two assertion macros.

ENUMS_EQUAL_INT(excepted, actual)
  Compares two enums which their underlying type is int

ENUMS_EQUAL_TYPE(underlying_type, excepted, actual)
  Compares two enums which they have the same underlying type


basvodde:
Thanks.

PR #1196, opened 2019-02-23
Support Microsoft UWP platform by extending VisualCppTimeInMillis().

dmitrykos:
Support Microsoft UWP platform by providing an alternative timer function in ```VisualCppTimeInMillis()``` due to absence of ```timeGetTime()``` API in UWP.

```TIMERR_NOERROR``` can be used to detect presence of timeGetTime() API.

UWP supports ```QueryPerformanceFrequency()/QueryPerformanceCounter()``` API which can be used for time calculation and this API is available on all Windows versions, so the proposal is to use this implementation for Windows platform in general and fallback to ```timeGetTime()``` or ```GetTickCount``` in case ```QueryPerformanceFrequency()``` fails due to some unknown reason.

```timeGetTime()``` also suffers from lower time resolution  which can be avoided by new implementation too.

### DIFF
--- a/src/Platforms/VisualCpp/UtestPlatform.cpp
+++ b/src/Platforms/VisualCpp/UtestPlatform.cpp
@@ -83,7 +83,7 @@ static long VisualCppTimeInMillis()
 	{
 		LARGE_INTEGER now;
 		QueryPerformanceCounter(&now);
-		return (now.QuadPart * 1000) / s_frequency.QuadPart;
+		return (long)((now.QuadPart * 1000) / s_frequency.QuadPart);
 	} 
 	else 
 	{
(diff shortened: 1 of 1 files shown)


basvodde:
Could you have a look at the build failure. It seems to be caused by the changes.

dmitrykos:
Did not know that project is configured to treat warning as error, applied necessary changes!

dmitrykos:
@basvodde, the CI failed again but it failed on DOS target and according logs somehow unrelated to my commit, e.g. /Dos/UtestPlatform.cpp. Could it be failure on CI's side? Would you please check.

basvodde:
Yes, thanks

PR #1199, opened 2019-03-06
Test Double (an alternative to CppUMock)

redengin:
see [README_TestDouble.md] for rationale and design.


redengin:
hmm, nullptr was standardized in C++11 but yet this harness requires C++03 compatibility. 
http://open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#903

whats the path forward?

basvodde:
I'm still not sure why this pull request, but I'll have a look at it at a later time.

We will keep compatibility as the unit test framework is frequently used in embedded systems and they often use old compilers.

redengin:
As far as I can tell, nullptr is now also a compiler intrinsic that can't be decoupled from the compiler.
https://stackoverflow.com/questions/10496824/how-to-define-nullptr-for-supporting-both-c03-and-c11

As well, this test harness's use of CLANG's opinion to deprecate implicit class methods seems like an attempt to redefine C++ in a way that is no longer compatible with the C++11 standard nor code written to C++ standards before C++11.

basvodde:
Gauntlet harness?

redengin:
By "gauntlet harness" I mean the set of test harnesses to validate pull requests (travis, coveralls, appveyor)

basvodde:
Ok. It isn't tuned ?

redengin:
By "tuned" I mean does it ignore some errors that would have been generated by standard CppUTest code (i.e. 0 vs nullptr initialization of pointers).

basvodde:
I'm not 100% sure, it should be in the CMake and autotools files.

redengin:
What do you think about [Test Double design](https://github.com/redengin/cpputest/blob/master/README_TestDouble.md)?

basvodde:
I'm yet to discover why :) If I have more time, I'll browse through the code.

PR #1200, opened 2019-03-15
Added missed header files when cpputest is being installed as external project though cmake

digiwhite1:
Added Shuffle.h and SimpleMutex.h in the list of headers to be installed in CMakeLists.txt.


basvodde:
tnx

PR #1204, opened 2019-03-30
Add public headers as an exported properties of the library

### LINKED ISSUE #1203: CMake: Both CppUTest and CppUTestExt does not install their public headers as a usage requirement
Both targets do not specify `PUBLIC_HEADER` and `INCLUDES` in their install command.
Instead, the public headers are installed manually.

This causes the `EXPORT`ed targets to not have their `INTERFACE_INCLUDE_DIRECTORIES` set to the correct value.
What this means is that when importing these targets using the `find_package(CppUTest/CppuTestExt)` command, the targets will be imported, but without the knowledge about their public headers location, so any target that links against them cannot compile correctly without manually adding the proper include directory.

This could be easily fixed by adding `PUBLIC_HEADER` and `INCLUDES` to the install command

avrahamshukron:
Fixes #1203 

Signed-off-by: Avraham Shukron <avraham.shukron@gmail.com>

### DIFF
--- a/src/CppUTest/CMakeLists.txt
+++ b/src/CppUTest/CMakeLists.txt
@@ -64,11 +64,15 @@ set(CppUTest_headers
         ${CppUTestRootDirectory}/include/CppUTest/SimpleMutex.h
 )
 
+set_target_properties(CppUTest PROPERTIES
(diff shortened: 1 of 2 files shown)


basvodde:
Thanks

PR #1205, opened 2019-04-04
Signedness corrected

offa:
Signedness correction (reported by clang 8).

@basvodde isn't `size_t` a better choice here anyway? Neither test count nor `shuffle_list()` parameter should become `< 1`.

### DIFF
--- a/include/CppUTest/Shuffle.h
+++ b/include/CppUTest/Shuffle.h
@@ -46,16 +46,15 @@ static inline int rand_(void) // rand_func_t
 // "Durstenfeld shuffle" according to Wikipedia
 static inline void shuffle_list(rand_func_t rand_func, size_t numElems, void* listToShuffleInPlace[])
 {
-    if( numElems > 0 )
+    if( numElems == 0 ) return;
+
+    for (size_t i = numElems - 1; i >= 1; --i)
     {
-        for (size_t i = numElems - 1; i >= 1; --i)
-        {
-            const size_t j = ((size_t)rand_func()) % (i + 1); // distribution biased by modulo, but good enough for shuffling
-            void* e1 = listToShuffleInPlace[j];
-            void* e2 = listToShuffleInPlace[i];
-            listToShuffleInPlace[i] = e1;
-            listToShuffleInPlace[j] = e2;
-        }
+        const size_t j = ((size_t)rand_func()) % (i + 1); // distribution biased by modulo, but good enough for shuffling
+        void* e1 = listToShuffleInPlace[j];
+        void* e2 = listToShuffleInPlace[i];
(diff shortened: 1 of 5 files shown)


src/CppUTest/TestRegistry.cpp
```diff
@@ -234,20 +234,20 @@ void TestRegistry::shuffleRunOrder(rand_func_t rand_func)
     {
         return;
     }
-    int numTests = getFirstTest()->countTests();
+    const size_t numTests = static_cast<size_t>(getFirstTest()->countTests());
```
basvodde: Isn't it better to fix the return parameter here?

basvodde:
I noticed this commit has the 'new' style casts. I think we always avoided these in CppUTest, though I noticed that in the recent pull requests 3 slipped through. I prefer still not using the 'new' style casts, mostly so that it keeps compileable on old systems.

src/CppUTest/TestRegistry.cpp
offa: From my point of view `countTests()` and `shuffle_list()` should use an unsigened type anyway. But I am not sure if this breaks any code. If you don't see any problems I'll change this. This also get's rid of the cast (https://github.com/cpputest/cpputest/pull/1205#issuecomment-480574854).

basvodde: Yeah, if you could try and see if that leads to problems then that would be great!

offa:
I updated my PR to use `size_t`.

include/CppUTest/Shuffle.h
```diff
@@ -44,15 +44,18 @@ static inline int rand_(void) // rand_func_t
 }
 
 // "Durstenfeld shuffle" according to Wikipedia
-static inline void shuffle_list(rand_func_t rand_func, int numElems, void* listToShuffleInPlace[])
+static inline void shuffle_list(rand_func_t rand_func, size_t numElems, void* listToShuffleInPlace[])
 {
-    for (int i = numElems - 1; i >= 1; --i)
+    if( numElems > 0 )
```
basvodde: Could you change this to:

if (numElems == 0) return;

So that the code is easier to read and not indented

include/CppUTest/TestRegistry.h
```diff
@@ -33,6 +33,7 @@
 #ifndef D_TestRegistry_h
 #define D_TestRegistry_h
 
+#include <stddef.h>
```
basvodde: You'll need to include the StdC include instead of including Standard C directly.

basvodde:
Looks good. 2 small comments on it.

offa:
Thanks for your review! I have addressed both issues.

basvodde:
Tnx

PR #1213, opened 2019-04-22
In README: CPPUTEST_HOME variable prefixed with a $

saraf:
The CPPUTEST_HOME variable to be added to CPPFLAGS perhaps needs to be prefixed with a $ to work in a Makefile.


basvodde:
Tnx

PR #1214, opened 2019-04-26
JUnit output: Encode special chars in XML Text

alameth:
This change extends escaping of XML special characters to text blocks as well
as attributes.  It also changes the mapping to use standard escape encoding:

	"   &quot;
	<   &lt;
	>   &gt;
	&   &amp;

This change maintains the historic behavior of mapping newlines to "{newline}"
in the XML text. We're not sure whether this is optimal for Bamboo and Jenkins,
although our experience to date is it's better then emitting the raw newlines
or mapping newlines to spaces.

This implementation is expensive, calling SimpleString::replace() repeatedly
over the same string.  The cost was mitigated by modifying the replace() method
to exit quickly if the pattern is not found.

### DIFF
--- a/include/CppUTest/JUnitTestOutput.h
+++ b/include/CppUTest/JUnitTestOutput.h
@@ -71,6 +71,7 @@ protected:
     virtual void writeTestSuiteSummary();
     virtual void writeProperties();
     virtual void writeTestCases();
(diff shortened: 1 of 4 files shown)


basvodde:
Looks good, thanks!

Issue #1206, opened 2019-04-04
Clang 8 build fails due empty expressions


offa:
Compilation on clang 8 (C++11 and Werror enabled) fails with many _"empty expression statement has no effect"_ warnings.

This problem is easy to fix:

1. Remove the `;` from the macro definition – callers are required to end the statement with `;`.
1. Add the `;` to the macro definition and remove it from any internal callers – only clients get the warning

Both solution can potentially break existing code, depending on whether clients have ended the statement with `;` or not.

I'd prefer _1.)_ as it maintains the function-like usage. This is also the fix I previously included into #1205, but removed it as I'm not sure about the compatibility impacts.

basvodde:
Sucks. I guess we'll need to do with (1). Does it have any potential to mess up or is it just lead to compile errors?

offa:
It should only lead to compile errors. Within CppUTest I don't even expect any impact as those macros are already called correctly:

```cpp
UT_THROW_BAD_ALLOC_WHEN_NULL(memory);
```

3rd party callers outside of this project could omit the `;` and see a compile error therefore.

I'll create a PR once fixed.

offa:
Fixed via #1215

PR #1215, opened 2019-04-27
Empty expression warnings fixed

offa:
Empty expression warnings fixed (#1206).

Together with #1205 this concludes the Clang 8 support.

### DIFF
--- a/include/CppUTest/MemoryLeakWarningPlugin.h
+++ b/include/CppUTest/MemoryLeakWarningPlugin.h
@@ -31,8 +31,8 @@
 #include "TestPlugin.h"
 #include "MemoryLeakDetectorNewMacros.h"
 
-#define IGNORE_ALL_LEAKS_IN_TEST() MemoryLeakWarningPlugin::getFirstPlugin()->ignoreAllLeaksInTest();
-#define EXPECT_N_LEAKS(n)          MemoryLeakWarningPlugin::getFirstPlugin()->expectLeaksInTest(n);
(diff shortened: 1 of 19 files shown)


basvodde:
I'm a bit about the while (0) causing a "always false" warning. Why wouldn't it?

offa:
I don't hope so. The `do { .. } while(0)` is a well known C / C++ construct for multiline macros. For C it's the only recommended way by MISRA C to solve this problem.

Do you have some specific platforms in mind?

basvodde:
Not really, just a worry. But let's proceed with this.

jwgrenning:
This did break pretty much all my training examples.  Now I have to add semi-colons to CHECK, LONGS_EQUAL, FAIL usages.

I wonder how many more people will have their code broken.

offa:
Do you write different code in your trainings than in the book, articles and papers? :confused:

jwgrenning:
I do, and often guiding clients.  With a closer look, it's not all my training examples. I was somewhat inconsistent with semi-colons on checks.  I've fixed them. Now its consistent, which is good too.  

The book code was consistent to start with :-)

jwgrenning:
We might want to warn the mailing list.  Was the mailing list warned? I could not find it quickly. I would have been really surprised if I did not remember this thread on github.

offa:
There is a mailing list? :astonished:

jwgrenning:
I'll warn the cpputest google group for starters.

PR #1217, opened 2019-05-24
0️⃣ add suffixes to floating point literals

noahp:
To enable compiling with `-Werror=unsuffixed-float-constants`, add some
suffixes to double literal constants.

Without using gnu extensions, there's no suffix available for double
literals, so instead use the suffix for `long double`, `L`, and cast to
explicit `double`. This should result in generated code equivalent to
using a double literal, even if precision loss occurs (it doesn't in
these cases though, and implicit conversion should be equivalent, but
just to be sure).

### DIFF
--- a/include/CppUTestExt/MockCheckedActualCall.h
+++ b/include/CppUTestExt/MockCheckedActualCall.h
@@ -274,7 +274,7 @@ public:
     virtual unsigned int returnUnsignedIntValue() _override { return 0; }
     virtual unsigned int returnUnsignedIntValueOrDefault(unsigned int value) _override { return value; }
 
-    virtual double returnDoubleValue() _override { return 0.0; }
+    virtual double returnDoubleValue() _override { return (double)0.0L; }
     virtual double returnDoubleValueOrDefault(double value) _override { return value; }
 
     virtual const char * returnStringValue() _override { return ""; }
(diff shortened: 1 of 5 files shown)


basvodde:
Isn't a 0.0 by default a double, which is why there is no suffix for double?

noahp:
> Isn't a 0.0 by default a double, which is why there is no suffix for double?

That's correct! on some target arch's, doubles are expensive and should be used cautiously. Enabling the `-Werror=unsuffixed-float-constants` compiler warning enables us to identify accidental double usage, but has the downside of requiring the `(double)0.0L` syntax for double literals when not using gcc extensions (those provide a `d` suffix for explicit double literals, eg `0.0d`).

It's useful to enable the same `-Werror=unsuffixed-float-constants` flag in the off-target unit tests, since those are faster for our developers to get feedback.

basvodde:
The double usage isn't really likely to cause a problem in performance. Is there any other advantage of accepting this PR? Personally I wouldn't really want to see all the casts and suffixes in the code...

noahp:
Doubles are quite costly on certain architectures:
https://mcuoneclipse.com/2019/03/29/be-aware-floating-point-operations-on-arm-cortex-m4f/

That said, it's easy enough for us to select different compilation flags for the cpputest library vs. our tests/code under test. Just offering this PR in case it was interesting, feel free to decline!

PR #1221, opened 2019-06-05
Fix #1219.

### LINKED ISSUE #1219: The value of .withOutputParameterReturning() is changeable setting the expectation
I think setting the expectation `.withOutputParameterReturning()` should copied the value pointed to. However, it stores the pointer and dereferences it on the `actualCall().withOutputParameter()`.  

If the call  `.withOutputParameterReturning()` copies the given memory it would allow the memory to be changed between expect and actual calls.

The problem was encountered when trying to mock an interface where the `.withOutputParameterReturning()` is given a chunk of dynamically allocated memory which is free'd (and reused) before the actual call.

This is the simplest example I can come up with to reproduce the issue.

```cpp
TEST(MockParameterTest, outputParameterReturningShouldCopyOnExpect)
{
    int param = 1;
    int retval = 2;

    mock().expectOneCall("function").withOutputParameterReturning("parameterName", &retval, sizeof(retval));
    retval = 7; 

    mock().actualCall("function").withOutputParameter("parameterName", &param);
    CHECK_EQUAL(param, 1); // Check fails !!!

    mock().checkExpectations();
}

```

asgeroverby:
On expect .withOutputParameterReturning() store content of
output parameter.

### DIFF
--- a/include/CppUTestExt/MockCheckedExpectedCall.h
+++ b/include/CppUTestExt/MockCheckedExpectedCall.h
@@ -79,7 +79,7 @@ public:
     virtual MockExpectedCall& onObject(void* objectPtr) _override;
 
     virtual MockNamedValue getInputParameter(const SimpleString& name);
-    virtual MockNamedValue getOutputParameter(const SimpleString& name);
+    virtual MockNamedValue* getOutputParameter(const SimpleString& name);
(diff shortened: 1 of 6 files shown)


asgeroverby:
Moved the PR to a seperate branch.

Issue #1193, opened 2019-01-11
IAR 8.1 CppUTest fails to build


nowox:
The project `cpputest\platforms\iar\CppUTest.eww` fails to build with IAR 8.1:

```
CommandLineArguments.cpp  
Error[Pe135]: namespace "std" (declared at line 185 of "C:\Program Files (x86)\IAR Systems\Embedded Workbench 8.1\arm\inc\c\yvals.h") has no member "nullptr_t" cpputest\include\CppUTest\SimpleString.h 182 
Error while running C/C++ Compiler 
```

I have added `cstddef` as preinclude:

```
Preinclude file: 
C:\Program Files (x86)\IAR Systems\Embedded Workbench 8.1\arm\inc\cpp\cstddef
```

But then I have these errors:

```
Error[Pe079]: expected a type specifier cpputest\src\CppUTest\SimpleString.cpp 548 
Warning[Pe815]: type qualifier on return type is meaningless cpputest\src\CppUTest\SimpleString.cpp 548 
Error[Pe018]: expected a ")" cpputest\src\CppUTest\SimpleString.cpp 548 
Error while running C/C++ Compiler 
```

Which occur here: 

```c
#if __cplusplus > 199711L && !defined __arm__
SimpleString StringFrom(const std::nullptr_t __attribute__((unused)) value)
{
    return "(null)";
}
```

It seems neither `__arm__` nor `__cplusplus` is defined. So I added them as defines. I also disabled the exceptions because the library is meant to be used on embedded devices. 

```
CPPUTEST_STD_CPP_LIB_DISABLED=1
__arm__
__cplusplus=200000L
```

This time the project builds without errors

watersne:
IAR's compiler guide for 8.10 claims `__cplusplus=199711L` is defined, but I am seeing that `__cplusplus` is actually `201402L` on my system. IARs docs are out of date. To get past this I had to add `!defined __ICCARM__` to that line. That feels like a hack.

basvodde:
I think that this has been resolved? Can you confirm that?

Is there any reasonable way to get the IAR build in the travis CI?

PR #1234, opened 2019-09-06
Add header that defines std::nullptr_t.

dandwyer:
My GHS 7.1.4 compiler requires cstddef to get std::nullptr_t. This is somewhat expected given:
https://en.cppreference.com/w/cpp/types/nullptr_t
It may be better to put the cstddef include in StandardCLibrary.h, but I wanted to localize the change for context.


basvodde:
Thanks. Instead, the std::null_ptr ought to be removed from CppUTest.

dandwyer:
Are you saying you want the following function removed entirely from the header and source?:
SimpleString StringFrom(const std::nullptr_t value)

basvodde:
If possible, yes. Need to check why it was there, but usually we do not want things like nullptr_t in the code because CppUTest ought to work with not fully latest standard compilers.

Either it got in because I didn't notice it, or it needs to be there in a way that can work when compilers don't fully support this.

dandwyer:
Thank you for your quick replies to this pull request. I am concerned I may do more damage than good if I try to resolve your higher level concerns. In this case, my compiler has no issue with the use of std::nullptr_t, but it does need the header that provides the definition. Is this pull request viable as a quick fix on the way to a larger scope change?

dandwyer:
Just found a datapoint that someone else has added cstddef as a pre-include to their build to work around this issue:
https://github.com/cpputest/cpputest/issues/1193

basvodde:
Since it is the only place we use null_ptr, I cannot merge this pull request as it will take it away from where it needs to go.

Removing the null_ptr shouldn't be a large change.

basvodde:
I spend a bit more time looking at this. It looks like the nullptr_t is around the appropriate #ifdef already and outside the class, so that can stay.

I cannot accept this pull request though as including a C++ library header outside the CStandardLibrary.h is not ok. When I looking in the StandardLibrary.h, it already has a #include <stddef.h>  in there, which ought to be equal to the include you are trying to add.

Would you mind checking why that one isn't enough?

Thanks!

basvodde:
Yeah, this will do. Thanks!

dandwyer:
@basvodde , Thanks for your help on these pull requests. I am happily building off of master now!
-Dan

PR #1238, opened 2019-09-11
Travis image update and C++98 compat warnings

offa:
Travis updated to the latest bionic image.

This caused the non-C++11 CMake clang build to fail because of  warnings:

* -Wc++98-compat
* -Wc++98-compat-pedantic
* -Winconsistent-missing-destructor-override

Since all three are only disabled with `C++11=ON` it failed to compile on clang with C++11 disabled.

While it makes sense to disable the last one somehow, I'm not that sure about the C++98 compat.

Other options:

1. Disable the C++98 warnings only for clang
1. Set C++98/03 explicitly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: cpp
 
-dist: trusty
+dist: bionic
 
 matrix:
     include:
--- a/scripts/travis_ci_build.sh
+++ b/scripts/travis_ci_build.sh
@@ -34,8 +34,14 @@ if [ "x$BUILD" = "xautotools" ]; then
 fi
 
 if [ "x$BUILD" = "xcmake" ]; then
+    BUILD_ARGS=("-DWERROR=ON" "-DC++11=${CPP11}")
+
+    if [ "x$CPP11" = "xOFF" ]; then
+        BUILD_ARGS+=("-DCMAKE_CXX_STANDARD=98")
+    fi
+
     cmake --version
-    cmake -DWERROR=ON -DC++11=${CPP11} ..
+    cmake "${BUILD_ARGS[@]}" ..
     make
     ctest -V
 fi


offa:
There's a connection issue on openwatcom:

> Connecting to ftp.openwatcom.org (ftp.openwatcom.org)|70.165.27.117|:80... failed: Connection timed out.
> 
> Retrying.

basvodde:
Before disabling, I'm curious about what the failures were? Should we do something with the warnings?

offa:
I set up a Clang with Cpp11=OFF build, you can see the warnings [here](https://travis-ci.org/offa/cpputest/jobs/587067947). Probably recent versions of clang use a modern C++ by default?

basvodde:
Ok, so then we would need to just swap it and add a C++98 build?

offa:
I came across a better solution than disabling the warnings:

The C++11=OFF ci build can pass the C++ standard to cmake by argument. Since that  build should cover the pre-C++11 scenario, it could simply force C++98 (or 03). This prevents compilers from using a more recent version even tough they update their defaults.

Only a small change of the Travis build script and it doesn't touch any warnings.

Users with recent compilers who run into the warnings can either switch to C++11 (preferred) or use C++98/03 their own.

I think this is a better and less brutal solution than disabling the warnings. If you agree with that @basvodde I'll update this PR accordingly. Other suggestions are still welcome though :-).

basvodde:
.. probably. Would like to see how it would become.

basvodde:
Not done yet, right?

offa:
It depends, if you are ok with the solution the PR is ready. I don't have found any issues so far.
Though I'm wondering why the coverage degraded.

basvodde:
Ok, I expected you to do something else :)

I had hoped for different builds for the different C++ standard versions. But I'll merge this for now and open an issue for that.

offa:
I see, this shouldn't be a problem. Feel free to assign the issue to me.

basvodde:
I never assigned items to anyone, but people can take it, I guess :)

PR #1241, opened 2019-09-30
Allow nullpointers in Memory comparer when comparison size is zero

sensslen:
This Pull request ignores nullpointers when comparing between memory regions with size 0

### DIFF
--- a/src/CppUTest/Utest.cpp
+++ b/src/CppUTest/Utest.cpp
@@ -486,6 +486,7 @@ void UtestShell::assertDoublesEqual(double expected, double actual, double thres
 void UtestShell::assertBinaryEqual(const void *expected, const void *actual, size_t length, const char* text, const char *fileName, int lineNumber, const TestTerminator& testTerminator)
 {
     getTestResult()->countCheck();
+	if (length == 0) return;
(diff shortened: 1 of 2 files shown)


basvodde:
Interesting. No case where this logic can lead to surprises?

sensslen:
I don't see any - when comparing zero sized buffers even memcmp does not do much. Either ways comparing buffers of size zero must always be correct, as there is nothing to compare.

basvodde:
I'm mostly thinking whether this might lead to accidentally allowing a test to pass that should have failed.

I guess I'm not sure when this case would actually happen in a proper test...

sensslen:
For me the line below the inserted one does not make much sense - I expect that memcmp (even when called with two null pointers) would fail with a segmentation fault due to invalid memory region access.

sensslen:
Well we do have this kind of case where we check a function to operate correctly when fed with different buffer sizes. In order to avoid code duplication we use the same test code even for zero sized comparisons.

basvodde:
Ok, can't think of serious harm in this, I'll merge it.

basvodde:
Thanks!

Issue #1242, opened 2019-10-02
C++ Standards in build


basvodde:
Add a build for all the C++ standards in Travis.

offa:
Here's what's covered in the ci builds (only the standard related):

| Compiler | C++11 option | Standard used |
|-------------|-------------------|--------------------|
| GCC | On | C++11 |
| GCC | Off | C++98 |
| Clang | On | C++11 |
| Clang | Off | C++98 |

What should we add further? C++14 and 17 for both Gcc and Clang? Is C++03 necessary too?

basvodde:
I think it would be good to add all of the C++ versions.

offa:
For both compiler?

basvodde:
I guess clang would do as it tends to be the picker

offa:
This has been solved already, I guess we can close here @basvodde.

PR #1246, opened 2019-10-04
CI builds for additional C++ standards

offa:
CI Builds for _gcc_ and _clang_ using _C++98/11/14/17_ (#1242). I intentionally skipped _C++03_ since it shouldn't differ from C++98 and the build list is already long. I can add it at any time if it's reasonable – no problem.

C++11 and above need to set the `C++11=ON` option to correct warning settings. If another C++ standard is passed via `CMAKE_CXX_STANDARD` it shouldn't set the compiler standard option though. If compilation with a newer standard than C++11 is something users typically do, we should rethink this.


offa:
The DOS build failed with a strange ALSA issue.

.travis.yml
```diff
@@ -5,22 +5,38 @@ dist: bionic
...
     - compiler: gcc
-      env: BUILD=cmake CPP11=OFF
+      env: BUILD=cmake CPP_STD=98 CPP11=OFF
+    - compiler: gcc
+      env: BUILD=cmake CPP_STD=11 CPP11=ON
+    - compiler: gcc
+      env: BUILD=cmake CPP_STD=14 CPP11=ON
     - compiler: gcc
-      env: BUILD=cmake CPP11=ON
+      env: BUILD=cmake CPP_STD=17 CPP11=ON
```
basvodde: The CPP11=ON is a bit weird here :) Can we do something about that?

basvodde:
Any idea about the DOS failure?

.travis.yml
offa: Indeed, this is horrible. I have updated the PR to take `CMAKE_CXX_STANDARD` into account when setting the warnings. If it's set with a standard >= C++11 it'll add the warnings accordingly, making the `CPP11` option no longer necessary.

offa:
> Any idea about the DOS failure?

Seems like a temporary problem. Previous and current builds don't have this problem.

basvodde:
Tnx!

PR #1254, opened 2019-11-07
Return non-zero for command line errors

### LINKED ISSUE #1243: Return code is zero if command line parsing fails
CommandLineTestRunner::RunAllTests returns 0 if command line arguments are invalid.
We faced it by chance when runner scripts go broken and provided incorrect command line. No tests were run, but the application looked like all tests passed due to success code returned.
Is this expected behavior? How to detect that nothing was run due to some error and fail?

vpod:
If command line arguments parsing fails, return non-zero from CommandLineTestRunner::RunAllTests.

Fixes #1243

### DIFF
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2007, Michael Feathers, James Grenning and Bas Vodde
+ * Copyright (c) 2019, Nokia
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -73,7 +74,7 @@ CommandLineTestRunner::~CommandLineTestRunner()
 
 int CommandLineTestRunner::runAllTestsMain()
 {
-    int testResult = 0;
+    int testResult = 1;
 
     SetPointerPlugin pPlugin(DEF_PLUGIN_SET_POINTER);
     registry_->installPlugin(&pPlugin);
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -73,7 +73,7 @@ CommandLineTestRunner::~CommandLineTestRunner()
 
 int CommandLineTestRunner::runAllTestsMain()
 {
-    int testResult = -1;
+    int testResult = 1;
 
     SetPointerPlugin pPlugin(DEF_PLUGIN_SET_POINTER);
     registry_->installPlugin(&pPlugin);
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -1,6 +1,5 @@
 /*
  * Copyright (c) 2007, Michael Feathers, James Grenning and Bas Vodde
- * Copyright (c) 2019, Nokia
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -74,7 +73,7 @@ CommandLineTestRunner::~CommandLineTestRunner()
 
 int CommandLineTestRunner::runAllTestsMain()
 {
-    int testResult = 1;
+    int testResult = -1;
 
     SetPointerPlugin pPlugin(DEF_PLUGIN_SET_POINTER);
     registry_->installPlugin(&pPlugin);
--- a/tests/CppUTest/CommandLineTestRunnerTest.cpp
+++ b/tests/CppUTest/CommandLineTestRunnerTest.cpp
@@ -139,14 +139,14 @@ TEST(CommandLineTestRunner, NoPluginsAreInstalledAtTheEndOfARunWhenTheArgumentsA
 
 }
 
-TEST(CommandLineTestRunner, ReturnsNegativeWhenTheArgumentsAreInvalid)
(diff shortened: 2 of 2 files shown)


src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2007, Michael Feathers, James Grenning and Bas Vodde
+ * Copyright (c) 2019, Nokia
```
basvodde: We usually do not add the copyright of the company where the contributor works. If we would do so, there would be huge list of copyright notifications, not think that would benefit anyone.

basvodde:
Thinking this over more. Returning 1 doesn't really make sense as it returns the # of tests that were run...

basvodde:
It would make more sense to return a negative number on failure...

vpod:
> Thinking this over more. Returning 1 doesn't really make sense as it returns the # of tests that were run...

I guessed so. I'll fix.

basvodde:
I think we'll need to reconsider this. If we change the error code, then it is likely that a lot of the CppUTest builds will start failing, right?

Can we go back to the why? The reason is that a test run of 0 tests ought to be considered successful?

In what case would you want a test case of 0 tests to actually be seen as successful?

vpod:
I'll be verbose.
I've met few strange things in cpputest that were hiding bugs from our CI (examples based on cpputest master, this PR baseline):

1. (_this issue/PR_) passing invalid command-line parameter results in printing help with no tests executed (event attempted), RunAllTests() returns 0 and no way to detect that there was an error actually:
```
./tests/CppUTest/CppUTestTests -v -some-error
usage [-v] [-c] [-p] [-lg] [-ln] [-ri] [-r#] [-g|sg|xg|xsg groupName]... [-n|sn|xn|xsn testName]... [-s [randomizerSeed>0]] ["TEST(groupName, testName)"]... [-o{normal, junit, teamcity}] [-k packageName]
Process finished with exit code 0
```
I believe this should return non-zero (-1?) and should not cause issues to any existing CIs anywhere, unless those do not run tests due to bugs in CI scripts.

2. (_your last statement_) test filter results in no tests executed (often due to some typo):
```
./tests/CppUTest/CppUTestTests -v -g InvalidGrouppp
OK (749 tests, 0 ran, 0 checks, 0 ignored, 749 filtered out, 1 ms)
Process finished with exit code 0
```
Is this an issue? Probably. At least, CI system wasted time to build and run the code. You would not want having that on cycle accurate (= very slow) simulator of some chip or even real hardware board, on which startup may take minutes. It could do more useful work instead of running no tests. However, it may be that there are valid cases for this behavior.
I don't try to fix it in this PR. I'll add a test to indicate current behavior (as legacy requirement) to make it explicit.

3. (your last statement as well?) no tests linked to the test binary (e.g. used static test library and forgot to import test groups):
```
./tests/CppUTest/CppUTestNoTests -v
OK (0 tests, 0 ran, 0 checks, 0 ignored, 0 filtered out, 0 ms)
Process finished with exit code 0
```
Is this an issue? Very likely.
I have not seen it in such pure version, but in hard to detect cases when new static library is added to existing test executable, but new test groups are not imported.
I don't try to fix it in this PR. But probably someone may want to do it. I'm not sure how to add a test for this one, so I'll skip. However I tested manually ;) .

In my opinion the first item above is definitely a problem.

basvodde:
But in all cases, return 0 is an error. So, I'm unclear why this would cause a problem in the CI system?

vpod:
I have no answer. I would personally prefer all listed above cases to return failure.
As this argument parsing error is in separate class from the basic runner that could return non-zero for 0 run tests, my proposal is to split the "0 tests run" topic to separate issue.

BTW, how to retrigger the appveyor tests? They got stuck (v140 version is in the queue for many hours).

vpod:
If you want, I can address "0 test run" as well, but not right here to keep deliveries focus on single issue.

basvodde:
Looking this through, I think you are right to put the default on 1. I was confused with the return value (sorry!). Suggest to go back to keeping the return value at 1 which marks an error and therefore also returns an error when there are 0 tests.

vpod:
Not sure if we are on the same page.
`int CommandLineTestRunner::runAllTestsMain()` line 81 is where the first issue comes: if `parseArguments()` fails, default value of `testResult` is returned. It should not be zero.

For the other issues, `testResult = runAllTests();` is executed inside the `if` (line 82). So that function needs fixing as it returns 0 far all error cases. My proposal is to handle that function separately as it has too many conditions to verify in tests.

("Just Talk" is really great idea of yours. We would already conclude something. :-) )

basvodde:
Ok, so you propose the change on your first commit, right?

Now, I think that is good. I was wrong in my first comment, sorry about that. :) Can you change the -1 back to 1, then I'll merge this.

vpod:
Right.

Should I create issue(s) for the scenarios when 0 tests were run and zero is returned?

vpod:
@basvodde: so, should I create issue(s) for the scenarios when 0 tests were run and zero is returned? (with valid command line arguments, like my above examples 2 and 3)

basvodde:
I'm not sure anymore. Although I think 0 tests is probably usually a failure, the argument could be made that it should pass. Let's skip it for now.

Issue #991, opened 2016-05-25
Gcc 5.3.0 warns about sized operator delete


arstrube:
```
./../include/CppUTest/MemoryLeakDetectorNewMacros.h:45:10: warning: 'void operator 
delete(void*, size_t)' is a usual (non-placement) deallocation function in C++14 (or with 
-fsized-deallocation) [-Wc++14-compat]
     void operator delete (void* mem, size_t size) UT_NOTHROW;
```
I suppose we might want "-Wno-c++14-compat".

It compiles, links and tests okay, but there are lots of warnings, because for each 'real' warning, it also spits out:
```
cc1plus: warning: unrecognized command line option '-Wno-weak-vtables'
cc1plus: warning: unrecognized command line option '-Wno-exit-time-destructors'
cc1plus: warning: unrecognized command line option '-Wno-global-constructors'
cc1plus: warning: unrecognized command line option '-Wno-keyword-macro'
cc1plus: warning: unrecognized command line option '-Wno-reserved-id-macro'
cc1plus: warning: unrecognized command line option '-Wno-disabled-macro-expansion'
cc1plus: warning: unrecognized command line option '-Wno-c++11-long-long'
```
These tend to go away when the warning causing it is silenced.

basvodde:
What does the warning even mean :)

basvodde:
Before disabling that, I guess we'll need to have a C++14 compilation in the CI :)

arstrube:
My interpretation: It wants to warn you that you shouldn't define this, because if you do, it won't be compatible with C++14 which also defines it. When you do have C++14, you might get no problem -- or worse, I guess.....

If you take this warning serious, it means adding the sized destructor is not compatible with C++14, so silencing the previous warning might have been the better option :P

arstrube:
We don't need a C++14 compilation on the CI, me thinks. But someone needs to do it offline, to see what happens. To play around with all the different possibilities....

arstrube:
1. Why is it that CppUTest is built without `-Werror`?
2. The warnings are alive and well, plus they trigger more warnings about unknown warning options.
3. The warning does not appear with Gcc old enough to not know about C++14. Which version are we using on the TravisCi Gcc build?
This is REALLY UGLY to say the least. If -Werror was enabled, it would also fail the build.

arstrube:
[...]

PR #1251, opened 2019-10-29
Gcc 5 c++14-compat warning

offa:
Disables the _c++14-compat_ warning regardless of the used C++ standard (#991).

PR #1255, opened 2019-11-17
Gcc 5 c++14-compat warning  - std check

offa:
Another approach to #991 by including these operators only when compiling C++14 or newer. This version does not depend on any build system.


basvodde:
Ok. What other pull requests and issues can we close now?

offa:
The other PR was #1251 (have closed it already), the related issue is #991 (wasn't reopened).

PR #1259, opened 2019-11-26
Extension attribute replaced

offa:
Replaces the GNU `__attribute__((unused))` with a simple cast to void which should be portable on all systems.


offa:
Wooops, I'll replace the C++ cast ...

PR #1260, opened 2019-11-27
Unsigned comparison

### LINKED ISSUE #1231: comparison with size_t
/src/CppUTest/MemoryLeakDetector.cpp: line 53 
positions_left can not be below zero
probably better
if (positions_left == 0) return;

offa:
Unsigned comparison corrected (fixes #1231).


offa:
The DOS Build has become stuck because of an Alsa issue again ...

basvodde:
Yeah, that DOS build is annoying.... need to think about that.

Anyways, thanks for this change.

PR #1261, opened 2019-11-28
nullptr is defined in Std namespace

bneumann:
Not 100% sure if this is really necessary, but I had some problems compiling without the std so I thought it might do no harm


basvodde:
Yeah, we added it, removed it, added it and removed it.

But now with the new #ifdef around it, let's try to add it again. Thanks!

PR #1266, opened 2019-12-06
Add package to JUnit report file name

### LINKED ISSUE #1256: Allow to alter JUnit output file name
We've come up with innovative way to screw things up: the test binaries have the same test group, but test different things (probably not uncommon). If they run in the same "make check" or similar, we get only 1 file instead of as many as binaries.
It seems that cpputest either needs to append "package" to file name (as we can easily say that these tests are for different packages), or provide some other way to alter file names to make them unique (extra "prefix" argument?).

Am I right that currently there is no way to alter the junit output behavior? IMO, we need to add it.
Seems to be lacking part of #263...

vpod:
If JUnit package is set, add it to the report file name.
Slightly expand list of special characters not allowed in file names
as package is provided via command line.

Closes #1256

### DIFF
--- a/src/CppUTest/JUnitTestOutput.cpp
+++ b/src/CppUTest/JUnitTestOutput.cpp
@@ -151,8 +151,24 @@ void JUnitTestOutput::printCurrentTestStarted(const UtestShell& test)
 SimpleString JUnitTestOutput::createFileName(const SimpleString& group)
 {
     SimpleString fileName = "cpputest_";
+    if (!impl_->package_.isEmpty()) {
+        fileName += impl_->package_;
+        fileName += "_";
+    }
     fileName += group;
+
+    // special character list based on: https://en.wikipedia.org/wiki/Filename
     fileName.replace('/', '_');
+    fileName.replace('\\', '_');
+    fileName.replace('?', '_');
+    fileName.replace('%', '_');
+    fileName.replace('*', '_');
+    fileName.replace(':', '_');
+    fileName.replace('|', '_');
+    fileName.replace('"', '_');
+    fileName.replace('<', '_');
+    fileName.replace('>', '_');
+
     fileName += ".xml";
     return fileName;
 }
@@ -208,7 +224,7 @@ void JUnitTestOutput::writeTestCases()
         SimpleString buf = StringFromFormat(
                 "<testcase classname=\"%s%s%s\" name=\"%s\" assertions=\"%d\" time=\"%d.%03d\" file=\"%s\" line=\"%d\">\n",
                 impl_->package_.asCharString(),
-                impl_->package_.isEmpty() == true ? "" : ".",
+                impl_->package_.isEmpty() ? "" : ".",
                 impl_->results_.group_.asCharString(),
                 cur->name_.asCharString(),
                 cur->checkCount_ - impl_->results_.totalCheckCount_,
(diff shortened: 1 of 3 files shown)


vpod:
How to retry Travis CI as it says it was timeout.
Or was it may change that caused timeout?

src/CppUTest/JUnitTestOutput.cpp
```diff
@@ -151,8 +151,24 @@ void JUnitTestOutput::printCurrentTestStarted(const UtestShell& test)
 SimpleString JUnitTestOutput::createFileName(const SimpleString& group)
 {
     SimpleString fileName = "cpputest_";
+    if (!impl_->package_.isEmpty()) {
+        fileName += impl_->package_;
+        fileName += "_";
+    }
     fileName += group;
+
+    // special character list based on: https://en.wikipedia.org/wiki/Filename
```
basvodde: Could you extract the replacement in a function. Also please make it a look over a static array :)

tests/CppUTest/JUnitOutputTest.cpp
```diff
@@ -162,7 +162,7 @@ class JUnitTestOutputTestRunner
 
 public:
 
-    JUnitTestOutputTestRunner(TestResult result) :
+    explicit JUnitTestOutputTestRunner(const TestResult& result) :
```
basvodde: Please remove the "explicit"

vpod: why? is implicit conversion expected? no tests failed (my static analyzer complaints about this pattern)

vpod: By the way, what about TestResult: I see it to be a pattern to pass it by value. Is it OK that I changed it to cost reference?
(As I'm now touching most of test outputs to handle returning error for 0 run tests, I'm just fixing those where I touch the file. But I can stop.)

offa:
> How to retry Travis CI as it says it was timeout.
Or was it may change that caused timeout?

This is an issue with the DOS build, not related to your PR.

tests/CppUTest/JUnitOutputTest.cpp
basvodde: re: explicit. Actually, you can keep it. I thought the explicit keyword wasn't used anywhere, but I am wrong.

basvodde: re: TestResult. Since a TestResult is just a simple value object, there is no real benefit for passing it as reference. It is not needed... but also not really needed to be removed as... it isn't very important.

vpod: Test result: I would call this "simple value object":
```cxx
private:
    TestOutput& output_;
    int testCount_;
    int runCount_;
    int checkCount_;
    int failureCount_;
    int filteredOutCount_;
    int ignoredCount_;
    long totalExecutionTime_;
    long timeStarted_;
    long currentTestTimeStarted_;
    long currentTestTotalExecutionTime_;
    long currentGroupTimeStarted_;
    long currentGroupTotalExecutionTime_;
```
I.e. copying so many bytes is questionable. But I agree that probably the compiler is clever enough to remove the copy.


Bottom line: I'll keep this line as as and only fix the other one regarding function extraction

src/CppUTest/JUnitTestOutput.cpp
vpod: done

PR #1268, opened 2019-12-10
Print error result and return 1 for zero run tests

### LINKED ISSUE #1210: With -p option, count of tests run is always 0
If a test suite is run with the -p option, the number of tests run is always zero:

`OK (4 tests, 0 ran, 0 checks, 0 ignored, 0 filtered out, 71 ms)`

That's because the run count is incremented in `UtestShell::runOneTestInCurrentProcess()`, which is run in the child's memory space and discarded on exit.

I moved the call to `result.countRun()` into `UtestShell::runOneTest()`, before the check is made for whether `-p` is set.

Happy to submit a pull request if you'd like.

### LINKED ISSUE #1211: With -p set, test cases that take unusual exit (sanitizers) are counted as success
Function `GccPlatformSpecificRunTestInASeperateProcess()` relies on tunnelling `result->getFailureCount()` into the child process and back out through the exit status code. However, a test case failure that takes its own `_exit()` (e.g., a Sanitizer error or other non-signal runtime error) will confuse the logic for counting the number of failures.

My fix is to change the parent/child logic to a simple zero/non-zero check: if the child process returns non-zero, then the test failed; if it returns 0, the test passed. (This means preserving the value of `result->getFailureCount()` across the call to `runOneTestInCurrentProcess()`.)

Happy to submit a pull request if you like.

### LINKED ISSUE #1262: Return error for zero run tests
As discussed in https://github.com/cpputest/cpputest/pull/1254 before, if there are no tests run, cpputest returns 0.
It seems that a lot of persons agree that "0 run" is usually not an expected result.
However, I can come up with situations when CI scripts run a pack of test binaries with some default filters and it is OK if some binary has no tests to run. Example out of my mind: we package all UTs to single executable and then run the executable on many test systems that may or may not have some library or some hardware instruction or co-processor (e.g. AVXnnn algorithm testing and we want to verify the algorithm, but skip in on non-enabled systems in a reliable way). Maybe not the best example, but I hope the point is clear.
We can also have a debate if 0 tests passed out of 0 run is 100% passed or 0% passed. :-)
As I've already faced few times the issue that CI is "green", but runs nothing due to bug in build scripts, I think, this should be addressed.
My proposal is to choose one of these options (I can implement it, but would like to know what to implement):
1. just assume that zero test run is an issue and return error, let caller fix itself to expect that (seems to be the most likely to cause issues);
2. add new parameter, return failure if zero tests run and the parameter _is not_ provided (may need fixing callers during cpputest update);
3. add new parameter, return failure if zero tests run and the parameter _is_ provided (whoever wants the feature has to enable it in the caller);
4. add build configuration flag to choose 2 or 3;
5. add build configuration flag to choose 1 or current behavior.

If I would have to decide, I'd choose 3 or 4 (possibly 3 for simplicity). It seems to be aligned to valgrind's `--error-exitcode=n` (do we want the code to be configurable?).

Could maintainers please give some guidance?

vpod:
- Print error result and return 1 for zero run tests (fixes #1262).
- Count tests run in separate process as 'run' (fixes #1210).
- Treat any non-zero return from child process as failure (fixes #1211).
- Fix waiting for test in separate process under debugger in GCC platform by avoiding checking uninitialized `status` variable (happens when `continue` checks loop condition)

### DIFF
--- a/include/CppUTest/TestResult.h
+++ b/include/CppUTest/TestResult.h
@@ -41,7 +41,7 @@ class UtestShell;
 class TestResult
 {
 public:
-    TestResult(TestOutput&);
+    explicit TestResult(TestOutput&);
     DEFAULT_COPY_CONSTRUCTOR(TestResult)
     virtual ~TestResult();
 
@@ -59,6 +59,8 @@ public:
     virtual void countIgnored();
     virtual void addFailure(const TestFailure& failure);
     virtual void print(const char* text);
+
+    virtual void clear();
 
     int getTestCount() const
     {
@@ -83,6 +85,11 @@ public:
     int getFailureCount() const
     {
         return failureCount_;
+    }
+
+    bool isSuccess() const
+    {
+        return getFailureCount() == 0 && getRunCount() > 0;
     }
 
     long getTotalExecutionTime() const;
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -113,8 +113,9 @@ int CommandLineTestRunner::runAllTests()
 {
     initializeTestRun();
     int loopCount = 0;
-    int failureCount = 0;
-    int repeat_ = arguments_->getRepeatCount();
+    int failedTestCount = 0;
+    int successCount = 0;
+    int repeatCount = arguments_->getRepeatCount();
 
     if (arguments_->isListingTestGroupNames())
     {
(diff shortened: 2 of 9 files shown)


vpod:
It seems that cpputest's own test fail when run via ctest:
```
68: Test command: /home/travis/build/cpputest/cpputest/cpputest_build/tests/CppUTestExt/CppUTestExtTests "-sg" "IEEE754ExceptionsPlugin2"
68: Test timeout computed to be: 1500
68: !
68: Errors (ran nothing, 510 tests, 0 ran, 0 checks, 1 ignored, 509 filtered out, 0 ms)
68: 
68/70 Test #68: CppUTestExtTests.IEEE754ExceptionsPlugin2 ....................................................***Failed    0.00 sec
```
@basvodde : so, do we need an extra command line flag to adjust this failure behavior?

basvodde:
Sorry to ask, but would you mind splittin this into smaller pull requests. Usually I actually take quite a while for long pull requests :(

vpod:
OK; I'll split gcc platform stuff and return code; and continue discussion of return code in the issue

PR #1269, opened 2019-12-11
Fix GCC platform-specific error cases for -p flag

### LINKED ISSUE #1210: With -p option, count of tests run is always 0

### LINKED ISSUE #1211: With -p set, test cases that take unusual exit (sanitizers) are counted as success

vpod:
- Treat any non-zero return from child process as failure (fixes #1211).
- Fix waiting for test in separate process under debugger by not checking uninitialized status variable (happens when continue checks loop condition).
- Count tests run in separate process as 'run' if fork() succeeds (fixes #1210).

### DIFF
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -115,8 +115,8 @@ static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, Test
                     return;
                 }
             } else {
-                if (WIFSTOPPED(status)) kill(w, SIGCONT);
                 SetTestFailureByStatusCode(shell, result, status);
+                if (WIFSTOPPED(status)) kill(w, SIGCONT);
             }
         } while ((w == syscallError) || (!WIFEXITED(status) && !WIFSIGNALED(status)));
     }
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -74,6 +74,19 @@ static int PlatformSpecificWaitPidImplementation(int, int*, int)
 
 #else
 
+static void SetTestFailureByStatusCode(UtestShell* shell, TestResult* result, int status)
+{
+    if (WIFEXITED(status) && WEXITSTATUS(status) != 0) {
+        result->addFailure(TestFailure(shell, "Failed in separate process"));
+    } else if (WIFSIGNALED(status)) {
+        SimpleString message("Failed in separate process - killed by signal ");
+        message += StringFrom(WTERMSIG(status));
+        result->addFailure(TestFailure(shell, message));
+    } else if (WIFSTOPPED(status)) {
+        result->addFailure(TestFailure(shell, "Stopped in separate process - continuing"));
+    }
+}
+
 static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, TestPlugin* plugin, TestResult* result)
 {
     const pid_t syscallError = -1;
@@ -93,27 +106,17 @@ static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, Test
         shell->runOneTestInCurrentProcess(plugin, *result);        // LCOV_EXCL_LINE
         _exit(initialFailureCount < result->getFailureCount());    // LCOV_EXCL_LINE
     } else {                    /* Code executed by parent */
-        result->countRun();
         do {
             w = PlatformSpecificWaitPid(cpid, &status, WUNTRACED);
             if (w == syscallError) {
-                if(EINTR == errno) continue; /* OS X debugger */
-                result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
-                return;
-            }
-
-            if (WIFEXITED(status) && WEXITSTATUS(status) != 0) {
-                result->addFailure(TestFailure(shell, "Failed in separate process"));
-            } else if (WIFSIGNALED(status)) {
-                SimpleString signal(StringFrom(WTERMSIG(status)));
-                {
-                    SimpleString message("Failed in separate process - killed by signal ");
-                    message += signal;
-                    result->addFailure(TestFailure(shell, message));
+                // OS X debugger causes EINTR
+                if (EINTR != errno) {
+                    result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
+                    return;
                 }
-            } else if (WIFSTOPPED(status)) {
-                result->addFailure(TestFailure(shell, "Stopped in separate process - continuing"));
-                kill(w, SIGCONT);
+            } else {
+                if (WIFSTOPPED(status)) kill(w, SIGCONT);
+                SetTestFailureByStatusCode(shell, result, status);
             }
         } while ((w == syscallError) || (!WIFEXITED(status) && !WIFSIGNALED(status)));
     }
--- a/src/Platforms/Gcc/UtestPlatform.cpp
+++ b/src/Platforms/Gcc/UtestPlatform.cpp
@@ -76,29 +76,33 @@ static int PlatformSpecificWaitPidImplementation(int, int*, int)
 
 static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, TestPlugin* plugin, TestResult* result)
 {
-    pid_t cpid, w;
+    const pid_t syscallError = -1;
+    pid_t cpid;
+    pid_t w;
     int status;
 
     cpid = PlatformSpecificFork();
 
-    if (cpid == -1) {
+    if (cpid == syscallError) {
         result->addFailure(TestFailure(shell, "Call to fork() failed"));
         return;
     }
 
     if (cpid == 0) {            /* Code executed by child */
-        shell->runOneTestInCurrentProcess(plugin, *result);   // LCOV_EXCL_LINE
-        _exit(result->getFailureCount());                     // LCOV_EXCL_LINE
+        const int initialFailureCount = result->getFailureCount(); // LCOV_EXCL_LINE
+        shell->runOneTestInCurrentProcess(plugin, *result);        // LCOV_EXCL_LINE
+        _exit(initialFailureCount < result->getFailureCount());    // LCOV_EXCL_LINE
     } else {                    /* Code executed by parent */
+        result->countRun();
         do {
             w = PlatformSpecificWaitPid(cpid, &status, WUNTRACED);
-            if (w == -1) {
-                if(EINTR ==errno) continue; /* OS X debugger */
+            if (w == syscallError) {
+                if(EINTR == errno) continue; /* OS X debugger */
                 result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -76,29 +76,33 @@ static int PlatformSpecificWaitPidImplementation(int, int*, int)
...
     }
 
     if (cpid == 0) {            /* Code executed by child */
-        shell->runOneTestInCurrentProcess(plugin, *result);   // LCOV_EXCL_LINE
-        _exit(result->getFailureCount());                     // LCOV_EXCL_LINE
+        const int initialFailureCount = result->getFailureCount(); // LCOV_EXCL_LINE
+        shell->runOneTestInCurrentProcess(plugin, *result);        // LCOV_EXCL_LINE
+        _exit(initialFailureCount < result->getFailureCount());    // LCOV_EXCL_LINE
     } else {                    /* Code executed by parent */
+        result->countRun();
```
vpod: Here I assume that the test is "run" only if `fork()` returns success to parent (as child is not run).
Alternatively, we can count in caller function like proposed in #1210. Or we can count in line 104 below, i.e. if `waitpid()` also passes.
I'm open if this needs to be changed.

vpod:
Travis CI failure: I would prefer if default cmake flags match CI flags (in this case -DWERROR=ON) as default cmake with clang 9.0 did not fail for me.

src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -111,7 +115,7 @@ static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, Test
                 result->addFailure(TestFailure(shell, "Stopped in separate process - continuing"));
                 kill(w, SIGCONT);
             }
-        } while (!WIFEXITED(status) && !WIFSIGNALED(status));
+        } while ((w == syscallError) || (!WIFEXITED(status) && !WIFSIGNALED(status)));
```
vpod: This prevents reading uninitialized `status` if `continue` from line 100 happens (I've also updated the test to validate that he loop does not exit after the 1st iteration as it did before randomly)

tests/CppUTest/UtestPlatformTest.cpp
```diff
@@ -56,8 +58,17 @@ static void _failFunction()
...
+static void _exitNonZeroFunction() __no_return__;
+static void _exitNonZeroFunction()
+{
+    exit(1);
+}
+
 #include <errno.h>
 
+static int waitpid_while_debugging_stub_number_called = 0;
+const int waitpid_while_debugging_stub_forced_failures = 10;
```
vpod: these are needed in test CallToWaitPidWhileDebuggingInSeparateProcessWorks

src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -76,29 +76,33 @@ static int PlatformSpecificWaitPidImplementation(int, int*, int)
...
-    pid_t cpid, w;
+    const pid_t syscallError = -1;
+    pid_t cpid;
+    pid_t w;
     int status;
 
     cpid = PlatformSpecificFork();
 
-    if (cpid == -1) {
+    if (cpid == syscallError) {
```
basvodde: Could you try not to make changes that do not change anything and are probably a matter of style? Or at least put these changes in a separate pull requests so we can discuss them there. I personally do not think the syscallError variable adds any benefit or makes it clearer. Similarly, the splitting of cpid and w in two lines makes no change.

src/Platforms/Gcc/UtestPlatform.cpp
basvodde: Seems that proposal would be better because then there is just one count, right?

Any trade-offs to be made here?

src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -76,29 +76,33 @@ static int PlatformSpecificWaitPidImplementation(int, int*, int)
...
+                if(EINTR == errno) continue; /* OS X debugger */
                 result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
                 return;
             }
 
-            if (WIFEXITED(status) && WEXITSTATUS(status) > result->getFailureCount()) {
+            if (WIFEXITED(status) && WEXITSTATUS(status) != 0) {
                 result->addFailure(TestFailure(shell, "Failed in separate process"));
             } else if (WIFSIGNALED(status)) {
                 SimpleString signal(StringFrom(WTERMSIG(status)));
```
basvodde: Do you have any idea why this signal code here (not shown in review) is so weird?

src/Platforms/Gcc/UtestPlatform.cpp
basvodde: Confusing. Could you remove the continue by reverting the if on line 100 and changing the next if to an else if. That ought to make the logic a bit clearer. Having continue in a do-while loop feels bad.

Perhaps also extract the status logic of creating the right failure message in a separate function.

src/Platforms/Gcc/UtestPlatform.cpp
vpod: I think, it is so. If we consider that "attempt to run"is "run" (i.e. failure to fort means that the test is not actually even started), then having it in one place is better.
I'll do it in one place then (unless I find strange things in "run one test" function

src/Platforms/Gcc/UtestPlatform.cpp
vpod: I added syscallError to not use the same magic number `-1` yet another time and aligned whole function to be consistent.
If I'm the only one that finds it not very trivial knowledge that -1 is the error and other value is not, I'll roll back to -1 usage as it was before me.

src/Platforms/Gcc/UtestPlatform.cpp
vpod: No. But you proposed yourself to avoid touching code that I don't have to modify...

src/Platforms/Gcc/UtestPlatform.cpp
basvodde: Ack

src/Platforms/Gcc/UtestPlatform.cpp
basvodde: Its ok. No need to roll back. If the error code was 6384 then creating a variable for that makes sense. However, -1 is such a common error code number and it is so unlikely to change that creating a variable for that is mostly a matter of taste.

I'm good with making code improvements (promote it even) but they shouldn't be mostly a matter of taste. E.g. extracting the signal logic to a new function IMHO would be clearly an improvement. Still, having them in separate PR is preferred as I'll need to understand whether they are fixing the issue or whether they are just code improvements.

src/Platforms/Gcc/UtestPlatform.cpp
basvodde: Hah :) Right! Thanks for that :P

src/Platforms/Gcc/UtestPlatform.cpp
vpod: After refactoring, it became evident to me that `WIFSTOPPED()` case looks weird:
1. it is not listed in while() -> seems logical as we resume the process via `SIGCONT`;
2. but there is a `TestFailure` added in such case - why? if we resume the process and wait, it means it did not fail...
As this goes beyond my understanding and debugging capabilities, I propose to create an issue for this scenario. Someone may address it if they face it.

src/Platforms/Gcc/UtestPlatform.cpp
vpod: done

src/Platforms/Gcc/UtestPlatform.cpp
vpod: Done (I kept `waitpid()` failure in the same function to reduce code movement - it seems to me to be ok, but I can try to move it as well)

src/Platforms/Gcc/UtestPlatform.cpp
vpod: as I moved it to new function, I merged few lines to make it look better (in my opinion)

src/Platforms/Gcc/UtestPlatform.cpp
```diff
@@ -93,27 +106,17 @@ static void GccPlatformSpecificRunTestInASeperateProcess(UtestShell* shell, Test
...
+                // OS X debugger causes EINTR
+                if (EINTR != errno) {
+                    result->addFailure(TestFailure(shell, "Call to waitpid() failed"));
+                    return;
                 }
-            } else if (WIFSTOPPED(status)) {
-                result->addFailure(TestFailure(shell, "Stopped in separate process - continuing"));
-                kill(w, SIGCONT);
+            } else {
+                if (WIFSTOPPED(status)) kill(w, SIGCONT);
```
basvodde: At least this one ought to be below the setTestFailure, right? The Set TestFailure will still add a failure on STOPPED.

tests/CppUTest/UtestPlatformTest.cpp
```diff
@@ -25,6 +25,8 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <stdlib.h>
```
basvodde: I missed out this one. Can you avoid including stdlib directory and instead use the wrapped header for that.

src/Platforms/Gcc/UtestPlatform.cpp
vpod: Probably no difference, but I'll put it after to be consistent with original code

tests/CppUTest/UtestPlatformTest.cpp
vpod: what about lines 67, 101, 102?

vpod: Done, but I have no idea how to wrap exit() call in such case

vpod: Note that stdlib.h is directly included in `UtestPlatform.cpp` file. Probably whatever is inside `CPPUTEST_HAVE_FORK` later in this file can assume presence of `stdlib.h`, so it (`exit()`) is not an issue.

basvodde:
Thanks!

PR #1272, opened 2019-12-24
Return failure if no tests are run or ignored

### LINKED ISSUE #1262: Return error for zero run tests

vpod:
* Add helper function to `TestResult`
* Make `TestOutput` report error based on it
* Make `runAllTests()` return non-zero based on it

Fixes #1262

### DIFF
--- a/src/CppUTest/CommandLineTestRunner.cpp
+++ b/src/CppUTest/CommandLineTestRunner.cpp
@@ -113,8 +113,9 @@ int CommandLineTestRunner::runAllTests()
 {
     initializeTestRun();
     int loopCount = 0;
-    int failureCount = 0;
-    int repeat_ = arguments_->getRepeatCount();
+    int failedTestCount = 0;
+    int failedExecutionCount = 0;
+    int repeatCount = arguments_->getRepeatCount();
 
     if (arguments_->isListingTestGroupNames())
     {
@@ -139,17 +140,20 @@ int CommandLineTestRunner::runAllTests()
         output_->print("\n");
         srand(seed);
     }
-    while (loopCount++ < repeat_) {
+    while (loopCount++ < repeatCount) {
         if (shuffleEnabled)
         {
             registry_->shuffleRunOrder(rand_);
         }
-        output_->printTestRun(loopCount, repeat_);
+        output_->printTestRun(loopCount, repeatCount);
         TestResult tr(*output_);
         registry_->runAllTests(tr);
-        failureCount += tr.getFailureCount();
+        failedTestCount += tr.getFailureCount();
+        if (tr.isFailure()) {
+            failedExecutionCount++;
+        }
     }
-    return failureCount;
+    return failedTestCount != 0 ? failedTestCount : failedExecutionCount;
 }
 
 TestOutput* CommandLineTestRunner::createTeamCityOutput()
--- a/src/CppUTest/TestOutput.cpp
+++ b/src/CppUTest/TestOutput.cpp
@@ -146,12 +146,12 @@ void TestOutput::printTestsEnded(const TestResult& result)
 {
     print("\n");
     const bool isFailure = result.isFailure();
+    const int failureCount = result.getFailureCount();
     if (isFailure) {
         if (color_) {
             print("\033[31;1m");
         }
         print("Errors (");
-        const int failureCount = result.getFailureCount();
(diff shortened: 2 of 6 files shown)


src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -113,8 +113,9 @@ int CommandLineTestRunner::runAllTests()
 {
     initializeTestRun();
     int loopCount = 0;
-    int failureCount = 0;
-    int repeat_ = arguments_->getRepeatCount();
+    int failedTestCount = 0;
+    int failedExecutionCount = 0;
+    int repeatCount = arguments_->getRepeatCount();
```
vpod: This renaming from `repeat_` can be considered style preference, but aligning naming makes sense in my opinion. Also in this class names ending with underscore `_` are member variables, while local variables do not end with underscores.
If desired, I can roll back, but IMO the result will look odd with 3 variables using one convention and one using different convention.

vpod:
If (when?) new cpputest release is created, this change should be listed as significant/ breaking, I believe.

src/CppUTest/CommandLineTestRunner.cpp
basvodde: Makes sense, thanks!

src/CppUTest/TestOutput.cpp
```diff
@@ -145,14 +145,20 @@ void TestOutput::printCurrentGroupEnded(const TestResult& /*res*/)
...
         print("Errors (");
-        print(result.getFailureCount());
-        print(" failures, ");
+        const int failureCount = result.getFailureCount();
+        if (failureCount > 0) {
+            print(failureCount);
+            print(" failures, ");
+        }
+        else {
+            print("ran nothing, ");
```
basvodde: Can we make the error much more descriptive for the failure because no tests run? I think this might be a surprise to people and I think the error ought to explain that. Perhaps:

"Test run failed because no tests were run or ignored. Assuming something went wrong"

vpod: I think, better to put this as a note after the statistic to keep it readable. Maybe even new line.
Or do you want it to be printed right next to `Errors` before the stats?

basvodde: A new line will be fine. At least, the reason for failing needs to be very clear as it will come as a surprise to people.

vpod: Done

basvodde:
Good and clear!

vpod:
DOS build failed with timeout: `The job exceeded the maximum time limit for jobs, and has been terminated.`
No idea what's wrong, but seems all was passing before it got aborted.

basvodde:
Yeah, the dos build is because it needs to download something. We need to find a way to fix that.

PR #1276, opened 2020-02-07
add .withOutputParameter to expected call

jacob-keller:
When writing mock expectations, a mock implementation may have an output
parameter. In some tests it may be that the output parameter remains
unmodified.

To write an expectation which has an output parameter that is unmodified
you can use the following code:

    .withOutputParameterReturning("name", NULL, 0)

This is somewhat confusing, so add a shorthand

    .withOutputParameter("name")

This sets up the mock function to expect the "name" output parameter but
leave it unmodified.

### DIFF
--- a/tests/CppUTestExt/MockParameterTest.cpp
+++ b/tests/CppUTestExt/MockParameterTest.cpp
@@ -634,6 +634,17 @@ TEST(MockParameterTest, outputParameterSucceeds)
     mock().checkExpectations();
 }
 
+TEST(MockParameterTest, unmodifiedOutputParameterSucceeds)
+{
+    int param = 1;
+
+    mock().expectOneCall("function").withUnmodifiedOutputParameter("parameterName");
+    mock().actualCall("function").withOutputParameter("parameterName", &param);
+
+    CHECK_EQUAL(param, 1);
+    mock().checkExpectations();
+}
+
 TEST(MockParameterTest, noActualCallForOutputParameter)
 {
     MockFailureReporterInstaller failureReporterInstaller;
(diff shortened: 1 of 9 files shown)


jacob-keller:
Perhaps a better method might be something like ".ignoreParameter" or ".ignoreOutputParameter", which is something else that could be useful for tests which want to ignore a specific parameter but catch the addition of new parameters in the future (vs using .ignoreOtherParameters).

That looks more difficult to implement so I didn't bother.

basvodde:
Could it at least be called "withUnmodifiedOutputParameter or something like that?

jacob-keller:
I like that. I'm also thinking of the ability to expect a regular parameter but with an unknown value as well, so I will look into adding that and updating the test suite.

basvodde:
Will you change the name still?

jacob-keller:
Yes, i am planning on updating this pull request. I didn't manage to get to
it today.

jacob-keller:
I renamed everything to be .withUnmodifiedOutputParameter() and added tests.

The type is still "const void *". I considered adding a custom type like "<none>" or "<empty>" to represent this, but it doesn't seem to gain us much.

I also thought about going the route of adding ".ignoreParameter(name)" but that seems difficult to get right.


For implementing something like ".expectParameterWithAnyValue(name)", I think we would have to add a custom type that represents that fact.

jacob-keller:
Added a test case for the _c implementation.

tests/CppUTestExt/MockParameterTest.cpp
```diff
@@ -634,6 +634,20 @@ TEST(MockParameterTest, outputParameterSucceeds)
     mock().checkExpectations();
 }
 
+TEST(MockParameterTest, unmodifiedOutputParameterSucceeds)
+{
```
basvodde: This test is copy-pasted. The retval variable doesn't make a lot of sense in this test. Could you please clean this up and also check the rest of the code that this didn't happen?

Also, remove comments please :)

basvodde:
Ok, I was thinking of unwanted side-effects of this, but I cannot find any. So, it'll be good like this. Thanks!

jacob-keller:
Hmm seems like something is still not working right.

jacob-keller:
Looks like only a DOS build is failing with some sort of allocation issue..

basvodde:
Yeah, the DOS build needs to be fixed or disabled. It is on the todo list. Now it just causes confusion.

PR #1280, opened 2020-03-10
Use a safer method to toggle new/delete overloads.

### LINKED ISSUE #1278: Threadsafe memory leak detector turns itself off
#439 introduced an alternative threadsafe memory allocator, but there's a sneaky little bug in it.

https://github.com/cpputest/cpputest/blob/8f4aa13d3618a7a9fb6c864211a3d9fb7c32b04c/src/CppUTest/MemoryLeakWarningPlugin.cpp#L485-L495

In this code, which is called the first time that a memory allocation is performed with malloc/new, it turns off the memory allocator to allocate the allocator itself, then turns it back on.

The problem is that this only supports the two states "off" and "on".  If you had happened to call `MemoryLeakWarningPlugin::turnOnThreadSafeNewDeleteOverloads();` prior to any other memory allocations, this results in turning the leak detector off entirely.  (Because `areNewDeleteOverloaded()` returns false -- and even if it did return true, this would restore it to the non-thread-safe version, which would be even more of a surprise.)

This is often hidden in "real" apps due to the presence of global and static variables that end up triggering a memory allocation before `main()` runs, so that the plugin is still in the "on" state and it gets correctly restored, then later changed to threadsafe mode by the user code.  So it can be worked around by triggering some kind of memory allocation before calling `turnOnThreadSafeNewDeleteOverloads()` -- but it's still a bit of a trap that ought to be fixed.

Also: telling people to call this should probably be in the docs somewhere.  Or it should just be made the default, perhaps with the non-thread-safe version as an option.

uecasm:
Fixes #1278.

### DIFF
--- a/include/CppUTest/MemoryLeakWarningPlugin.h
+++ b/include/CppUTest/MemoryLeakWarningPlugin.h
@@ -67,7 +67,10 @@ public:
     static void turnOffNewDeleteOverloads();
     static void turnOnNewDeleteOverloads();
     static void turnOnThreadSafeNewDeleteOverloads();
-    static bool areNewDeleteOverloaded();
+
+    static void saveAndDisableNewDeleteOverloads();
+    static void restoreNewDeleteOverloads();
+
 private:
     MemoryLeakDetector* memLeakDetector_;
     bool ignoreAllWarnings_;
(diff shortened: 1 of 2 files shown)


basvodde:
Build failure seems related to the PR

PR #1284, opened 2020-03-14
Fix remaining missing class member function override specifiers

Xiche:
Tired of these warnings showing up when I compile my code. :smiley:


basvodde:
Thanks!

PR #1285, opened 2020-03-15
Added installation as external cmake project to README

bneumann:
Because I used this regularly and always forgot how to do it right. And also because googletest has something similar ;)


basvodde:
Thanks!

PR #1310, opened 2020-04-20
Compiler version in configure script

basvodde:
Report the compiler version. For some reason, the CXX doesn't seem to work. Not sure why.


basvodde:
Ah, accidentally left the docker files in this pull request. Well, so be it.

PR #1320, opened 2020-04-21
C++17 ci builds with latest gcc and clang

offa:
Builds the two C++17 linux builds with latest gcc and clang versions, providing feedback for the most recent compiler / standard versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,30 @@ jobs:
         apt:
           packages:
             - libc++-dev
+      env: BUILD=cmake CPP_STD=17
+      addons:
+        apt:
+          packages:
+            - libc++-dev
+
+    - compiler: gcc
+      env: BUILD=cmake CPP_STD=98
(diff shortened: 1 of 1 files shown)


offa:
Travis failed due to some dos issue (not related to this PR).

basvodde:
I'd prefer to have a different target for that, not the default one.

basvodde:
And, is there any way to extract the latest compiler from the default? That would be nice.

offa:
> I'd prefer to have a different target for that, not the default one.

You mean build? No problem, I'll update the PR accordingly.

> And, is there any way to extract the latest compiler from the default? That would be nice.

I don't think so. The default compiler is selected by ubuntu and typically not updated within a distribution version. Except the default, all compilers need to be installed with a version, eg. gcc-8 or clang-9.

basvodde:
Included this in the travis build. But did manually as the file had changed so much.

PR #1350, opened 2020-04-26
Add examples to cmake

jakubmiernik:
This add option: EXAMPLES to cmake to be able build examples by cmake.

In build dir:
```
cmake -DEXAMPLES=ON ..
make
```

### DIFF
--- a/examples/AllTests/CMakeLists.txt
+++ b/examples/AllTests/CMakeLists.txt
@@ -1,0 +1,17 @@
+enable_testing()
+
+include_directories(../ApplicationLib)
+
+add_executable(ExampleTests
+    AllTests.cpp
+    CircularBufferTest.cpp
+    EventDispatcherTest.cpp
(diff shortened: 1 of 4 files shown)


examples/AllTests/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+enable_testing()
```
offa, line 1: This should typically be called in the projects root path. According to the [cmake docs](https://cmake.org/cmake/help/latest/command/enable_testing.html) it's not necessary if the CTest package is included (it's in the root CMakeLists.txt).

offa: Interestingly changing this for the tests breaks ctest :confused:

offa: I have opened a PR #1351 to fix this.

PR #1351, opened 2020-04-26
enable_testing() moved to project source root directory

offa:
Moves `enable_testing()` to the project source root directory as mentioned in the [docs](https://cmake.org/cmake/help/latest/command/enable_testing.html).

Starting from v3.15 the documentation explicitly mentions that it's _"automatically invoked, when CTest module is included"_. Earlier version may do that too, but since the minimum version is v3.1 I kept the `enable_testing()` call, just be be safe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,8 @@ include("${CppUTestRootDirectory}/cmake/Modules/CppUTestBuildTimeDiscoverTests.c
 include("${CppUTestRootDirectory}/cmake/Modules/CppUTestNormalizeTestOutputLocation.cmake")
 include(GNUInstallDirs)
 
+enable_testing()
+
 configure_file (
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CppUTestRootDirectory}/cmake/Module
 
 include("${CppUTestRootDirectory}/cmake/Modules/CppUTestConfigurationOptions.cmake")
 include(CTest)
+enable_testing()
```
jakubmiernik, line 78: Why exactly in this line?
I mean it's between includes, is there any reason for this?

offa: Good catch! My intention was to keep it close to the related include, but yeah, this totally clutters the section.

PR #1360, opened 2020-04-28
Compiler extensions disabled

offa:
Compiler extensions disabled on the CMake build (see [docs](https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html)). This ensures the build pure standard and prevents eg. the usage of gnu extensions.

@basvodde what's your opinion on unconditionally set this option? Or alternatively, set it to `OFF` unless it's already set to `ON` (eg. option explicitly passed).


basvodde:
Sounds fine, thanks.

PR #1387, opened 2020-05-10
Missing SimpleString.h fixed

offa:
Fixes the missing `SimpleString.h` header file error, when compiling tests with CMake.


basvodde:
Oh, I must have accidentally deleted it, Thanks for catching that.

Issue #1144, opened 2018-03-07
CMake Cross-Compile Fails


jovere:
When using CMake to compile CppUTest, it automatically attempts to run the test for verification.  This is great.

However, it's not great when you're using CMake/CppUTest with a cross compiler.  If the executable is for a different architecture, it still tries to run, fails, and then proceeds to delete the executable.  It appears the issue is specifically in `CppUTestBuildTimeDiscoverTests.cmake` where `execute_process` is called.  

Is there some kind of flag to disable the compile time execution?

offa:
Is disabling the tests an alternative, you still want to build the them? There's currently only a option to disable tests completely; adding one to disable the discover shouldn't be a problem though.

basvodde:
@offa Close or add flag?

offa:
I can add an option, no problem.

basvodde:
Thanks. Merged and close this issue.

PR #1390, opened 2020-05-11
Option for build discover of tests

offa:
Adds an option `TESTS_BUILD_DISCOVER` (default: `ON`) to disable build time discover of tests, see #1144.


basvodde:
Thanks

Issue #1139, opened 2018-02-21
How to generate coverage using CMake?


tehKaiN:
I've read "How to create a coverage report" paragraph on main page, but still don't know how to generate `gcov_report.txt` using CMake. In cpputest_build I run:

```
cmake .. -G "MinGW Makefiles" -DCOVERAGE=ON
make
cd tests/cpputest
cpputesttests

.............................................!....
..................................................
..................................................
..................................................
..................................................
...................!..............................
..................................................
..................................................
..................................................
...................................!....!.!......!
.....!..!..!..!..!..!..!..!..!..!..!..!...!..!....
...!..!...!..!..!..!..!..!..!..!..!..!..!..!..!...
.....!..!..!..!..!..!..!..!..!..!.................
.........!........!................
OK (685 tests, 638 ran, 1192 checks, 47 ignored, 0 filtered out, 188 ms)
```

But gcov coverage report doesn't appear anywhere. Also I don't see any specific cmake-generated makefile goal which would generate it. I have gcov installed and it's in my PATH. I'm using win10, tdm-gcc and CMake 3.9.2. What am I missing?

basvodde:
@offa  ?

offa:
As far as I know, the option adds only `--coverage` to C/C++ flags. We can add targets for report generation indeed.

offa:
lcov has become a real pain with the recent versions . Can't we switch to [gcovr](https://github.com/gcovr/gcovr) which provides the same features but is generally more usable?

offa:
@basvodde I guess you can close here.

PR #1394, opened 2020-05-13
CMake coverage report generation

offa:
As mentioned in #1139.

@basvodde I used [gcovr](https://github.com/gcovr/gcovr) instead of lcov as it's much easier to install and use. It can create coverage results, reports in various formats etc. in a single pass and provides support for configuration file. If the the lcov is still set, can migrate it back.

Should I change the make target to `check_coverage` as used in `Makefile.am`? Would keep it consistent.


offa:
Travis was unable to clone the repository!? :confused:

basvodde:
I might have overloaded it with my experimenting :)

I restarted the job.

offa:
Thanks, unfortunately git clone failed again.

basvodde:
Changing the Makefile.am is ok. Leaving it is also ok... at some point it will likely change then.

offa:
Great, thanks for merging. Feel free to rename the target, no preference from my side.

Btw. gcovr is simple enough so everyone can call it directly, no matter what build system is used.

PR #1411, opened 2020-07-24
Types pid_t and int are not always the same.

oicontrib:
We get build error when pid_t is not equal to int.
e.g on solaris when building 32bit version of cpputest pid_t is typedef to long.


include/CppUTest/CppUTestConfig.h
```diff
@@ -32,6 +32,12 @@
 #include "CppUTestGeneratedConfig.h"
 #endif
 
+#ifdef _cpputest_pid_t
+#define PID_T _cpputest_pid_t
```
offa, line 36: Isn't a typedef a better choice here? `#define` can be harmful at many places.

oicontrib: Changed #define to typedef.

PR #1422, opened 2020-09-16
Add crash-on-fail mode to TestTerminator base class

CiderMan:
Split out the core crash-on-fail functionality into a PR (including some refactoring of how it is implemented in the TestTerminator classes)

### DIFF
--- a/include/CppUTest/Utest.h
+++ b/include/CppUTest/Utest.h
@@ -61,30 +61,33 @@ public:
 
 class TestTerminator
 {
-protected:
-    static bool crashOnFail_;
-    virtual void coreExitCurrentTest() const=0;
 public:
-    static void setCrashOnFail(bool crashOnFail = true);
-
-    virtual void exitCurrentTest() const;
+    virtual void exitCurrentTest() const=0;
     virtual ~TestTerminator();
 };
 
 class NormalTestTerminator : public TestTerminator
 {
-protected:
-    virtual void coreExitCurrentTest() const _override;
 public:
+    virtual void exitCurrentTest() const _override;
     virtual ~NormalTestTerminator() _destructor_override;
 };
 
 class TestTerminatorWithoutExceptions  : public TestTerminator
 {
-protected:
-    virtual void coreExitCurrentTest() const _override;
 public:
+    virtual void exitCurrentTest() const _override;
     virtual ~TestTerminatorWithoutExceptions() _destructor_override;
+};
+
+class CrashingTestTerminator : public TestTerminator
+{
+protected:
+    const TestTerminator &coreTestTerminator_;
+public:
+    explicit CrashingTestTerminator(const TestTerminator &coreTestTerminator);
+    virtual void exitCurrentTest() const _override;
+    virtual ~CrashingTestTerminator() _destructor_override;
 };
 
 //////////////////// UtestShell
@@ -93,6 +96,11 @@ class UtestShell
 {
 public:
     static UtestShell *getCurrent();
+
+    static const TestTerminator &getDefaultTestTerminator();
+    static const TestTerminator &getDefaultTestTerminatorWithoutExceptions();
+
+    static void setCrashOnFail(bool crashOnFail = true);
 
 public:
     UtestShell(const char* groupName, const char* testName, const char* fileName, size_t lineNumber);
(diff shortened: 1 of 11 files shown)


basvodde:
Why not make a CrashTestTerminator ?

CiderMan:
> 
> 
> Why not make a CrashTestTerminator ?

I could - though I would also have a to make a CrashTestTerminatorWithoutExceptions. Since, in general, we would need a CrashTestTerminator class for every non-crashing version, I felt that the switch between crashing and non-crashing sat better in the base class that all other terminators derive from. That avoids the need to duplicate code between the different derived, crashing classes.

If you are sure that you want to go this way, I can make the change.

basvodde:
To me, my initial thinking is that the crashing behavior seems a type of TestTerminator... not a behavior that is part of all the Terminators...

basvodde:
Ok, a bit better but still too much. Could you:

* Add one "currentTestTerminator" method to the UTestShell
* Remove the defaultTestTerminator methods
* Remove the choice to not throw exceptions
* Not change the C part. For now, it is ok that the crashOnFailure doesn't crash in C files. We can always look at that later
* Not have a "coreTerminator" in the crash on failure. Just crash.
* Remove the defaultTerminator from MemoryLeakPlugin and keep the original. The rare places where it doesn't use the default terminator are places that you do not want to change probably. E.g. this one is related to a destructor call and a potentially very nasty bug.
* Remove any change related to TestRegistry (for now).
* Remove the boolean in UTestShell related to crashing. You do not need a boolean and bools are best prevented completely.

CiderMan:
OK, I have (I hope) made the changes that you have requested.

include/CppUTestExt/GTestConvertor.h
```diff
@@ -199,6 +199,8 @@ class GTestDummyResultReporter : public ::testing::ScopedFakeTestPartResultRepor
 
 class GMockTestTerminator : public TestTerminator
 {
+protected:
+    virtual void coreExitCurrentTest() {};
```
CiderMan, line 203: Oops, this can go now...

basvodde: Reviewing your own code in the PR :)

src/CppUTest/Utest.cpp
```diff
@@ -129,6 +129,14 @@ extern "C" {
 
 }
 
+/************************************************************************** */
+/* Define the two possible TestTerminators that can be selected at run time */
```
basvodde, line 133: Can you remove the comments. We consider comments to be mostly a smell, especially when they explain relative obvious things.

src/CppUTest/Utest.cpp
```diff
@@ -676,6 +696,16 @@ TestTerminatorWithoutExceptions::~TestTerminatorWithoutExceptions()
 {
 }
 
+void CrashingTestTerminator::exitCurrentTest() const
+{
+    UtestShell::crash();
+    NormalTestTerminator::exitCurrentTest();
```
basvodde, line 702: Why do we still want to do the exitCurrentTest?

src/CppUTestExt/MockFailure.cpp
```diff
@@ -43,7 +43,7 @@ class MockFailureReporterTestTerminator : public NormalTestTerminator
         if (crashOnFailure_)
```
basvodde, line 43: I thought we were going to do the mock after this PR. If not, then this probably need to change as this functionality is not moved to CppUMock.

basvodde:
Tests for the new functionality seem to be missing? I would suggest to test-first, it will potentially improve your style significantly (and less need for debuggers)

src/CppUTest/Utest.cpp
CiderMan: This is needed in part due to the way that we test the crashing functionality. We do the same thing in the Mock TestTerminator class (albeit the crash is conditional based on the value of a Boolean in that case.)
Also, since the crash function can be changed, better to fail the test if the crash function returns than go back to the test and keep running it...

src/CppUTestExt/MockFailure.cpp
CiderMan: We are. This change (to call the current TestTerminator, rather than a hard coded one) is to support case #3 - when the command-line option is provided, all failures will cause a crash.

CiderMan:
> 
> 
> Tests for the new functionality seem to be missing? I would suggest to test-first, it will potentially improve your style significantly (and less need for debuggers)

I did test-drive to an extent - from the CppUTestExt test as it already had tests of scenarios 1 and 2 so I added scenario 3 and wrote the code from there. I can certainly add more tests (in the CppUTest tests) if you are happy with where the change-set is...

basvodde:
Well... what do you think of this compared to the original PRs ?

src/CppUTest/Utest.cpp
```diff
@@ -575,6 +583,18 @@ UtestShell* UtestShell::getCurrent()
...
+    return *currentTestTerminator_;
+}
+
+void UtestShell::setCrashOnFail(bool crashOnFail)
+{
+    if (crashOnFail)
+        currentTestTerminator_ = &crashingTestTerminator;
+    else
+        currentTestTerminator_ = &normalTestTerminator;
+}
```
basvodde, line 597: Can we split this into two functions (set and restoreDefault) so that we don't have this boolean and if.

src/CppUTest/Utest.cpp
basvodde: Ok, lets keep that then for now.

src/CppUTestExt/MockFailure.cpp
basvodde: Ok, but if we do the mock in a separate PR, then I suggest we do that.

If we do the mock in this PR, then we need to remove the crashOnFailure_ boolean out of the mock and instead use the CppUTest one. Now it is just a half done implementation.

CiderMan: I do not agree that it is half done. This implements precisely what was agreed for the initial feature, in terms of the 3 scenarios. It does mean that we have two related mechanisms in mock that we should look at harmonizing later but this change is the simplest change that will cause the 3 tests for the 3 scenarios to pass.

CiderMan:
> 
> 
> Well... what do you think of this compared to the original PRs ?

I'm a big fan of Occam's Razor so that fact that we now have a change-set for part of the overall feature bigger than PR for the entire feature when the crashing was put into the base-class is, for me, a bad sign. I still feel that the best design was for all TestTerminators to support crashing (i.e. a base class feature) rather than having various derived specializations to add crashing on a case-by-case basis.
However, I respect your experience and oversight on CppUTest so I'm happy to implement the design that you feel fits best.

CiderMan:
Is there anything else that is needed for this change-set to be accepted?

src/CppUTestExt/MockFailure.cpp
basvodde: Sure... but it adds functionality duplication. I would prefer to either not add this functionality or do it proper. Leaving code like this makes it easy to not do the cleaning up... which leads to hard to maintain legacy code.

basvodde:
> > Well... what do you think of this compared to the original PRs ?
> 
> I'm a big fan of Occam's Razor so that fact that we now have a change-set for part of the overall feature bigger than PR for the entire feature when the crashing was put into the base-class is, for me, a bad sign. I still feel that the best design was for all TestTerminators to support crashing (i.e. a base class feature) rather than having various derived specializations to add crashing on a case-by-case basis.
> However, I respect your experience and oversight on CppUTest so I'm happy to implement the design that you feel fits best.

What would be the benefit of having crashing behavior in all of the TestTerminator? Isn't crashing a specific Terminator behavior?

If you look at the current PR code-wise, the code is significantly cleaner. There are no booleans, less code, and less changes to existing classes. In other words, to me, it seems to much better fulfil good design principles than the original PR.

basvodde:
> Is there anything else that is needed for this change-set to be accepted?

So far so good. I think there are still 2 open issues:

* clean up the Mock so that we do not have duplicate behavior.
* Add a test to CppUTest itself specifically for the crashing behavior. It seems to be missing...

CiderMan:
OK, I'll add the appropriate CppUTest tests for this. Are you happy for the Mock clean up to be dealt with in a separate PR?

basvodde:
> OK, I'll add the appropriate CppUTest tests for this. Are you happy for the Mock clean up to be dealt with in a separate PR?

As long as it is really done and not postponed :)

CiderMan:
> > OK, I'll add the appropriate CppUTest tests for this. Are you happy for the Mock clean up to be dealt with in a separate PR?
> 
> As long as it is really done and not postponed :)

He he! Yes, it will (once I understand exactly what you want.) I have the command-line change pretty much done from my first PR so I will create that PR once this one is delivered and do the Mock work in parallel with that review.

tests/CppUTest/UtestTest.cpp
```diff
@@ -120,6 +120,42 @@ TEST(UtestShell, ExitLeavesQuietly)
...
+
+    fixture.setTestFunction(_failMethod);
+    fixture.runAllTests();
+
+    CHECK(cpputestHasCrashed);
+
+    UtestShell::resetFailBehaviour();
+    UtestShell::resetCrashMethod();
+}
+
```
basvodde, line 158: Please don't copy-paste, challenge every line, and write tight code.

The "resetFailBehavior" in the first test seems not needed. The nonCrashing test will need to check the test fails (as you kept the NormalTermintor there)

Last request before I merge this. Can you rename "resetFailBehavior" to "restoreDefaulTestTerminator" or something like that. Reason: For symmetry, a reset ought to also have a set (like the crashMethod), but the method isn't called "resetCrashOnFail" and therefore it is probably to not call it reset at all.

basvodde:
Thanks, almost there for this.

include/CppUTest/Utest.h
```diff
@@ -97,7 +97,7 @@ class UtestShell
     static const TestTerminator &getCurrentTestTerminator();
 
     static void setCrashOnFail();
-    static void resetFailBehaviour();
+    static void resetDefaultTestTerminator();
```
basvodde: set... set ... set :)

basvodde:
Ok, lemme wait for the build to pass.

basvodde:
Thanks for your patience.

PR #1423, opened 2020-09-16
Add CHECK_EQUAL_ZERO macros

rnestler:
These are useful when working with error code return values, where one
often compares to zero which means no error has happened.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -142,6 +142,10 @@
     UtestShell::getCurrent()->assertLongsEqual((long)0, (long)0, NULLPTR, file, line); \
   } } while(0)
 
+#define CHECK_EQUAL_ZERO(actual) CHECK_EQUAL(0, (actual))
+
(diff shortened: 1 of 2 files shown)


basvodde:
No test?

rnestler:
> No test?

Since they are just aliases for `CHECK_EQUAL` I'm not sure if it makes much sense to add tests.

But I guess I could add some tests to tests/CppUTest/TestUTestMacro.cpp similar to the `CHECK_EQUAL` tests. Would that be good?

basvodde:
Yes, please do put a test there, and perhaps also tests that it behaves properly as a macro.

I'd suggest CHECK_EQUAL_ZERO instead to CHECK_ZERO.

rnestler:
I did the rename and added some tests.

Btw. It took me quite some time to figure out how to run the tests (apparently `make check` ). It would be nice to have a developer contributing guide. Maybe add a  CONTRIBUTING.md with instructions to describe how to contribute to the project.

basvodde:
Sorry, how to run the tests also does depend on the build system that you are using. We support multiple.

The test failed though.

rnestler:
> The test failed though.

A copy paste error :see_no_evil:

PR #1425, opened 2020-09-18
Add command-line option to control crash-on-fail mode

CiderMan:


include/CppUTest/Utest.h
```diff
@@ -171,6 +171,10 @@ class UtestShell
 
     virtual SimpleString getMacroName() const;
     TestResult *getTestResult();
+
+    static const NormalTestTerminator normalTestTerminator_;
+    static const CrashingTestTerminator crashingTestTerminator_;
```
basvodde, line 176: Why this change?

src/CppUTest/CommandLineTestRunner.cpp
```diff
@@ -95,6 +95,7 @@ void CommandLineTestRunner::initializeTestRun()
     if (arguments_->isColor()) output_->color();
     if (arguments_->runTestsInSeperateProcess()) registry_->setRunTestsInSeperateProcess();
     if (arguments_->isRunIgnored()) registry_->setRunIgnored();
+    if (arguments_->isCrashingOnFail()) registry_->setCrashOnFail();
```
basvodde, line 98: Why do this through the registery and not go directly to UTestShell?

include/CppUTest/Utest.h
CiderMan: I moved them from statics in the source file to protected statics in the class to make the testing easier.

src/CppUTest/CommandLineTestRunner.cpp
CiderMan: I was just following the pattern for the rest of the arguments. As far as I could see, they all went through the registry so I assumed that was the design intent and did the same for the option that I added.

basvodde: Nah, I suggest to not bother with the registry. It seems it doesn't need to know about crash on failure. Adding that won't create any benefits IMHO.

tests/CppUTest/TestRegistryTest.cpp
```diff
@@ -425,3 +430,16 @@ TEST(TestRegistry, reverseZeroTests)
...
+TEST(TestRegistry, doesNotCrashIfNotSetToCrash)
+{
+    CHECK_FALSE(MockTest::wouldCrash());
+}
+
+TEST(TestRegistry, crashesIfSetToCrash)
+{
+    myRegistry->setCrashOnFail();
+
+    CHECK(MockTest::wouldCrash());
```
basvodde, line 443: Why not use the "static void setCrashMethod(void (*crashme)());" instead of exposing the test to the inner state?

include/CppUTest/Utest.h
basvodde: Yeah, please don't do that ;) I've commented below how to change the test so that this isn't needed.

tests/CppUTest/TestRegistryTest.cpp
CiderMan: Actually, when I remove the Registry from the chain, I don't need these tests and I can revert the Utest.{cpp,h} changes anyway!

basvodde:
Yup easier, now the CppUMock changes, right?

PR #1427, opened 2020-10-06
Add extra debug new overloads to avoid conflict with MFC versions

CiderMan:


basvodde:
I'm surprised it passes all tests :) That means, I have no reason to not accept this... Thanks!

basvodde:
@CiderMan I completely missed this... but this merge request actually failed the cron build. Could you have a look at it, otherwise I'm likely to need to remove this again :(

CiderMan:
Hi @basvodde, yes, I can take a look if you could point me to "the cron build" as I've not come across it before. We are using this on both Windows and Linux so I really don't want it to disappear again!

CiderMan:
Ah, I think that I've found it. From what I can see, mem leak detection is turned off under certain circumstance [here](https://github.com/cpputest/cpputest/blob/138c775b607b55d0fc4d2fcb072e8f71682989f5/include/CppUTest/CppUTestConfig.h#L188), which sets `CPPUTEST_USE_MEM_LEAK_DETECTION` to 0.
However, [the tests](https://github.com/cpputest/cpputest/blob/138c775b607b55d0fc4d2fcb072e8f71682989f5/tests/CppUTest/MemoryOperatorOverloadTest.cpp#L131) are guarded by `#if ! defined CPPUTEST_MEM_LEAK_DETECTION_DISABLED || ! CPPUTEST_MEM_LEAK_DETECTION_DISABLED`
I guess that this condition needs to be updated to include `CPPUTEST_USE_MEM_LEAK_DETECTION` but I first wanted to check with you how the `CPPUTEST_USE_MEM_LEAK_DETECTION_*` definitions are supposed to interoperate?

basvodde:
Yeah, I think the check needs to not be on the #if defined at all. Just changing it to use the USE_MEM_LEAK_DETECTION ought to do, right?

CiderMan:
I'll take your word for it - I find all the different MEM_LEAK defines a bit opaque. Anyway, it works locally so I've created a PR. I'm not sure how (whether?) I can run the cron builds ahead of it being merged?

basvodde:
Yeah, there is some history to that. We wanted to have the disabled because it makes it easy to pass -DMEM_DISABLED as flag. But the #if defined led to some problems and decided to go with #if instead of #ifdef on most places. That is mostly consistent... but when not, it leads to surprise problems (like this one)

Issue #1415, opened 2020-08-20
Break into debugger on failure, instead of crashing


TheSoftwareGuy:
Usually when TDDing, you want know know from the outset about *all* the tests that fail, and because of this it is usually discouraged to have any `mock().crashOnFailure` checked into the source code. but when developing and iterating on a test/series of tests you want the tests to crash so you can observe everything in a debugger. But it becomes annoying to have to re-compile the tests just to use the debugger properly.

I suggest that the framework could try to break into a debugger without seg-faulting, as the Catch2 framework does with the `--break` [flag](https://github.com/catchorg/Catch2/blob/master/docs/command-line.md#breaking-into-the-debugger.)
 They do this with the CATCH_BREAK_INTO_DEBUGGER macro, which tries to use some platform-specific asm magic to force a breakpoint in the debugger: https://github.com/catchorg/Catch2/blob/bad3c930493d08dd7cca2af57cf355d9b7cddb21/include/internal/catch_debugger.h#L59

CiderMan:
I second the proposal for something like this. While we TDD in places, a lot of our code was written before our unit-test framework was put in place and, as a result, debugging a test failure can be quite involved. Several devs have requested that the framework trigger the JIT debugger when the test fails when running in debug mode - they currently have to run it from the debugger, insert breakpoints, etc., which is doable but this option would certainly make it easier and quicker.

I've spent some time looking (Googling and looking at the source code) into this (I would be happy if this were compile-time even) but can't see anything. It looks like a custom TestTerminator object would do this but I can't see how to use any implementation other than the default.

stephen-d-hill:
Was requested some time ago in [https://github.com/cpputest/cpputest/issues/1012](https://github.com/cpputest/cpputest/issues/1012)

CiderMan:
For traceability, the mailing list thread is [here](https://groups.google.com/g/cpputest/c/YccLe8MaGUU)

TheSoftwareGuy:
At the very least, it would be really great if "crashOnFailure" could be enabled/disabled via command line flag. It's a bit annoying to have to recompile every time I want to change that behavior

CiderMan:
@basvodde, I believe that this issue can now be closed.

Issue #1012, opened 2016-06-30
Test failure handling improvement request


stephen-d-hill:
When developing testcases, typically for legacy code, one finds it a little painful as you just get the test failure printed out on the shell.
It would be a nice feature if one could hook into the testcase failure mechanism to perform things like automatically bringing up the debugger at the file/line of the failure or maybe flushing logging files, if one were building for debug. This could be performed via a variety of methods, e.g. callback function or by raising an exception so one could use the standard unhandled exception mechanism.

uecasm:
For mock failures, you can use `mock().crashOnFailure()` to make it crash (thereby breaking into the debugger, if one is attached) for failures like unexpected calls.  These are generally the sort of things that a file+line number in the test aren't really sufficient for.

For test check failures, the line by itself should be sufficient in all cases except when you're using it in a loop or something more ambiguous like that -- you can use the `_TEXT` suffix macros to add some disambiguating information to the failure for those cases.

Another thing you can do is to set a breakpoint in `UtestShell::failWith`.  When this gets hit, you can examine the call stack to find the exact point and context of failure.

basvodde:
@stephen-d-hill as @uecasm mentioned, the crashOnFailure is created for that. There is no platform independent way to automatically start the debugger, so I think that won't be possible. Though the crashOnFailure should work independent of platform.

Close this issue?

stephen-d-hill:
Please close this issue.

PR #1430, opened 2020-10-15
Extend crashing to c interface

CiderMan:
This is a necessary step before the CppUMock crash-on-failure mechanism can be removed


tests/CppUTest/TestHarness_cTest.cpp
```diff
@@ -96,7 +96,7 @@ TEST(TestHarness_c, checkBool)
     CHECK_EQUAL_C_BOOL(1, 2);
     fixture->setTestFunction(_failBoolMethod);
     fixture->runAllTests();
-    fixture->assertPrintContains("expected <true>\n	but was  <false>");
```
CiderMan, line 99: Hard tabs in strings replaced with \t so that file can be retabbed, if necessary, without damaging these strings.

basvodde:
Is this needed?

A crashingWithoutExceptions makes no sense because... crashing will throw a runtime exception.

CiderMan:
> Is this needed?
> 
> A crashingWithoutExceptions makes no sense because... crashing will throw a runtime exception.

Yes, which can trigger the JIT debugger.

More to the point of this change-set however, the CppUMock crash-on-fail implementation crashes both with and without exceptions. This means that I can either replace the 'with exceptions' bit with the new UtestShell mechanism and leave the 'without exceptions' bit in CppUMock, or I can remove the mechanism from CppUmock completely, which requires this change-set.

It feel better to me to use the UtestShell mechanism for both use-cases: hence, this change-set

basvodde:
Ok, we'll need to consider this a bit more. I do not think a "CrashWithoutExceptions" terminator is the way we want to go. When it crashes, it will throw an exception. The termination part is the same, so making a new class and default for that doesn't make sense to me (yet).

CiderMan:
Hi @basvodde, have you had a chance to think more about this? I would like to get this (and then my last change-set) delivered. The CppUMock already uses UT_CRASH() (in MockFailureReporterTestTerminatorForInCOnlyCode) so this change-set (and the following one) have built on that existing approach.

basvodde:
Yes, sorry.

Basically, you want to have the TestTerminator that is used in C to be replaceable with a crashable terminator. If so, I suggest that you create a C interface (in TestHarness_c.h) for changing the TestTerminator and not put that in UTestShell. That way you can separate this from the UTestShell and keep it in the place where it belongs.

And, I still do not think there is a need for a CrashingWithoutExceptions class :)

Would that work?

CiderMan:
I'm not sure that I understand either what you are suggesting not where the problem is. CppUMock has a class that derives from TestTerminatorWithoutExceptions and adds crashing to it. Hence, I have created a (core) class that derives from TestTerminatorWithoutExceptions and adds crashing to it. This change-set adds that and the next change-set removes the CppUMock class so that, combined, they are really just moving an existing class from CppUMock into core CppUTest... I think that I must be missing something so please help me out.

CiderMan:
@basvodde Since this change-set is now a dependency for issue #1709, I have update from the latest codebase. Since it has been a while, here is that history of this change-set:

Issues #1415 and #1012 had both requested the ability to (optionally) trigger a JIT debugger on test failure. This feature was already available in Mock but not in the core CppUTest. I came across these as I had a number of my development team requesting the same functionality, so I offered to add the feature.

We had [a discussion in the Google group](https://groups.google.com/g/cpputest/c/YccLe8MaGUU) where we agreed the approach and the following phases:

1. Add the command line option and support for it in the normal TestTerminator (PRs #1422 and #1425) - **Complete**
2. Add a crashing version of the TestTerminatorWithoutExceptions for C support (This PR #1430) - **In review**
3. Remove the crashing support in Mock, as we can now use the new core mechanism for that - **Ready for review**

You signed off the first phase (which allowed me to roll it out to my dev team (who have found it very useful on those occasions when they need it -- thanks!) on the basis that I had to deliver the 2 subsequent phases.

I am also happy to pick up #1709 but it requires phase 2 (this PR) to be delivered as it needs the `getCurrentTestTerminatorWithoutExceptions()` method that is introduced by it.

Hence, please can we progress this PR so that I can 1) deliver the follow up work that I promised you , and 2) fix #1709?

Thanks in advance.

basvodde:
Sorry @CiderMan , let me try to give this priority the next days and find a solution to this.

I reread the PR and my previous comments... and I'm still convinced that this is not the right way... or I am seeing it wrong :)

A crash means it will throw an exception. So a CrashingTestTerminatorWithoutExceptions is a contradiction because the crash will cause an exception. Therefore, I'm super uncomfortable having a class in the code that inherently has a contradiction in the name :) Also, I'm uncomfortable having an additional static member in the UTestShell class especially as crashWithoutException should not be a common use case.

So, what am I missing here?

CiderMan:
Thanks @basvodde. I think that the confusing aspect is the overloading of the term "Exception". The `CrashingTestTerminatorWithoutExceptions` class does **not** use C++ Exceptions (which is the fundamental requirement for the class) but still raises an OS Exception (by trying to execute a NULL pointer), which is fully supported in C code and will trigger the JIT debugger for platforms where that is supported. I could rename the `WithoutExceptions` classes to `WithoutCxxExceptions` fi you think that would be clearer?

basvodde:
That would make it a bit more clear... but still, I don't think UTestShell is the right place for this behavior. When this is C-Specific behavior then I would prefer to keep this in the C interface and not pollute the UTestShell with it. I consider adding a static member to a class to be a big thing that I would always try to avoid.

So, is it possible to have a C interface for replacing the TestTerminator in the C interface, rather than putting a new TestTerminator as a static member in the UTestShell ?

CiderMan:
To be clear, `CrashingTestTerminatorWithoutExceptions` (and, in fact, `CrashingTestTerminator`) will result in a `SIGSEGV` on Linux. On Windows, it will result in a hardware exception being triggered under the [Structured Exception Handling](https://learn.microsoft.com/en-gb/windows/win32/debug/structured-exception-handling) framework. From the link:
> Structured exception handling enables you to have complete control over the handling of exceptions, provides support for debuggers, and is usable across all programming languages and machines.

CiderMan:
> So, is it possible to have a C interface for replacing the TestTerminator in the C interface, rather than putting a new TestTerminator as a static member in the UTestShell ?

Hmm, it has been a couple years since I looked at this, so I will have to take a look. However, the current code follows the same pattern as the existing `CrashingTestTerminator`, which is nice. Do we really want to use a different pattern for the C support compared to the C++ support?

basvodde:
So far, all the C functionality is put in the C interface, so if the C interface is the main (only) user of this, then yeah, that would seem a better place.

Any functionality in a heavily used class such as UTestShell ought to be used from everywhere.

So, my second thought (after, a crashing terminator without exceptions is a contradiction) is that I'm not sure we would like functionality that is purely for the C interface in a core class like UTestShell. Unless we have very good reasons for that... but even then... adding a second static variable of the same type as the first one but then named after its subclass feels like a huge code smell.

src/CppUTest/Utest.cpp
```diff
@@ -774,6 +784,16 @@ CrashingTestTerminator::~CrashingTestTerminator()
 {
 }
 
+void CrashingTestTerminatorWithoutExceptions::exitCurrentTest() const
+{
+    UtestShell::crash();
+    TestTerminatorWithoutExceptions::exitCurrentTest();
```
basvodde, line 790: This line would never get executed, right?

basvodde:
Let me think a bit further on alternatives here. I'll comment within 15 minutes again, exploring the code a bit.

basvodde:
Ok. I do think it is best to make a NormalTestTerminatorForC and a CrashingTestTerminatorForC and put them in the TestHarness_c.cpp.

Then add a function to switch to crashing terminator also in the C interface... as that is where it seems to belongs.

This way, there does not need to be any changes int he UTestShell at all and the change can be local to the C interface.

Also, we can keep the "WithoutExceptions" class to mean just that. The reason is that this is also used in the MemoryWarningPlugin and there it really means that it should not throw any exceptions.

What do you think of that?

CiderMan:
> Ok. I do think it is best to make a NormalTestTerminatorForC and a CrashingTestTerminatorForC and put them in the TestHarness_c.cpp.
> 
> Then add a function to switch to crashing terminator also in the C interface... as that is where it seems to belongs.
> 
> This way, there does not need to be any changes int he UTestShell at all and the change can be local to the C interface.
> 
> Also, we can keep the "WithoutExceptions" class to mean just that. The reason is that this is also used in the MemoryWarningPlugin and there it really means that it should not throw any exceptions.
> 
> What do you think of that?

Changing the MemoryWarningPlugin to use the new class (via the `getCurrentTestTerminatorWithoutExceptions()` method provided in this change-set) is exactly what issue #1709 needs so, once that issue is fixed, all uses will have to go through the new classes in TestHarness_c.cpp except for the `NormalTestTerminator` (and sub-classes) when built without `CPPUTEST_HAVE_EXCEPTIONS`. Hence, I guess that the two test terminators "WithoutExceptions" (both the existing one and my addition) can be pulled out of UTestShell and into TestHarness_c - but the normal test terminator will need to call into TestHarness_c to terminate a test if it does not have exception support. I'm assuming that you want to keep the code DRY, rather than duplicating the same code in UTestShell and TestHarness_c? However, this does a add a dependency between UTestShell and TestHarness_c.

Ultimately, I am happy to go with what you want to do so you are a chief maintainer and you have to live with the result more than I do (or anyone else!) However, FWIW, my feelings are:
* The WithoutExceptions terminators are used by the normal terminators so it seems sensible to keep the definitions together. However, I think that your suggestion of renaming them as `NormalTestTerminatorForC` and `CrashingTestTerminatorForC` makes this a lot clearer
* I think that having the enable/disable of crashing in UTestShell for both C++ and C terminators is cleaner when approached from the perspective of the feature (especially if the terminator class is still going to be in UTestShell). Also, if the "ForC" terminator is managed inside TestHarness_c, the fix for #1709 will then require MemoryLeakWarningPlugin to include, and call into, TestHarness_c. In short, I think that moving that functionality into TestHarness_c will cause more coupling between the C++ core and that module. However, if you would rather keep the switching between normal and crashing modes of operation of the "ForC" terminators inside TestHarness_c instead of UTestShell, it will be a fair amount more code (there is nothing like it there currently) but I am happy to do it.

Thanks,

S.

basvodde:
Ok, I'm going to merge the original PR anyways. I do not like the code but I don't like the alternatives either. Let me see if I can find a better way around this ugly UTestShell statics another time.

Sorry for the constant change and thanks for your patience.

CiderMan:
Many thanks @basvodde. I'll create another PR now for the last part of my promised clean up but I'll hold off starting #1709 until you've had a chance to decide what (if anything) we should do to address it.

PR #1432, opened 2020-10-28
Clang 11 support

offa:
Update of the clang build to latest version and missing overrides added.


basvodde:
Wow, that sucks.

Still, thanks! :)

PR #1433, opened 2020-11-13
Errors for string comparisons to show non-printable control characters as C escape codes

jgonzalezdr:
When a string comparison (e.g. STRCMP_EQUAL) fails and the string contains non-printable characters (newlines, tabs, or even worse, ANSI color escape codes!) it's a mess to locate in the error log exactly where the comparison failed.

To avoid this, this PR makes cpputest replace non-printable characters with their corresponding C escape cores when printing the error log, making it easier to compare both strings.

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -422,6 +422,64 @@ void SimpleString::replace(const char* to, const char* with)
         setInternalBufferAsEmptyString();
 }
 
+SimpleString SimpleString::printable() const
+{
+    const char* shortEscapeCodes[] =
+    {
+        "\\a",
+        "\\b",
+        "\\t",
+        "\\n",
+        "\\v",
+        "\\f",
+        "\\r"
+    };
+
+    size_t str_size = size();
+    size_t new_str_size = str_size;
+
+    for (size_t i = 0; i < str_size; i++)
+    {
+        unsigned char c = (unsigned char) buffer_[i];
+        if ((c >= 0x07) && (c <= 0x0D))
+        {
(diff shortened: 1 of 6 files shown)


src/CppUTest/SimpleString.cpp
```diff
@@ -422,6 +422,64 @@ void SimpleString::replace(const char* to, const char* with)
...
+        "\\r"
+    };
+
+    size_t str_size = size();
+    size_t new_str_size = str_size;
+
+    for (size_t i = 0; i < str_size; i++)
+    {
+        unsigned char c = (unsigned char) buffer_[i];
+        if ((c >= 0x07) && (c <= 0x0D))
```
basvodde, line 444: Can you extract this if in a separate method so it is more clear what this is about... and to remove the duplication with the if below?

jgonzalezdr: I don't get your intention. Do you mean that you'd like the if conditions to be extracted as separate methods (like `if (hasShortEscapeCode(c)) ...`? Or do ypu want the if/else if/else statements to be extracted to a couple of methods like `adjustPrintableSize()`/`writePrintableChar` and in this method just keep the two loops calling these methods?

basvodde: Start with extracting the condition as they are and cryptic and duplicate. If the function is too long, we can look at extracting other parts too.

jgonzalezdr: What do you think about the new refactoring?

basvodde:
Much better. Still a bit large... but good enough for now. Tnx!

PR #1442, opened 2020-12-27
Fix reference to CMakeLists.txt (add missing s)

fazl:
There is no CMakeList.txt file so I guess it was a typo on line 37..

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ You can use `make install` if you want to install CppUTest system-wide
 You can also use CMake, which also works for Windows Visual Studio.
 
 * Download latest version
-* cmake CMakeList.txt
+* cmake CMakeLists.txt
 * make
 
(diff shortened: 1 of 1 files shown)


basvodde:
Thanks!

fazl:
Believe it or not, after I installed cygwin and  g++ and tried building
cpputest, I wasted a lot of time before I learned that I was missing both
make and pkg-config that were needed to build it... Doh!  Not sure how best
to protect other new users from such simple problems.

Would probably be good if the cmake or the autoconf approaches would detect
this and warn?

Apparently the pkg-config need is a cygwin specific thing, but I'm not
certain about that...

Hope this is useful feedback!

Kindest regards, Fazl
Help prevent spam by using the BCC field -
http://lmgtfy.com/?q=how+to+use+bcc

PR #1444, opened 2021-01-20
filterGcov.sh fails if your filename or path contains "File"

cfr-mir:
If you filename contains "File" like ....configFile.c
The filter removes the whole file name minus ".c".
This inhibits the html output to access the coverage file.


basvodde:
Fascinating :)

Thanks for fixing that.

PR #1447, opened 2021-02-01
Add missing clang check

### LINKED ISSUE #1445: Compiler warning when building library with g++ on Ubuntu
Hello! When building cpputest with g++ 9.3.0 on Ubuntu 20.04 and enabling -Wundef I get the following warning

```
warning:
../libs/cpputest/src/Platforms/Gcc/UtestPlatform.cpp:178:8: warning: "clang_major" is not defined, evaluates to 0 [-Wundef]
178 | #if !((clang_major == 3) && (clang_minor == 0))
```
I wonder if there is an` #ifdef __clang__` missing in UtestPlatform.cpp?

spacer12345:
There is a check for a clang version that is not guarded by a check if clang is defined. This MR adds this guard otherwise there is a compiler warning for compiling with g++ and -Wundef.

Fixes #1445


basvodde:
Thanks!

Issue #1449, opened 2021-02-08
Update this project for the addition of CppUTest to Conan Center


claremacrae:
Now that CppUTest is in Conan Center, what updates should be made to this repo?

Options:

* [x] Document how to consume the package, in CMake and possible other generators too
* [ ] Document what needs to be done in conan-center-index to add any future new releases of this project

claremacrae:
See the related ticket: https://github.com/cpputest/conan-cpputest/issues/22

offa:
> Document what needs to be done in conan-center-index to add any future new releases of this project

I don't see this as part of the CppuTest project. Additionally, Conan Center provides related docs already.

claremacrae:
> > Document what needs to be done in conan-center-index to add any future new releases of this project
> 
> I don't see this as part of the CppuTest project. Additionally, Conan Center provides related docs already.

OK, that's a fair point... Perhaps instead a reminder that when a new release is done, the recipe in Conan-center-index should be updated...

The text I would have added would mainly have said to follow Conan's PR procedure add the new version to these two files:

* https://github.com/conan-io/conan-center-index/blob/master/recipes/cpputest/config.yml
* https://github.com/conan-io/conan-center-index/blob/master/recipes/cpputest/all/conandata.yml

offa:
I'm working on a conan example for the readme.

Btw. huge thanks to you @claremacrae for getting CppUtest into conan-center.

claremacrae:
> I'm working on a conan example for the readme.

Oh that's great news - thanks very much!

> Btw. huge thanks to you @claremacrae for getting CppUtest into conan-center.

You're very welcome... it took some persistence, but I learned a lot along the way! 😄

bryceschober:
I was going to suggest that we should add a conan badge to CppUTest... but it seems that maybe that isn't yet supported.

offa:
> I was going to suggest that we should add a conan badge to CppUTest... but it seems that maybe that isn't yet supported.

Already added a few minutes ago. But yeah, no official one available yet.

offa:
I've updated to badge, it shows the current version now.

claremacrae:
What remains on this is to decide if/how to write up how to update the files in Conan-center-index when new releases of this project are made...

There was a discussion about about whether to do this - and what it might say.

[...]

PR #1450, opened 2021-02-08
Conan doc

offa:
Adds a basic example of how to consume the library through Conan (#1449).

I have cleaned up the shields; can be reverted if the old format was better.


offa:
@claremacrae I have updated to link to point to conan center (again).

basvodde:
Thanks!

PR #1454, opened 2021-03-02
make test shows test binary return codes

### LINKED ISSUE #1443: `make test` hides nonzero return codes
I recently learned that if you make a test case that divides by 0 the test binary will crash, which is to be expected.
The problem is that `make test` returns 0 (meaning success), and the only difference while glancing at the output is that the green `OK` was replaced by an uncoloured `Floating point exception` that just blends in with the other uncoloured text.

Returning 0 here is quite suboptimal, I'd really like any nonzero return code and maybe even some red text.

I run tests via `make test` (or in CI by having `junit: CPPUTEST_EXE_FLAGS := $(CPPUTEST_EXE_FLAGS) -o junit` and running `make junit`)

I'm running cpputest from [master from a few weeks ago](https://github.com/cpputest/cpputest/commit/887621db6c7cb6a9bd25e6fee90be22b352b296c).
I did check current master and it does the same thing.

Proposed fix approach:
In `MakefileWorker.mk` change the test and vtest targets to be something along the lines of
```
TEST_RETURN_CODE=/tmp/foo
test: $(TEST_TARGET)
        ($(RUN_TEST_TARGET); echo $$? > $(TEST_RETURN_CODE)) | tee $(TEST_OUTPUT)
        @ret=$$(cat $(TEST_RETURN_CODE)); rm $(TEST_RETURN_CODE); if [ "$$ret" -ne 0 ]; then echo "$$(tput setaf 1)$(TEST_TARGET) returned $${ret}$$(tput sgr0)"; fi; exit $$ret
```
I don't like the hardcoded filename or how hard the shell script is to read, but it's working on my debian machine and a coworkers cygwin.

mtfurlan:
Previously if your tests segfaulted, `make test` would return
successfully, and print a simple warning in white text mixed in with all
the other output.
This checks the return code from the test binary and prints a warning if
it's nonzero.

Closes #1443

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -494,11 +494,14 @@ $(TARGET_LIB): $(OBJ)
 	$(SILENCE)$(AR) $(ARFLAGS) $@ $^
 	$(SILENCE)$(RANLIB) $@
 
+TEST_RUN_RETURN_CODE_FILE:=$(shell mktemp /tmp/cpputestResult.XXX)
 test: $(TEST_TARGET)
-	$(RUN_TEST_TARGET) | tee $(TEST_OUTPUT)
+	($(RUN_TEST_TARGET); echo $$? > $(TEST_RUN_RETURN_CODE_FILE)) | tee $(TEST_OUTPUT)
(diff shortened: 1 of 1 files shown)


basvodde:
So far so good. I'll merge this as soon as the daily cron build is green again. That is being worked on in another PR thread

PR #1457, opened 2021-03-02
Fix clang version check and remove clang 12 warning

### LINKED ISSUE #1456: Clang version checking bugs
I saw an error `/bin/sh: 1: [[: not found` and traced it to how `CLANG_VERSION_NUM_GT_700` is getting set.
Turns out the entire clang version checking stuff doesn't work on my system.

A sampling of the output of `clang++ -v 2>&1 | head -n1`:
* my debian buster with clang from apt:
  * `clang version 7.0.1-8+deb10u2 (tags/RELEASE_701/final)`
* coworkers macs with clang from xcode:
  * `Apple clang version 11.0.3 (clang-1103.0.32.62)`
  * `Apple clang version 12.0.0 (clang-1200.0.32.2)`
* [random ubuntu people](https://askubuntu.com/questions/561750/how-i-know-what-version-of-clang-i-have-installed):
  * `Ubuntu clang version 3.5-1ubuntu1 (trunk) (based on LLVM 3.5)`
  * `Ubuntu clang version 11.0.0-++20200916113328+8f2c29681ce-1~exp1~20200916094014.99`

`CLANG_VERSION` is set by putting that through [`grep -o 'clang-[0-9][0-9][0-9]*.'`](https://github.com/cpputest/cpputest/blob/master/build/MakefileWorker.mk#L205)
In the end, `CLANG_VERSION_NUM` has `1103` on my coworkers mac, empty for my debian, and empty for the ubuntu strings.

The best idea I have is replace the grep with
`sed -n 's/.* \([0-9]*\.[0-9]*\.[0-9]*\).*/\1/p' | sed 's/\.//g'`
Works for everything except the first ubuntu example, which is from before 7, so the end check of "is it greater than 7.0.0" will be false, which is correct.

Also, clang 12 added `-Wno-poison-system-directories`, so it would be cool if we could add a second check for that.

Going back to the reason I was looking at this, on my debian  `$(shell whatever)` runs whatever with sh, not bash, and sh doesn't believe in `[[`.
`[` does support `-ge` though, so I don't see any reason not to change
```diff
-CLANG_VERSION_NUM_GT_700 := $(shell [[ $(CLANG_VERSION_NUM) -ge 700 ]] && echo Y)
+CLANG_VERSION_NUM_GT_700 := $(shell [ "$(CLANG_VERSION_NUM)" -ge 700 ] && echo Y)
```

mtfurlan:
The clang version check in build/MakefileWorker.mk only worked for apple
clang, so this changes the version checking to work with apple, debian,
and ubuntu clang version strings.

Additionally, clang 12 added -Wpoison-system-directories, which warns if
you if you include stuff from /usr/local/include when doing a cross
compile.
Apple clang thinks it's always doing a cross compile for some reason, so
that always complains.

Closes #1456

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -201,16 +201,22 @@ ifeq ($(COMPILER_NAME),$(CLANG_STR))
 	CPPUTEST_CXX_WARNINGFLAGS += -Weverything -Wno-disabled-macro-expansion -Wno-padded -Wno-global-constructors -Wno-exit-time-destructors -Wno-weak-vtables -Wno-old-style-cast -Wno-c++11-long-long -Wno-c++98-compat-pedantic -Wno-reserved-id-macro -Wno-keyword-macro
 	CPPUTEST_C_WARNINGFLAGS += -Weverything -Wno-padded
 
-# Clang "7" or newer (Xcode 7 or newer command-line tools) introduced new warnings by default that don't exist on previous versions of clang and cause errors when present.
-CLANG_VERSION := $(shell echo $(CC_VERSION_OUTPUT) | grep -o 'clang-[0-9][0-9][0-9]*.')
-CLANG_VERSION_NUM := $(subst .,,$(subst clang-,,$(CLANG_VERSION)))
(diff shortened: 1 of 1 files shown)


basvodde:
Thanks!

PR #1462, opened 2021-04-20
fix project page link

eugeniomiro:
Just tried to open this link and found it broken so fixed it with a working one, hope that helps


eugeniomiro:
I don't think those failed tests are mine but let me know if there's anything I can do with it

basvodde:
Indeed, it seems the new docker policy is causing the build to fail. That is very interesting... 

Thanks for your contribution!

PR #1470, opened 2021-06-18
Change TestCase to TestSuite in GTestConvertor

mickeykkim:
This commit changes references of testing::TestCase to testing::TestSuite in GTestConvertor.h. This is necessary to update Google Test integration to the latest version.

### DIFF
--- a/include/CppUTestExt/GTestConvertor.h
+++ b/include/CppUTestExt/GTestConvertor.h
@@ -54,7 +54,7 @@ class GTestFlagsThatAllocateMemory;
 
 namespace testing {
     class TestInfo;
-    class TestCase;
+    class TestSuite;
     class Test;
 }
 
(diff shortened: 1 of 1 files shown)


NikosZampelis:
Please check lines 130 and 142 if you have the time... the GTEST_FLAG(internal_run_death_test) calls need further updates. I guess with the GTestFlagSaver class.

basvodde:
I think I cannot merge this as it is now. I want to, but we currently support GTest 1.4-1.7 and, as far as I can see, this change breaks the support for the previous versions. So, the changes will need to be in a new #ifdef (unfortunately). Or am I wrong about that?

NikosZampelis:
@basvodde Unfortunately, you are correct. This would break older versions without #ifdefs. This PR needs to also be updated further as it does not update Google Test integration with the latest Google Test version. My previous comment is an example of the differences.
Note: Currently GTest is in version 1.11. Do you want to consider dropping support for older versions at some point?

basvodde:
We do not really want to stop the support of earlier versions. CppUTest is often used in large-embedded products and they are often stuck on an old version of a tool (or framework) and have trouble changing that. In order to help with that, we try to support as much old compilers/framework as possible. It also restricts the C++ that is used in building CppUTest

PR #1473, opened 2021-07-03
Fix compile gcc from cygwin

arthurbiancarelli:
By letting CMake handle the windows library suffix (.lib or .a)

_First contribution Hope requirements are met_


basvodde:
Thanks

PR #1477, opened 2021-07-20
Fixed build using bcc2 compiler

Glamhoth:
I am doing some R&D regarding the bcc2 compiler from Gaisler and I wanted to use cpputest as a test library in the future.

While trying to build the library using that compiler, with STD_C `on`, I encountered a linker error - undefined reference to `waitpid`. After some debugging I found, that the library checks if the compiler provides `fork` function and if it does it uses platform specific fork and waitpid implementations (UtestPlatform.cpp:64). Apparently Gaisler uses a stub fork (among some others), but waitpid is missing:
image

To be able to compile the library in such situation I added an additional check - for waitpid availability. With this I am able to use cpputest with my project, and run tests using LEON3 simulator.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,11 @@ if(HAVE_FORK)
   add_definitions(-DCPPUTEST_HAVE_FORK)
 endif(HAVE_FORK)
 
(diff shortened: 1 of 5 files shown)


basvodde:
Looks good, thanks!

Issue #1165, opened 2018-07-10
A question about handling multiple initial conditions in a better way


SaeedYasin:
Many times we have multiple initial conditions, but our test case remains the same. At the moment, using CppUTest we have to define multiple groups and then copy paste same test code for each test group. So, if we have e-g 4 different initial conditions [which will be written in setup functions], then 4 test groups, which means 4 identical test cases for each group, even though we have only one test case scenario which is being tested, but under 4 different initial conditions. 

Or the other approach to solve this problem, is to use empty setup function in test group and then write 4 test cases with initial condition in the start of each test case and then same test code. In both cases, we end up with 4 test cases, when in fact there is only one test scenario being tested.

This means, lots of code redundancy, in this case e-g, instead of 1 test case code, we end up with 4 tests with different initial conditions, but with identical test code. Now, Imagine a practical scenario where we have 10 different initial conditions, but our test scenarios are 30. **We have to write 300 test cases, but in fact there should be only 30 test cases to look at, not 300**.

This makes it hard to debug and maintain test code in the long run. It would be awesome if we can write only the test scenarios being tested in test cases and move the initial setup code to test groups or e-g create a **multigroup** containing multiple setups and teardowns, so we can focus on actual tests instead of making a bloated version of test file which is hard to debug and maintain and more prone to errors.

Sorry for long post, but I just want to know if there is already an existing solution for this?
If not, then any idea of what is the best way to handle this?

I do have a solution for this in which user can create a multigroup and **TEST(multigroupname, test_name)** will run same test on each setup/teardown written in that multigroup definition and I want to make a PR for this, but before that just want to get an idea if my approach is good or is there any better way to do this. Thanks :)

jacob-keller:
I've used functions in the test file for this purpose, so commonize the actual test code into a function, and have different test groups with the respective different setup, but each test calls the common function logic.

I do like the idea of a multigroup though.

jacob-keller:
I'm not sure what the best way to resolve this case is.

The important key to remember is that you want to be able to specify the parameters for a set of tests and repeat those tests with different initial conditions, *without* having to duplicate the test code. I wonder if something like this is possible with something akin to mixins?

i.e. define a single base set of tests and groups, and then define a new group that extends it and sets up different conditions? I'm not quite sure how CppUTest figures out which tests to run for a group, but that would be the direction I would investigate myself.

The idea is you'd have something like the following:

```
TEST_GROUP(group) {
  // test parameters
[...]

PR #1480, opened 2021-07-25
Add MultiGroup support

SaeedYasin:
This is a proposed implementation of the enhancement suggested in #1165

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -80,6 +80,7 @@ include_cpputest_HEADERS = \
 	include/CppUTest/MemoryLeakDetectorMallocMacros.h \
 	include/CppUTest/MemoryLeakDetectorNewMacros.h \
 	include/CppUTest/MemoryLeakWarningPlugin.h \
+	include/CppUTest/MultiGroup.h \
 	include/CppUTest/PlatformSpecificFunctions.h \
 	include/CppUTest/PlatformSpecificFunctions_c.h \
 	include/CppUTest/SimpleString.h \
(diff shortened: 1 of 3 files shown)


SaeedYasin:
@basvodde @jacob-keller does this PR look okay? If possible please provide some feedback, it's been a long time. Thank you.

basvodde:
Sorry @SaeedYasin , I've been super busy and thus usually do not spend enough time on PRs that are large and need some focus.

As it is, it is unlikely to be included in the core of CppUTest. The implementation contains quite a lot of macro code, which we tend to avoid. I've written a lot of tests in CppUTest and never really needed this feature as I can just have multiple tests and call a setup in the test body instead of in a setup.

My initial reaction is to just decline it (sorry) as too complicated for a rare specific scenario.

jacob-keller:
I think there are two cases here:

1st) setup specific to a test. For this, you can define a function with the parameters of the test in the test class and then call it as the first step of the test with the specific test parameters.

2nd) multiple groups of tests where they all share underlying setup and then each group needs its own setup.

For this, my teams tend to use TEST_BASE and TEST_GROUP_BASE. Additionally, we found a convenient "hack" that allows us to also use inheritance on test classes not defined as TEST_BASE:

We define a special macro, TGN:

`#define TGN(group) TEST_GROUP_##CppUTestGroup##group`

this lets us write things like

```
TEST_GROUP_BASE(new_group, TGN(existing_group)) {
  TEST_SETUP {
    TGN(existing_group)::setup()

    // extended setup
  }
}
```

Basically using the TGN macro lets us extend an existing non-base class easily.


However, looking closer, that doesn't fully resolve your problem. I think you should provide more detail and examples.

Your goal is to cover the situation where you have tests that you want to repeat with many different sets of conditions, without having multiple tests defined for each one.

That can definitely be helpful, but I wonder if there is a better way to implement and explain the goal.

Essentially you want to parameterize a test group to expand it based on initial conditions to run many copies with different setups. I suspect there has to be a good way to do what you want. I know other test frame works have somethings similar to this...

SaeedYasin:
@basvodde completely understandable, the reason for the macro based implementation was, that it makes it possible to achieve this functionality with zero risk to the existing code. But anyway, thank you for the feedback.

@jacob-keller yes the goal is to "have tests that you want to repeat with many different sets of conditions, without having multiple tests defined for each one". This removes test code duplication and makes it easier to understand and maintain tests, when the number of tests is high.

THE TGN macro and TEST_GROUP_BASE techniques look very interesting and definitely useful in certain cases. Thank you for the guidance and feedback.

PR #1481, opened 2021-08-04
redirect to github.io

Dinuirar:
github pages are now hosted under `.io` domain

( https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages )


basvodde:
Thank!

PR #1495, opened 2021-10-18
Fix reserved identifier usage

offa:
Clang 13 has become more strict on usage of reserved identifier.


offa:
ping @basvodde

basvodde:
PONG!

basvodde:
Interesting. Thanks!

PR #1496, opened 2021-10-27
More reliable `gettimeofday` detection.

willson556:
`src/Platforms/Gcc/UTestPlatform.cpp:38` relies on `CPPUTEST_HAVE_GETTIMEOFDAY` as an indicator that
`sys/time.h` will exist but that header is not present with TI-CGT-ARM despite the original test succeeding.


basvodde:
Thanks

PR #1499, opened 2021-11-01
Update SimpleString.cpp for Borland compiler

pjshelton:
the Borland compiler does not like the expression (*s1++ = *s2++) as part of a while condition,  This commit rewrites the while loop, moving the assignment out of the while test expression.


basvodde:
Super interesting compiler :) Thanks for the PR!

PR #1501, opened 2021-11-03
Update SimpleString.cpp for Borland 5.4 compiler

pjshelton:
Borland 5.4 does not like (str = StrStr(str, substr.getBuffer())) in a while condition. This rewrites the while() moving the assignment out of the while condition

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -358,7 +358,7 @@ size_t SimpleString::count(const SimpleString& substr) const
 {
     size_t num = 0;
     const char* str = getBuffer();
-    const char* strpart;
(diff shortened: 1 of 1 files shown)


pjshelton:
I think your code scanning issue is a false positive as the uninitialised  use of strpart is guarded by the short circuit guarantee of &&

basvodde:
Sorry for delay. I noticed one of the builds wasn't running anymore and wanted to fix that before merging this code.

I think it would be good to ensure there is no warning even when the warning is a false positive. And initializing to NULL likely does not harm.

pjshelton:
pointer initialised.

basvodde:
Thanks

PR #1503, opened 2021-11-05
add help text for -ll test locations argument

lokimon:


basvodde:
Unfortunately, the second change of printing the line numbers ought to be removed, now it conflicts with the fix that I had made. Sorry about that.

jgonzalezdr:
This PR is not needed anymore since #1569 already implements that.

PR #1507, opened 2021-11-27
Required to  compile with Borland Compiler v5.4

pjshelton:
In the expression a? true : false;  True and False have to be the same type.  The Borland Compiler v5.4 is unable to automatically  promote the char array "(null)" to a SimpleString in this expression.


basvodde:
Interesting

pjshelton:
There are 4 more of these.  Do you want each to be in it's own PR or would it be okay to do all 5 in this PR.

basvodde:
If its ok, I'd rather keep them in separate PRs. That makes it easier to quickly see what they are and how they are (not) related. It'll make the merging faster.

PR #1510, opened 2021-12-03
Required to compile with Borland Compiler v5.4 part6

pjshelton:
In the expression a? true : false; True and False have to be the same type. The Borland Compiler v5.4 is unable to automatically promote the char array "(null)" to a SimpleString in this expression.


basvodde:
Thanks!

PR #1514, opened 2021-12-13
cpputest_longlong for long long

pjshelton:
long long is compiler specific and we already have a cpputest_longlong typedef that takes care of this for us.


basvodde:
Thanks, I'll keep checking them and merging them. These small PRs are easy to quickly check, thanks. Appreciate a lot.

PR #1515, opened 2021-12-14
cpputest_ulonglong for unsigned long long

pjshelton:
unsigned long long is compiler specific and we already have a cpputest_ulonglong typedef that takes care of this for us.


basvodde:
Sorry, wrong button

PR #1518, opened 2021-12-17
Borland 5.4 does not do -include ... on the commandline

pjshelton:
I have to turn the MEMORY_LEAK_DETECTION functionality off because Borland 5.4 does not do -include as a flag to the compiler
And this is the method used to inject the MEMORY_LEAK_DETECTION headers into the code under test.


pjshelton:
Maybe it should be a set when BORLAND is true and not an option.

pjshelton:
Given what my next PR is going to be maybe I don’t even need the else(). 
I am going to add an extra ifelse(BORLAND) to the top of the CppUTestconfigurationOptions.cmake file and set the CPP_PLATFORM variable to Borland and the MEMORY LEAK TEST variable to NO

basvodde:
Ok, so .. what should I do with this PR?

In general, it would probably be better if you can make the default case more clear. E.g., I'm not sure the option in cmake can be overridden, but if so, you can just turn it on (default) and then turn if off again for Borland. That probably makes it easier to see what is going on.

basvodde:
Also, doesn't the Borland compiler have an alternative to -include ?

pjshelton:
Not that I was able to find.  I suppose  I am trying to ask what is the best way to fence the memory leak test.  And if I read your other comment correctly, I should make this my second if needed change.  so you can reject this, I will revert my fork and work on   my next change.

basvodde:
So close this?

pjshelton:
Yes.  I will try again later on this one

Issue #1493, opened 2021-10-12
Compiling with Borland C++ 5.4 for Win32 Copyright (c) 1993, 1999 Inprise Corporation


pjshelton:
Hi,
I am trying to get an application that was first written 20 years ago, using the above compiler, to compile with a more modern compiler.
To do so, I am looking for a Mocking framework that compiles using the above compiler.
My modern compiler is the MinGW-64 compiler.  -- gcc (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0

I have compiled CppUTest with the MinGW-64 compiler and it passes all it's self tests.

I have been compiling with the Borland compiler and have had to make a few changes to the code to get it to compile.
I only needed to make changes to SimpleString.cpp, TestMemoryAllocator.cpp and MockExpectedCall.cpp that did not need \#if statements.  UtestPlatform.cpp was the first file that needs a \#if directive. 

I expect it is in vain, but does any one know how to either  disable compiling MemoryLeakwarningTest.cpp or how to trick the Borland compiler into understanding std::nothrow? This is my current sticking point. 

The changes I have made all still compile and tests all pass using the MinGW-64 compiler.  I am happy to supply patches of the changes if requested.

basvodde:
For different platforms there need to be their own UTestPlatform.cpp. I would expect the Borland compiler requires its own.

As for std::nothrow. I believe it is defined in the CppuTest configuration header files so that we can also support compilers that do not support it. I'm not sure if these are currently in the build, so I'm not 100% sure it is enough... but I assume it shouldn't be too hard to get that too work.

Gosh, Borland... that is a while ago :)

pjshelton:
I added 

#ifdef __BCPLUSPLUS__
const std::nothrow_t std::nothrow;
#endif

at line 38 in MemoryLeakWarningTest.cpp as the compiler was looking for an external invocation and not finding it.  So I made an explicit internal invocation.  cmake was now able to compile and link CppUTestTest.exe. 

I now have a new error.  That executable does not do anything.  it does not respond to any of its command line inputs.
 it just starts then stops having done nothing.   it just stops working without giving any clue as to why.  

CppUTestExtTest.exe runs and only fails one test.  and that test is MockSupport_c

pjshelton:
okay,
TestFailureNaNTest.cpp creates a invalid floating-point operation so I removed it from the compile

of the rest of the tests,
Utestshell (NUMERICAL)
TestHarness_c (Failed)
[...]

PR #1519, opened 2021-12-19
Creating a Borland/UtestPlatform.cpp

pjshelton:
I am using as the base for the Borland/UtestPlatform.cpp file the Gcc/UtestPlatform.cpp file.  This is currently just a copy of the Gcc/UtestPlatform.cpp file.

I will make changes to it in later PR.  This way it will be easier for you to see what I do change, rather than just dump the final updated file on you with unknown changes and no reasoning as to why the differences.

This is in response to the first reply in #1493.


basvodde:
So, I will not merge this? Now it still even has Gcc all over it, so slightly uncomfortable merging it :)

pjshelton:
Okay, I can remove what I think I don't need.  unfortunately this is going to be a case of you are going to have trust that what is left works because at the moment you don't have the borland  compiler to do your own testing with.  it becomes a "large" change.
I made this work by taking what you had done and make just enough changes to get it working. So there is going to be a large number of things that strictly don't need to be there but are because I don't strictly need to remove them because the compiler does.

basvodde:
Ok, that is ok too. So, I'll merge this now?

pjshelton:
Yes please.  The change that will let this file be used in a compilation will come later.

Issue #1520, opened 2021-12-24
Less verbose output


mpfusion:
This is a feature request to allow for less verbose output.

# Background
cpputest, or more precisely, `build/MakefileWorker.mk` provides quite some output even if the `SILENCE = @` variable is set in the `Makefile`. This clutters the terminal and might push more important information off the screen.

# Current behaviour
cpputest outputs `compiling <filename>` once per source file. Then `Building archive` once. Then `a - <directory>/<filename>` once per source file. Then `Linking …` and `Running …`, followed by the test results. This easily leads to +20 lines of output even on small projects.

# Expected behaviour
If the tests pass, I propose to have minimal console output. Compiling, building archives and linking is not important at this point, so the only output I think makes sense is `Running …_tests`, to provide context for the next line reading `OK` (test result).

# Proposed solution
To allow for backwards compatibility I suggest to introduce an additional variable, e.g. `MORE_SILENCE` to be added to the `Makefile`. If the variable is defined, abovementioned output is silenced. The name is of course just an example.

basvodde:
I think I just approved the pull request for this before I saw this :)

PR #1521, opened 2021-12-24
Add option for less verbose output

mpfusion:
This allows the user to define MORE_SILENCE in the Makefile and get rid
of output like “compiling …”, “Building archive …”, etc. The idea is
that a successful test run should have minimal output.

This is a proposed implementation of the enhancement suggested in #1520.


basvodde:
So silent :)

PR #1522, opened 2022-01-02
The first of a number of changes to make this file a Borland specific file.

pjshelton:
removed the first #ifdef __clang__  as this will never be called as this file will only be included when __clang__ is not defined.


pjshelton:
opps code fragment not rendered as i had hoped

PR #1535, opened 2022-01-26
Improve support for strongly-typed/scoped enums

jushar:
This commit specifically:
- Introduces support for strongly-typed enums for mock parameters and return values
- Support strongly-typed enums in CHECK_EQUAL
- Introduces a generic `MockActualCall::returnValueOrDefault` with overloads for each support type (just like it's done for `MockExpectedCall::withParameter`

### DIFF
--- a/include/CppUTest/SimpleString.h
+++ b/include/CppUTest/SimpleString.h
@@ -241,4 +241,14 @@ SimpleString StringFrom(const std::string& other);
 
 #endif
 
+#if defined(__cplusplus) && __cplusplus >= 201103L
+
+template<typename E, typename = typename std::enable_if<std::is_enum<E>::value>::type>
+SimpleString StringFrom(E enumValue)
+{
+  return StringFrom(static_cast<typename std::underlying_type<E>::type>(enumValue));
+}
+
+#endif
+
 #endif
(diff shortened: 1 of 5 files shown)


basvodde:
First of all, I appreciate the contribution!

But unfortunately, my first reaction is to reject it. We have always avoided template usage in the main codebase and I'm not (yet) willing to change that design decision. I know this is very strict and perhaps old-fashioned, yet I'd like to keep the C++ usage as simple as possible to maximize the support for bad compilers (embedded compilers).

Also, I believe we have not added the overloaded returnValueOrDefault because of the amount of times it went wrong or led to compiler errors. We figured out that at the time you call this function, you always know the type and hence there is no benefit to having the overloaded function. Unless you see a common (non-template) scenario where it is needed?

Jusonex:
Thanks for your feedback!

While I principally do understand your concerns about compiler support, I think it's a trade-off between features/usability and support for old compilers. In this case, not supporting scoped enums leads to pretty bad mock code as your only chance to get it to work is by hacking in casts everywhere. Scoped enums are standardized since C++11 so adopting them in 2022 should really be possible.

Could you imagine integrating this compromise?
* Move overloaded `returnValueOrDefault` to a private namespace so it's not part of the public api
* Add an #ifdef around all the `enum class` stuff that is disabled by default and can be enabled if required (and compiler support has been verified)

This surely blows up the code a bit, but that way compiler support can be verified and as soon as the statistics are sufficiently good, the #ifdefs could be removed eventually.

kimkulling:
There is an otion to make these kind of template-usage optional for special compilers. Just as an idea to help.

PR #1539, opened 2022-01-31
Clarification of the message displayed when an actual call fails because of missing parameters

jgonzalezdr:
When CppUTest fails because an actual call matched partially one or more expected calls, but these have still parameters not fulfilled, the message shown is badly formatted and hard to understand.

For example, let's execute a test where a function with several overloads is mocked like the following one: 
``` C++
TEST( TestSuite, ExampleTest )
{
    mock().expectOneCall("foo").withParameter("bar", 0);
    mock().expectOneCall("foo").withParameter("bar", 5);
    mock().expectOneCall("foo").withParameter("bar", 2).withParameter("boo", "blah");
    mock().expectOneCall("foo").withParameter("bar", 2).withParameter("boo", "bleh");

    mock().actualCall("foo").withParameter("bar", 0);
    mock().actualCall("foo").withParameter("bar", 2); // < This call fails

    // ... 

    mock().checkExpectations();
}
```
The message that is displayed is:
```
error: Failure in TEST(TestSuite, ExampleTest)
 	Mock Failure: Expected parameter for function "foo" did not happen.
 	EXPECTED calls that WERE NOT fulfilled related to function: foo
 		foo -> int bar: <5 (0x5)> (expected 1 call, called 0 times)
 		foo -> int bar: <2 (0x2)>, const char* boo: <blah> (expected 1 call, called 0 times)
 		foo -> int bar: <2 (0x2)>, const char* boo: <bleh> (expected 1 call, called 0 times)
 	EXPECTED calls that WERE fulfilled related to function: foo
 		foo -> int bar: <0 (0x0)> (expected 1 call, called 1 time)
 	MISSING parameters that didn't happen:
 		int bar
 int bar
 const char* boo
 const char* boo
```
Besides the bad indentation of the last 3 lines, the information regarding the missing parameters is very hard to understand, because it even mixes parameters from fulfilled and non-fulfilled expectations.

With this PR, the message becomes:
```
error: Failure in TEST(TestSuite, ExampleTest)
 	Mock Failure: Expected parameter for function "foo" did not happen.
 	EXPECTED calls with MISSING parameters related to function: foo
 		foo -> int bar: <2 (0x2)>, const char* boo: <blah> (expected 1 call, called 0 times)
 			MISSING parameters: const char* boo
 		foo -> int bar: <2 (0x2)>, const char* boo: <bleh> (expected 1 call, called 0 times)
 			MISSING parameters: const char* boo
 	EXPECTED calls that WERE NOT fulfilled related to function: foo
 		foo -> int bar: <5 (0x5)> (expected 1 call, called 0 times)
 		foo -> int bar: <2 (0x2)>, const char* boo: <blah> (expected 1 call, called 0 times)
 		foo -> int bar: <2 (0x2)>, const char* boo: <bleh> (expected 1 call, called 0 times)
 	EXPECTED calls that WERE fulfilled related to function: foo
 		foo -> int bar: <0 (0x0)> (expected 1 call, called 1 time)
```
This way, the missing parameters are related to their expectations.

### DIFF
--- a/tests/CppUTestExt/MockFailureTest.cpp
+++ b/tests/CppUTestExt/MockFailureTest.cpp
@@ -63,13 +63,20 @@ TEST_GROUP(MockFailureTest)
         MockFailureReporterForTest::clearReporter();
     }
 
-    void addCallsToList( unsigned int count )
+    void addThreeCallsToList()
     {
-        if(count >= 1) list->addExpectedCall(call1);
-        if(count >= 2) list->addExpectedCall(call2);
-        if(count >= 3) list->addExpectedCall(call3);
-        if(count >= 4) list->addExpectedCall(call4);
-        if(count >= 5) list->addExpectedCall(call5);
+        list->addExpectedCall(call1);
+        list->addExpectedCall(call2);
+        list->addExpectedCall(call3);
+    }
+
+    void addFiveCallsToList()
+    {
+        list->addExpectedCall(call1);
+        list->addExpectedCall(call2);
+        list->addExpectedCall(call3);
+        list->addExpectedCall(call4);
+        list->addExpectedCall(call5);
     }
 
     void checkUnexpectedNthCallMessage(unsigned int count, const char* expectedOrdinal)
@@ -122,7 +129,7 @@ TEST(MockFailureTest, expectedCallDidNotHappen)
     call2->withName("world").withParameter("boo", 2).withParameter("hello", "world");
     call3->withName("haphaphap");
     call3->callWasMade(1);
-    addCallsToList(3);
+    addThreeCallsToList();
 
     MockExpectedCallsDidntHappenFailure failure(UtestShell::getCurrent(), *list);
     STRCMP_EQUAL("Mock Failure: Expected call WAS NOT fulfilled.\n"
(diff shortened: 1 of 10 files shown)


tests/CppUTestExt/MockFailureTest.cpp
```diff
@@ -229,21 +237,35 @@ TEST(MockFailureTest, MockUnexpectedParameterValueFailure)
...
     call2->callWasMade(1);
     call2->inputParameterWasPassed("bar");
     call2->inputParameterWasPassed("boo");
-    call3->withName("unrelated");
-    addAllToList();
+    call3->withName("foo").withParameter("bar", 2).withParameter("boo", "blah").withParameter("baa", 0u);
+    call3->inputParameterWasPassed("bar");
+    call4->withName("foo").withParameter("bar", 20);
+    call5->withName("unrelated");
+    addCallsToList(5);
```
basvodde, line 249: Since this is used only once and the parameter adds a bit of complexity to the test, could you change this to have a function "addThreeCallsToList" and one that does "addFiveCallsToList" Then remove the if statements from the "addCallsToLists" and let the addFive just call the addThree and then add 2 more.

jgonzalezdr: Done

PR #1543, opened 2022-02-28
Update CppUTestConfigurationOptions.cmake for compiling with Borland 5.4 compiler

pjshelton:
Add cxx flags for Borland 5.4 compiler to not report warnings  W8008 (if condition always false) and 8066 (unreachable code)


pjshelton:
okay, should I have done this in file [cmake/Modules/CppUTestConfigurationOptions.cmake] or cmake/tests/CppUTest/Cmakelists.txt?

basvodde:
I think it is ok like this.

PR #1549, opened 2022-03-17
Update CppUTestConfigurationOptions.cmake for compiling with Borland 5.4 compiler

### LINKED ISSUE #1547: Borland v5.4 doesn't do long long
the Borland v5.4 compiler doesn't do the long long integer type.  

Suggesting setting the cmake variable LONGLONG to OFF in cmake/Modules/CppUTestConfigurationOptions.cmake

this is part of #1493

pjshelton:
setting LONGLONG to OFF
This will close #1547 
and is part of #1493


basvodde:
Excellent!

PR #1550, opened 2022-03-17
Add support for generating a test failure when an unexpected exception is thrown

jgonzalezdr:
This way when an unexpected exception is thrown it is handled as a test failure, so failure info is displayed, the remaining tests execution is not aborted, and in some platforms the OS does not display an error message or tries to launch a debugger.

### DIFF
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -392,9 +392,10 @@ UnexpectedExceptionFailure::UnexpectedExceptionFailure(UtestShell* test)
 {
 }
 
-#if defined(__GNUC__)
-static SimpleString demangle(const char* name)
+static SimpleString getExceptionTypeName(const std::exception &e)
 {
+    const char *name = typeid(e).name();
+#if defined(__GNUC__)
     int status = -1;
 
     std::unique_ptr<char, void(*)(void*)> demangledName {
@@ -402,15 +403,14 @@ static SimpleString demangle(const char* name)
         std::free
     };
 
-    return (status==0) ? demangledName.get() : name ;
-}
-#define DEMANGLE(name) demangle(name).asCharString()
+    return (status==0) ? demangledName.get() : name;
 #else
-#define DEMANGLE(name) (name)
+    return name;
 #endif
+}
 
 UnexpectedExceptionFailure::UnexpectedExceptionFailure(UtestShell* test, const std::exception &e)
-: TestFailure(test, StringFromFormat("Unexpected exception of type '%s' was thrown: %s", DEMANGLE(typeid(e).name()), e.what()))
+: TestFailure(test, StringFromFormat("Unexpected exception of type '%s' was thrown: %s", getExceptionTypeName(e).asCharString(), e.what()))
 {
 }
 #endif
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -31,6 +31,13 @@
 #include "CppUTest/SimpleString.h"
 #include "CppUTest/PlatformSpecificFunctions.h"
 
+#if CPPUTEST_USE_STD_CPP_LIB
+#include <typeinfo>
+#if defined(__GNUC__)
+#include <cxxabi.h>
+#endif
+#endif
+
 TestFailure::TestFailure(UtestShell* test, const char* fileName, size_t lineNumber, const SimpleString& theMessage) :
     testName_(test->getFormattedName()), testNameOnly_(test->getName()), fileName_(fileName), lineNumber_(lineNumber), testFileName_(test->getFile()), testLineNumber_(test->getLineNumber()), message_(theMessage)
 {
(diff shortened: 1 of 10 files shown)


basvodde:
Thanks. We'll need to have a discussion on this, because this functionality was not there on purpose :)

When you catch the exception and let it fail the test, then it actually slows down the process of finding the reason for the exception. When you let the exception go through, you immediately get the stack trace and thus it is easy to find where the exception come from. Hence, it was decided not to catch exceptions.

So... why/when would you want the test to fail instead of a crash?

jgonzalezdr:
I've developed this change because we were working on a project and an unreported change in a library made it throw an exception. The continuous integration system logged the stderr output apart, so the test report just showed a truncated last line with just "TEST(xxx, yyy)" and nothing else, so everybody went mad because of this unexpected outcome, and moreover since the rest of the tests were not executed it was difficult to evaluate the scope of the problem (making management people even madder). 

Tests on the embedded target were even worse because it just kept resetting without further information.

Detailed examination of the CI reports, and running it manually in host, just showed additional info printed in stderr:
```
terminate called after throwing an instance of 'std::runtime_error'
  what(): bla bla bla...
```
but no stacktrace.

But I see your point, in some cases it could be more useful to just let the exception go through, mostly when executing the tests manually during development.

I think it would then be a good idea to improve this PR by adding a compile option to enable rethrowing exceptions, and even more useful, a command-line option to do the same thing on demand. Is this OK for you?

jgonzalezdr:
It could be also useful to add on the failure message the stacktrace (if available), that should be fairly feasible for the usual compiler families (gcc, clang and msvc).

basvodde:
A runtime flag could be an option. Generating a stack trace is something that we have looked at in the past, but it is hard to do that in a compiler independent way. Hence we never managed to implement that.

jgonzalezdr:
I've updated the PR to add a runtime flag.

Regarding stacktrace support, I've got a half baked branch that I've started some years ago when we discussed adding support for getting the memory allocation calls' addresses from the stack to avoid having memleak reports with unknown caller info, but I had no time to work on that. If I find time to work on that I'll shoot to birds with a stone and also add stacktrace printing to unexpected exceptions.

basvodde:
I personally do much more prefer that the default behavior is crash on exception. CppUTest is made with TDD in mind and then you would virtually always want to crash and stack trace. Perhaps we can add an -ci option (or something like that) which optimized the test run for in an CI system.

jgonzalezdr:
I've made throwing exceptions the default behavior, inverted the meaning of the -e parameter, and added a -ci parameter.

For the moment -ci is equivalent to -e, but -ci could be used in the future to also set other options suitable for CI environments. We could also make it set -ojunit which is often used in CI environments. That's up to you.

basvodde:
Yes, that looks good. Only thing still is that some of the try/catch seem to be outside the #if CPPUTEST_USE_STD_CPP_LIB
 block. So far, when CPPUTEST_USE_STD_CPP_LIB is turned off then it also assumes there is no exception support in the compiler and hence exceptions shouldn't be used in that case. Could you check whether all exceptions have a non-exception alternative? I hope it fails the daily build (which should include a few more builds as the commit-trigger build) on it but I'm not sure.

jgonzalezdr:
Done.

I've nevertheless did not compile out the command line options and member variables and methods used for this feature when std c++ lib is not used, to avoid overcluttering the code and this PR. The member variables and functions will still be there, but will not be used, so this is not a problem. The command line options will be present in the help display, but will have no effect.

basvodde:
Yeah, that is fine. As long as the exception handling has an alternative

src/CppUTest/TestFailure.cpp
```diff
@@ -378,3 +385,32 @@ FeatureUnsupportedFailure::FeatureUnsupportedFailure(UtestShell* test, const cha
...
+    std::unique_ptr<char, void(*)(void*)> demangledName {
+        abi::__cxa_demangle(name, nullptr, nullptr, &status),
+        std::free
+    };
+
+    return (status==0) ? demangledName.get() : name ;
+}
+#define DEMANGLE(name) demangle(name).asCharString()
+#else
+#define DEMANGLE(name) (name)
```
basvodde, line 409: One last thing here (I think it will be the last thing). Could you not use the macro here and just put the the demangle function always included... but then an #if in the implementation with a "return name" as alternative implementation ?

jgonzalezdr: Done

basvodde:
Thanks!

steverb1:
I'm getting a compile error in TestFailure.cpp with the latest code, running "make tdd":

************************************************************************
src/CppUTest/TestFailure.cpp:401:43: error: no matching constructor for initialization of 'std::unique_ptr<char, void (*)(void *)>'
    std::unique_ptr<char, void(*)(void*)> demangledName {
************************************************************************
If I change the curly braces on 401 and 404 to parentheses, the compile error goes away, but I don't know if that is correct.

jgonzalezdr:
@steverb1 Can you please give some information on the toolchain, environment, build system, OS, etc. which you are using to compile CppUTest?

steverb1:
I'm on a MackBook Pro, 12.3.1 (Monterey).  Just using make to build ("make tdd").   The shell is /bin/zsh.
Are you saying that it compiles for you, with those curly braces?

basvodde:
Indeed, it doesn't work for me locally on my Mac either. I was surprised. The travis build is quite strong but that does pass.

jgonzalezdr:
Fixed that problem in #1574.

PR #1554, opened 2022-03-24
Update UtestTest.cpp for compiling with Borland v5.4 compiler

### LINKED ISSUE #1545: Borland v5.4 throws a runtime error on division by zero: tests\CppUTest\UtestTest.cpp
This is similar to #1544 
The Borland v5.4 compiler does not have a NaN or Inf .
if one wants the program to catch the division by zero  error, and not crash, it has to contain an signal handler that catches the exception and deal with it itself. 
From the help file:
>For / and %, op2 must be nonzero op2 = 0 results in an error. (You can't divide by zero.)

For  tests/CppUTests/UtestTest.cpp this means that the following test fails with a NUMERICAL error. 
```
static volatile double zero = 0.0;

TEST(UtestShell, compareDoubles)
{
    double not_a_number = zero / zero;
    double infinity = 1 / zero;
    CHECK(doubles_equal(1.0, 1.001, 0.01));
    CHECK(!doubles_equal(not_a_number, 1.001, 0.01));
    CHECK(!doubles_equal(1.0, not_a_number, 0.01));
    CHECK(!doubles_equal(1.0, 1.001, not_a_number));
    CHECK(!doubles_equal(1.0, 1.1, 0.05));
    CHECK(!doubles_equal(infinity, 1.0, 0.01));
    CHECK(!doubles_equal(1.0, infinity, 0.01));
    CHECK(doubles_equal(1.0, -1.0, infinity));
    CHECK(doubles_equal(infinity, infinity, 0.01));
    CHECK(doubles_equal(infinity, infinity, infinity));
    double a = 1.2345678;
    CHECK(doubles_equal(a, a, 0.000000001));
}
```
How should we remove this one test when compiling with Borland v5.4?

This is part of #1493

pjshelton:
Split TEST(UtestShell, compareDoubles) into

- TEST(UtestShell, compareDoubles)  for finite only tests,
- TEST(UtestShell, compareDoublesNaN) for tests against NaN and 
- TEST(UtestShell, compareDoublesInf) for tests against Inf. 

These last two are guarded by CPPUTEST_HAS_NAN and CPPUTEST_HAS_INF defines respectively

This closes #1545 
and is part of #1493


pjshelton:
How many build tool chains do I have to make this work in?

my count is four?
CMake, MSVC, configure, and MakefileWorker?

basvodde:
None, as long as you ensure the default is on. Then you can keep Borland for CMake.

You can do that by adding it to the CppUTestConfig.h. There are some other examples there you can look at. No need to do the compiler check (as you did that already in CMake) unless you want to. But it will need to check whether the CPPUTEST_HAS_XXX is set and, if not, set it to 1. Then change #ifdef to #if... The #if is much better as it will give a compiler error when not set and that leads to less accidental errors.

pjshelton:
So we actually want CppUTestGeneratedConfig.h to define a CPPUTEST_HASNOT_XXX when the build tool is CMake and the compiler is Borland.  That way CppUTestConfig.h can then define a CPPUTEST_HAS_XXX to be 0 if CPPUTEST_HASNOT_XXX exists and 1 otherwise

pjshelton:
Went with CPPUTEST_NO_NAN and CPPUTEST_NO_INF.
code now sets CPPUTEST_HAS_INF and CPPUTEST_HAS_NAN to 1 unless CPPUTEST_NO_INF or CPPUTEST_NO_NAN are defined respectively. 
And currently only CMake with Borland compiler define those two flags.

basvodde:
Ok, now it ought to work on all platforms without changing the build config files.

PR #1559, opened 2022-03-28
Scope cmake properties to targets

thetic:


examples/AllTests/CMakeLists.txt
```diff
@@ -10,6 +8,11 @@ add_executable(ExampleTests
     PrinterTest.cpp
 )
 
+target_include_directories(ExampleTests
```
thetic, line 11: This should be deleted.

PR #1566, opened 2022-04-02
Don't hijack the user's build config

thetic:
- When used as a subproject, don't assign `CMAKE_BUILD_CONFIG`. The default (i.e. empty) build type is valid and should not be overridden.
- Avoid setting a build type when using a [multi-configuration generator](https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html) in which case there is not a single build config.
- Print a message so that users know that a default build type has been assigned implicitly.


thetic:
This has caused me problems, In instances where a project uses the default build config and sets optimization or debug settings (e.g. `-Og -g3`), CppUTest's hijacking of `CMAKE_BUILD_CONFIG` appends its own flags (e.g. `-O2 -g`) that override the project's settings.

PR #1567, opened 2022-04-02
Backport CMake option fix when available

thetic:
Since [CMake v3.13](https://cmake.org/cmake/help/v3.13/release/3.13.html#other-changes), CMake has been able to correctly honor existing variables when interpretting `option()`.  The current minimum version (3.1 from 2014) suppresses this behavior regardless of the version of CMake used.

This change adopts this CMake feature when using a new enough version of CMake without breaking compatibility with older versions. This enables the expected behavior in a use case like this which didn't work before.

```cmake
cmake_minimum_required(VERSION 3.13)

project(someconsumingproject)

set(CPPUTEST_FLAGS OFF) # This will now be honored.
add_subdirectory(clone/of/cpputest)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ if (EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
     conan_basic_setup()
 endif()
 
+# Backport option handling fix from CMake 3.13 if available.
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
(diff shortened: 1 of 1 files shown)


thetic:
This change raises the issue the _very_ generic option names in the project (eg `TESTS`, `EXAMPLES`). CMake options are typically prefixed with something to act like a namespace. I think we'd be better of changing these to `CPPUTEST_TESTS`, `CPPUTEST_EXAMPLES`, etc. but that would be a breaking change for existing CMake configurations.

PR #1568, opened 2022-04-03
Normalize `find_package` and `add_subdirectory` usage

thetic:
Add `ALIAS` targets so that usage with `add_subdirectory` or `FetchContent` is the same as with `find_package`.


thetic:
Superceded by #1596

PR #1569, opened 2022-04-11
Add missing options to printed help text

jgonzalezdr:
Adds to usage and help the options that are missing, aligns indentation of command explanations, and homogenizes the nomenclature for parameter "placeholders" (i.e., non-literal texts) and optional parameters to commands.


basvodde:
Looks good!

PR #1577, opened 2022-05-07
Detect `long long` support in CMake

thetic:
https://cmake.org/cmake/help/latest/module/CheckTypeSize.html


basvodde:
Excellent, seems the automake variant for this one was done already!

Issue #1571, opened 2022-04-13
`CPPUTEST_HAVE_FENV` cannot be overridden


thetic:
CppUTest tries to deduce `CPPUTEST_HAVE_FENV` and `CPPUTEST_FENV_IS_WORKING` from what compiler is used, but this is sometimes incorrect and depends on explicit support for each compiler. For example, the ARM embedded GCC compiler may not have fenv.h, but all GCC builds with `CPPUTEST_USE_STD_C_LIB` defined are assumed to have it. I think this should be an explicit setting that may still be inferred when unset, but the user can predefine.

basvodde:
And then the checking can be moved to the autoconf file and the cmake file ?

thetic:
that's my thought. I know nothing about autconf, but it's a pretty straight-forward check in CMake: https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html

jjungo:
I'm having the same issue on my projects. My (very ugly) workaround is to `git apply` a patch after having fetched through `FetchContent_Declare` where all `IEEE754` related source files are remove from the build.
I would definitely like to have a cleaner way to do that.

PR #1578, opened 2022-05-07
Detect fenv.h in CMake

thetic:
Addresses #1571 (at least for CMake)


basvodde:
Good. It should also be done in the automake build but it is ok to be done later by someone else (like me). Thanks!

PR #1579, opened 2022-05-07
Detect _absence_ of fenv.h in CMake

thetic:
I realized I was only avoiding redundant checks when fenv.h is present when I really wanted to honor the _absence_ of fenv.h in cases that previously assumed its presence. (#1571 )

> Input file lines of the form “#cmakedefine01 VAR” will be replaced with either “#define VAR 1” or “#define VAR 0” [depending on whether VAR is set in CMake to any value not considered a false constant].

https://cmake.org/cmake/help/v3.1/command/configure_file.html


thetic:
This can also be overridden by users by passing `-DCPPUTEST_HAVE_FENV=<value>` on the command line or with `set(CPPUTEST_HAVE_FENV <value> CACHE BOOL "Enable IEEE754 extension")` before including CppUTest.

basvodde:
In the past, we've always used a construct like this:

/* Do we use Standard C or not? When doing Kernel development, standard C usage is out. */
#ifndef CPPUTEST_USE_STD_C_LIB
 #ifdef CPPUTEST_STD_C_LIB_DISABLED
  #define CPPUTEST_USE_STD_C_LIB 0
 #else
  #define CPPUTEST_USE_STD_C_LIB 1
 #endif
#endif

We can set the variables either thought a value from outside or through a DISABLED define. In the code, we prefer to use #if instead of #ifdef as the #if catches more cases (it fails when not set).

Perhaps that should be done for FENV also ?

thetic:
I'm not sure I follow. This change switches from `#ifdef` to `#if`. The only remaining `#ifdef` is [here](https://github.com/cpputest/cpputest/pull/1579/files#diff-168ea8f21c7bc702090b8a90bbf94dd495f14956503fbafb7fff8e09933b377aR214) where I do what you describe: if a value isn't provided one is deduced.

basvodde:
I was thinking whether we can make this #if block consistent with the other #if by having it the same way and having a DISABLED flag, then you can pass the DISABLED flag. Maybe also move the compiler checking then out of the #if block and let that set the DISABLED flag as the compiler checking should eventually go away as it ought to be done by CMake or automake

thetic:
gotcha. that makes sense. thanks for the explanation.

thetic:
Nevermind. I'm still confused. It sounds like you want me to add a DISABLE macro used to set the HAS macro when it isn't set? Why not let the user pass the HAS macro directly (as in my current implementation)?

basvodde:
My comment was mostly to make it more consistent with the other #if blocks in the CppUTestConfig.h (like the one I copy pasted related to Standard C. In this way, you CAN set the CPPUTEST_USE_FENV via the CMake/Automake file but you can also pass a -DCPPUTEST_FENV_DISABLED flag (which might be used by the Makefile templates that also still exist).

Then the second suggestion was to then pull out the compiler checking and put it before the #if block that would look the same as the other #if blocks. The compiler checking will eventually need to go because auto-detecting via CMake/Automake is likely the way to go. The compiler checks could then set the CPPUTEST_FENV_DISABLED so that it will be picked up by the #if block... when it wasn't set yet.

If you think it is a bad idea (I'd be interested to hear why) or rather just merge as is, that is also ok. I'll probably refactor it later sometime when doing a consistency check on the CppUTestConfig.h

thetic:
I think I've made the change you're asking for. Let me know what you think.

basvodde:
Yes, exactly. Now the compiler checking is already separated and can be perhaps thrown away at some point. And the #if is consistent with how it is done in other places. Thanks!

basvodde:
I will likely take a bit more time on the other CMake commits. While I wasn't paying attention, the TravisCI build stopped because we used all the credits (stupid company!). Now I understand also why earlier we merged some PRs that should have been caught by the automated build. So, I want to try to migrate that asap to something else so that your changes are covered by the automated build.

thetic:
How's the CI effort going? Is there anything I could do to help?

basvodde:
Oh, this one needs an update also still due to conflict. I think I'm ready to merge it.

PR #1580, opened 2022-05-07
DNM: Reorganize CMake configuration to permit safe usage in user project

thetic:
- Make `CPP_PLATFORM` a proper cache variable in the CMake configuration. When using an interactive interface like `ccmake` or `cmake-gui`, this allows us to present users with a list of supported platforms. Set `CPPUTEST_PLATFORM` to something falsey (e.g. `FALSE`, `OFF`) to provide your own implementation. (Spun out to https://github.com/cpputest/cpputest/pull/1586)
- Refactor the `target_compile_definitions` commands called after checking for functionality to instead populate the generated header, which accomplishes in a couple lines what was taking 8 before without any conditionals. (Spun out to https://github.com/cpputest/cpputest/pull/1583)
- Upgrade minimum CMake version to leverage native IAR support [introduced in 3.10](https://cmake.org/cmake/help/latest/release/3.10.html#platforms).
- Disable options that aren't supported for the given configuration instead of overriding them later, disregarding the user's config (Addressed in #1620).
- Leverage native project versioning support. (Spun out to https://github.com/cpputest/cpputest/pull/1587)
- `VERBOSE_CONFIG`, `TESTS`, and `EXAMPLES` are now enabled by default when CppUTest is the root project and disabled by defaults when included in a parent project. This remains configurable.
- The `C++11` variable is renamed to `CPPUTEST_USE_CXX11` for compatibility with `cmake_dependent_option` and adherence with [CMake documentation's guidance](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variables).(Removed in https://github.com/cpputest/cpputest/pull/1588)
- Features that depend on manipulating global settings are disabled when `CPPUTEST_FLAGS` is off: `MEMORY_LEAK_DETECTION`, `MAP_FILE`, `COVERAGE`, `CPPUTEST_USE_CXX11`, `WERROR`, `STDC_WANT_SECURE_LIB`.

And finally, the most consequential change: **`CPPUTEST_FLAGS` is now disabled by default when used as a subproject** (ie with `add_subdirectory` or `FetchContent`). Previously merely calling `add_subdirectory(cpputest)` in a project would contaminate *every build command in that project*. CppUTest's CMake would change:

- `CMAKE_BUILD_TYPE` (Addressed in https://github.com/cpputest/cpputest/pull/1566)
- `CMAKE_CXX_FLAGS`
- `CMAKE_C_FLAGS`
- `CMAKE_EXE_LINKER_FLAGS`
- include paths
- macro definitions
- `CMAKE_CXX_EXTENSIONS`

This made it all but impossible to safely use alongside production code. The sum of these changes should make it safe by default to use CppUTest in any project without contaminating users' projects. Users may of course choose to reenable any of these features explicitly.


CMakeLists.txt
```diff
@@ -16,20 +17,59 @@ if (EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
     conan_basic_setup()
 endif()
 
-option(STD_C "Use the standard C library" ON)
+if(NOT DEFINED CPP_PLATFORM)
+    if(MSVC)
+        set(CPP_PLATFORM VisualCpp)
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "IAR")
```
thetic, line 23: I don't know if/how `if(IAR)` ever worked before. It has never been [provided by CMake](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html) as far as I can tell. CMake didn't learn how to [identify an IAR toolchain](https://cmake.org/cmake/help/v3.10/variable/CMAKE_LANG_COMPILER_ID.html?highlight=iar) until [version 3.10](https://cmake.org/cmake/help/latest/release/3.10.html).

thetic: I found it. It was defined in the ewarm toolchain files.

thetic:
I know this is a lot at once, and I would be happy to break this up a series of incremental MRs, but I wanted to show what I'm driving towards and get a conversation started. My main concern is the following use case:

```cmake
project(user_project)
include(CTest)

# Production code
# CppUTest should have *no* affect on how this builds
add_executable(user_program #[[ ... ]])
add_library(user_library #[[ ... ]])
target_link_libraries(user_program PRIVATE user_library)

# Test code
# Turning this on or off should have *no* affect on production code
if(BUILD_TESTING)
    add_subdirectory(cpputest)
    add_executable(user_test #[[ ... ]])
    target_link_libraries(user_test PRIVATE CppUTest CppUTestExt)
    add_test(NAME user_test COMMAND user_test)
endif()
```

basvodde:
I would prefer smaller PRs. I'm not a CMake expert, I believe most of the CMake code is contributed by @offa . I do review the code to see if I understand what is going on, but this PR is a bit large now

thetic:
I definitely don't mind splitting this up. Are you OK with the direction generally?

Are you comfortable updating the minimum required CMake version for native IAR support?

Is it acceptable to change the existing `option` names? If we decide to change one option name, should we change them all at once to avoid churn? There is a high potential for name collisions with the current names. (Cache variables are typically prefixed with a project "namespace", e.g. `CPPUTEST_*`.

I am confident that these changes will make it easier to integrate CppUTest into existing CMake projects, but I would definitely appreciate any feedback.

offa:
> I believe most of the CMake code is contributed by @offa . 

Some contributions only :-)

> Is it acceptable to change the existing option names? If we decide to change one option name, should we change them all at once to avoid churn? There is a high potential for name collisions with the current names. (Cache variables are typically prefixed with a project "namespace", e.g. CPPUTEST_*.)

Makes sense :+1: . But since this is a breaking change, it might be good to consider a migration path.

CMakeLists.txt
```diff
@@ -16,34 +24,80 @@ if (EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
...
+
+option(EXAMPLES "Compile and make examples?" ${PROJECT_IS_TOP_LEVEL})
+
+option(VERBOSE_CONFIG "Print configuration to stdout during generation" ${PROJECT_IS_TOP_LEVEL})
+
+cmake_dependent_option(LIBNAME_POSTFIX_BITSIZE "Add architecture bitsize (32/64) to the library name?" OFF "PROJECT_IS_TOP_LEVEL" OFF)
+cmake_dependent_option(LIBNAME_POSTFIX_DEBUG "Add indication of debug compilation to the library name?" OFF "PROJECT_IS_TOP_LEVEL" OFF)
+
+cmake_dependent_option(HAS_INF "Compiler has Inf value for float" ON "NOT BORLAND" OFF)
+cmake_dependent_option(HAS_NAN "Compiler has NaN value for float" ON "NOT BORLAND" OFF)
```
offa, line 55: _(Not directly related to this PR)_ Does it make sense to let CMake detect these?

cmake/Modules/CppUTestGlobalFlags.cmake
```diff
@@ -108,7 +85,7 @@ set(GMOCK_HOME $ENV{GMOCK_HOME})
 if (DEFINED ENV{GMOCK_HOME})
     # GMock pulls in gtest.
     set(CPPUTEST_INCLUDE_GTEST_TESTS 1)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGTEST_USE_OWN_TR1_TUPLE=1")
+    set(CPPUTEST_CXX_FLAGS "${CPPUTEST_CXX_FLAGS} -DGTEST_USE_OWN_TR1_TUPLE=1")
```
offa, line 88: Suppoort for `GTEST_USE_OWN_TR1_TUPLE` has been removed from GTest some years ago. Should we consider removing it (does not harm though)? @basvodde

offa (approved):

CMakeLists.txt
thetic: Probably?
- https://en.cppreference.com/w/cpp/numeric/math/NAN
- https://en.cppreference.com/w/cpp/numeric/math/INFINITY

offa: That's C++11 (or later) only :-(

cmake/Modules/CppUTestGlobalFlags.cmake
thetic: https://github.com/google/googletest/commit/7d3b73c85a42811309eac26e5cbe054c40b64785

CMakeLists.txt
basvodde: The reason port to Borland C++ didn't supported NaN and INF properly, so I assume it is still needed.

cmake/Modules/CppUTestGlobalFlags.cmake
basvodde: Right now, I don't think CppUtest actually supports the latest version of GTest/GMock. The ones that are supported are also in the travis build and I think we want to keep supporting these. Thus then the define is still needed.

CMakeLists.txt
thetic: https://github.com/cpputest/cpputest/pull/1584

thetic: https://github.com/cpputest/cpputest/pull/1585

thetic:
> I would prefer smaller PRs.

https://github.com/cpputest/cpputest/pull/1586

thetic:
@offa @basvodde Do you know the intention of the `C++11` option. Why does this do that `CMAKE_CXX_STANDARD` doesn't? It seems like a lot of complexity that duplicates standard CMake behavior.

offa:
It guess it was used some time ago to decide whether C++11 or C++98 should be used (before the CMake variable was introduced). If I'm right, it's kept for compatibility reasons only and finally boils down to the CMake variable.

PR #1585, opened 2022-05-14
Check for infinity in CMake

thetic:
This deprecates the `HAS_INF` cache variable that sets the `CPPUTEST_NO_INF` variable that sets the `CPPUTEST_NO_INF` macro that sets the `CPPUTEST_HAS_INF` macro. Instead we use CMake to check for infinity support explicitly and use the result to define a `CPPUTEST_HAS_INF` macro scoped to the tests.

See:
- https://en.cppreference.com/w/cpp/types/numeric_limits/has_infinity
- https://cmake.org/cmake/help/latest/policy/CMP0067.html


offa (approved):

basvodde:
Unfortunately, this PR made the DOS build fail

thetic:
I suggest rolling this back. I can reintegrate with all the updated master and resubmit.

thetic:
https://github.com/cpputest/cpputest/pull/1618

PR #1586, opened 2022-05-14
Create CPPUTEST_PLATFORM cache variable

thetic:
This promotes platform implementation selection to a proper cache
variable.  This ensures that the selection is persistent and avoids
issues where a user's selection could be later overriden.  It also adds
the benefit of presenting a list of supported options when used with an
interactive interface like `ccmake` or `cmake-gui`.

Users can set `CPPUTEST_PLATFORM` to something falsey like `OFF` or
`FALSE` to avoid use of a provided implementation.

The old `CPP_PLATFORM` variable is still honored when determining a
default value, but raises a deprecation warning when provided by the
user.


offa (approved):

thetic:
A bit of a tangent; how is support for `-nostdlib` validated?

PR #1587, opened 2022-05-14
Leverage native project versioning

thetic:
https://cmake.org/cmake/help/latest/command/project.html


offa (approved):
Nice catch! :laughing:

PR #1588, opened 2022-05-14
Remove unecessary custom C++11 logic

thetic:
Users can/should just set `CMAKE_CXX_STANDARD`. This option just complicated and overlapped native CMake features.


thetic:
@martiert

cmake/Modules/CppUTestConfigurationOptions.cmake
```diff
@@ -95,13 +95,18 @@ if (COVERAGE AND NOT MSVC)
         )
 endif()
 
+if(DEFINED C++11)
```
offa, line 98: :+1:

thetic:
I've demonstrated backwards compatibility in the [cmake_coverage build.](https://github.com/cpputest/cpputest/runs/7265204469?check_suite_focus=true)

PR #1589, opened 2022-05-16
Hello World CircleCI file. Looking for replacing TravisCI.

basvodde:
The Travis 'free' build time was quickly spend and has not build for 6 months. Such a crap. Never sell your company and fuck over open source developers

### DIFF
--- a/.circleci/config.yaml
+++ b/.circleci/config.yaml
@@ -1,0 +1,16 @@
+version: 2.1
+
+jobs:
+  say-hello:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - checkout
+      - run:
+          name: "Say hello"
+          command: "echo Hello, World!"
+
+workflows:
+  say-hello-workflow:
+    jobs:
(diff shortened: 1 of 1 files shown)


basvodde:
Checking some comparison (from marketing material): https://circleci.com/circleci-versus-github-actions/

My initial reaction is still to keep trying CircleCI. Especially the multiple different machines and architectures is something that we used to use in TravisCI. Did you ever see the travis_ci yaml with all the different build configs that we had there?

offa:
> My initial reaction is still to keep trying CircleCI. 

No objections, GH Actions was just a suggestion – if Circle CI suits better, let's go for it! :+1: 

> Especially the multiple different machines and architectures is something that we used to use in TravisCI.

I haven't used other architectures myself or looked for it. As far as I know it's nothing supported out of the box, but available through actions. But really no experience from my side.

> Did you ever see the travis_ci yaml with all the different build configs that we had there?

Yes :-)

PR #1597, opened 2022-05-23
Support test discovery when cross-compiling

thetic:
This extends support for discovering and running tests to cross-compiling setups.

Tested with mingw+wine on Linux.


cmake/Modules/CppUTestBuildTimeDiscoverTests.cmake
```diff
@@ -1,19 +1,17 @@
...
-function (cpputest_buildtime_discover_tests EXECUTABLE)
-  # The path to the discover script depends on execution mode:
-  # - internal (building CppUTest it self).
-  # - imported (installed, imported, and executed by a client of the CppUTest lib)
-  if (PROJECT_NAME STREQUAL "CppUTest") # internal - (path is relative to source dir)
-    SET(DISCOVER_SCRIPT ${PROJECT_SOURCE_DIR}/cmake/Scripts/CppUTestBuildTimeDiscoverTests.cmake)
-  else (PROJECT_NAME STREQUAL "CppUTest") # Installed (path is relative to install directory)
-    SET(DISCOVER_SCRIPT ${CppUTest_DIR}/Scripts/CppUTestBuildTimeDiscoverTests.cmake)
-  endif (PROJECT_NAME STREQUAL "CppUTest")
+set(_DISCOVER_SCRIPT "${CMAKE_CURRENT_LIST_DIR}/../Scripts/CppUTestBuildTimeDiscoverTests.cmake")
```
thetic, line 1: The location of the script relative to this module is the same between the "internal" and "imported" cases.

PR #1600, opened 2022-05-24
Fix GCC 12 compilation

offa:
GCC 12 reports an use-after-free warning for two test cases.


basvodde:
Yes, I still live :) I'll be checking the too much pull requests in the coming months and fix the CI builds

PR #1601, opened 2022-05-24
Fix compilation on clang

offa:
Fixes some warnings reported by newer versions of Clang.

### DIFF
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -41,6 +41,7 @@
  */
 
 #ifdef __clang__
+ #define NEEDS_DISABLE_UNREACHABLE_CODE 1
  #pragma clang diagnostic push
  #if (__clang_major__ == 3 && __clang_minor__ >= 6) || __clang_major__ >= 4
   #pragma clang diagnostic ignored "-Wreserved-id-macro"
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -103,10 +103,12 @@
   #define __has_attribute(x) 0
 #endif
 
-#if __has_attribute(noreturn)
-  #define _no_return_ __attribute__((noreturn))
+#if defined (__cplusplus) && __cplusplus >= 201103L
+   #define _no_return_ [[noreturn]]
+#elif __has_attribute(noreturn)
+   #define _no_return_ __attribute__((noreturn))
 #else
-  #define _no_return_
+   #define _no_return_
 #endif
 
 #if defined(__MINGW32__)
@@ -195,7 +197,7 @@
 #define CPPUTEST_SANITIZE_ADDRESS 1
 #define CPPUTEST_DO_NOT_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
   #if defined(__linux__) && defined(__clang__)
-    #if CPPUTEST_USE_MEM_LEAK_DETECTION 
+    #if CPPUTEST_USE_MEM_LEAK_DETECTION
     #warning Compiling with Address Sanitizer with clang on linux will cause duplicate symbols for operator new. Turning off memory leak detection. Compile with -DCPPUTEST_MEM_LEAK_DETECTION_DISABLED to get rid of this warning.
     #undef CPPUTEST_USE_MEM_LEAK_DETECTION
     #define CPPUTEST_USE_MEM_LEAK_DETECTION 0
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -41,7 +41,6 @@
  */
 
 #ifdef __clang__
- #define NEEDS_DISABLE_UNREACHABLE_CODE 1
  #pragma clang diagnostic push
  #if (__clang_major__ == 3 && __clang_minor__ >= 6) || __clang_major__ >= 4
   #pragma clang diagnostic ignored "-Wreserved-id-macro"
--- a/tests/CppUTest/TestFailureTest.cpp
+++ b/tests/CppUTest/TestFailureTest.cpp
@@ -407,7 +407,7 @@ TEST(TestFailure, BitsEqualChar)
     BitsEqualFailure f(test, failFileName, failLineNumber, 0x01, 0x03, 0xFF, sizeof(char), "");
     FAILURE_EQUAL("expected <00000001>\n\tbut was  <00000011>", f);
 }
-#endif /* CPPUTEST_CHAR_BIT */
+#endif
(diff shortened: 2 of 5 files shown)


offa:
ping @basvodde

tests/CppUTest/TestFailureTest.cpp
```diff
@@ -442,4 +442,4 @@ TEST(TestFailure, UnexpectedExceptionFailure_StandardException)
     STRCMP_CONTAINS("runtime_error", f.getMessage().asCharString());
     STRCMP_CONTAINS("' was thrown: Some error", f.getMessage().asCharString());
 }
-#endif
\ No newline at end of file
+#endif /* CPPUTEST_USE_STD_CPP_LIB */
```
basvodde, line 445: I'll merge these, but AFAIK this isn't consistently used and I would myself not prefer comments at the end of the #endif.

tests/CppUTest/UtestTest.cpp
```diff
@@ -30,6 +30,10 @@
 #include "CppUTest/TestTestingFixture.h"
 #include "CppUTest/PlatformSpecificFunctions.h"
 
+#ifdef __clang__
+# define NEEDS_DISABLE_UNREACHABLE_CODE
```
basvodde, line 34: Should this be in the CppUTestConfig instead ? Now it feels a bit like a hack here ?

tests/CppUTest/UtestTest.cpp
```diff
@@ -218,12 +222,26 @@ TEST(UtestShell, TestStopsAfterSetupFailure)
 }
 
 #if CPPUTEST_USE_STD_CPP_LIB
-static void thrownUnknownExceptionMethod_()
+#ifdef NEEDS_DISABLE_UNREACHABLE_CODE
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wunreachable-code"
```
basvodde, line 227: Ouch.

Could we wrap the throw in an "if" instead, which is always true in the test case? It wouldn't make the test much better but it would prevent us to have #defines and #pragma in the code.

tests/CppUTest/UtestTest.cpp
```diff
@@ -218,12 +222,26 @@ TEST(UtestShell, TestStopsAfterSetupFailure)
...
+#endif /* NEEDS_DISABLE_UNREACHABLE_CODE */
+_no_return_ static void thrownUnknownExceptionMethod_()
 {
     throw 33;
```
basvodde, line 231: if (shouldAlwaysThrow) throw 33; ?

tests/CppUTest/TestFailureTest.cpp
offa: No problem, I'm going to remove them :+1:

tests/CppUTest/UtestTest.cpp
offa: Indeed …
It shouldn't be needed at any other place, but wont hurt in TestConfig.

tests/CppUTest/UtestTest.cpp
offa: I'll give it a try.

Generally compilers are quite good in preventing these kind of issues now.

tests/CppUTest/UtestTest.cpp
```diff
@@ -225,7 +225,7 @@ TEST(UtestShell, TestStopsAfterSetupFailure)
 #ifdef NEEDS_DISABLE_UNREACHABLE_CODE
 # pragma GCC diagnostic push
 # pragma GCC diagnostic ignored "-Wunreachable-code"
-#endif /* NEEDS_DISABLE_UNREACHABLE_CODE */
```
basvodde: What about the if to make it reachable ?

tests/CppUTest/UtestTest.cpp
offa: A simple *if* condition caused an "always-true" warning, but a little more trickery solved it.

tests/CppUTest/UtestTest.cpp
offa: *(as above)*

basvodde:
Sorry, I had a family emergency and couldn't wrap this up. I'll add some more comments as it isn't yet what I had in mind.

include/CppUTest/CppUTestConfig.h
```diff
@@ -41,6 +41,7 @@
  */
 
 #ifdef __clang__
+ #define NEEDS_DISABLE_UNREACHABLE_CODE 1
```
basvodde, line 44: This one can be deleted again, right? It isn't used now?

tests/CppUTest/UtestTest.cpp
```diff
@@ -218,9 +218,34 @@ TEST(UtestShell, TestStopsAfterSetupFailure)
...
+    if (avoidUnreachableCodeWarning())
+    {
+        throw 33;
+    }
+    stopAfterFailure++;
+}
+
+static void thrownStandardExceptionMethod_()
+{
+    if (avoidUnreachableCodeWarning())
```
basvodde, line 245: Nice job on tricking the compiler! :)

What I was thinking about was to not trick the compiler but to put a legitimate if there. It would be on a static-scoped boolean. We could then also add another test that actually sets the boolean to false thus not throwing. That way the unreachable code will actually be reached... and it will also not lead to non-coverage issues.

I guess it will be safer than the compiler trick as there will be a time the compiler (or another one) to notice the trick :)

include/CppUTest/CppUTestConfig.h
offa: Right!

tests/CppUTest/UtestTest.cpp
offa: Using a non-const static seems enough. :+1:

basvodde: Yeah, non-const will export it as a linker symbol though, which isn't very nice. I'd prefer a const one and then having one tests where it is set to false. That will then also cover the 'unreachable' lines in the test coverage measurements, which I think it probably better. And then I will really merge it and move on to the next one!

offa: > I'd prefer a const one and then having one tests where it is set to false.

Changing the value of a `const` wont work (or do I get this wrong?).

basvodde: Yeah, so I would suggest to not make it const :)

Perhaps call it "shouldThrowException" and set it to true by default.

offa: Done

basvodde:
Thanks

PR #1603, opened 2022-05-25
stop creating temp files that aren't cleaned up

mtfurlan:
When I made 7fc2a5e18e5567df67a135d6a7fe6311f92a9660 I stored the
return code from the test binary in a temp file and then deleted the
temp file.
But that temp file was created every time MakefileWorker.mk was sourced,
not just when the targets were run.

This commit changes it so the temp file is only created if the relevant
targets are being run so it will always get deleted.

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -520,13 +520,14 @@ else
 endif
 	$(SILENCE)$(RANLIB) $@
 
-TEST_RUN_RETURN_CODE_FILE:=$(shell mktemp /tmp/cpputestResult.XXX)
 test: $(TEST_TARGET)
-	($(RUN_TEST_TARGET); echo $$? > $(TEST_RUN_RETURN_CODE_FILE)) | tee $(TEST_OUTPUT)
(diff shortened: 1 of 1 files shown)


mtfurlan:
It works now.
Should we hide the command being run?
I don't think seeing 
```
$ make test
(  rm -f .cpputestBuild/objs/../src/lib.gcda .cpputestBuild/objs/./testrunner.gcda .cpputestBuild/objs/./tests.gcda gcov_output.txt gcov_report.txt gcov_error.txt ; echo "Running example_tests"; ./example_tests -c; echo $? > /tmp/cpputestResult.vY0) | tee example_tests.txt
Running example_tests
.
OK (1 tests, 1 ran, 1 checks, 0 ignored, 0 filtered out, 0 ms)
```
vs
```
$ make test
Running example_tests
.
OK (1 tests, 1 ran, 1 checks, 0 ignored, 0 filtered out, 0 ms)
```
I think it's better without, so I hid it in this PR but am happy to change.

PR #1604, opened 2022-06-03
Ignore apple clang warnings properly

mtfurlan:
Apple clang is different than normal clang for reasons.
We need to ignore some stuff specifically for apple clang.

I did some testing with https://github.com/sickcodes/Docker-OSX to
figure out what versions of apple clang need what warnings surpressed.

None of these warnings happen with normal clang.
```
10.15: "Catalina"
 Apple clang version 11.0.3 (clang-1103.0.32.62)  Xcode_11.5
   No extra warning suppression required
 Apple clang version 12.0.0 (clang-1200.0.32.2)   Xcode_12
 Apple clang version 12.0.0 (clang-1200.0.32.27)  Xcode_12.2
 Apple clang version 12.0.0 (clang-1200.0.32.28)  Xcode_12.3
 Apple clang version 12.0.0 (clang-1200.0.32.29)  Xcode_12.4
   poison-system-directories
11: "Big Sur"
 Apple clang version 12.0.5 (clang-1205.0.22.9)   Xcode_12.5
 Apple clang version 12.0.5 (clang-1205.0.22.11)  Xcode 12.5.1
   suggest-override
   suggest-destructor-override
   poison-system-directories
12: "Monterey"
 Apple clang version 13.1.6 (clang-1316.0.21.2.5) Xcode ???
   suggest-override
   suggest-destructor-override
```

I tested this change on all three versions of macos with one or two versions of clang and it worked fine on all of them.

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -214,6 +214,7 @@ CLANG_VERSION := $(shell echo $(CC_VERSION_OUTPUT) | sed -n 's/.* \([0-9]*\.[0-9
 CLANG_VERSION_NUM := $(subst .,,$(CLANG_VERSION))
 CLANG_VERSION_NUM_GT_700 := $(shell [ "$(CLANG_VERSION_NUM)" -ge 700 ] && echo Y || echo N)
 CLANG_VERSION_NUM_GT_1200 := $(shell [ "$(CLANG_VERSION_NUM)" -ge 1200 ] && echo Y || echo N)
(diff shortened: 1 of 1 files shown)


thetic:
Are these warnings related to #1605?

mtfurlan:
@thetic Are you running apple clang?
If so I bet it's the same things and I could remove -Wno-suggest-override from this PR and that would be cool.

This PR is specifically addressing flags that don't cause warnings with clang, but do with apple clang.
Apple provides a modified llvm/clang with xcode, and of course they don't document what they changed anywhere.

thetic:
It looks like Apple Clang supports `override` even in C++98. See #1612.

PR #1605, opened 2022-06-22
Resolve malformed override specifiers with function pointers

thetic:
These raised warnings with [`-Wsuggest-override`](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-Wsuggest-override). I personally find this easier to read as well.

### DIFF
--- a/examples/AllTests/EventDispatcherTest.cpp
+++ b/examples/AllTests/EventDispatcherTest.cpp
@@ -32,7 +32,7 @@
 class ObserverMock : public EventObserver
 {
 public:
-    virtual void notify(const Event& event, int timeOutInSeconds)
+    virtual void notify(const Event& event, int timeOutInSeconds) _override
     {
         mock()
             .actualCall("notify")
@@ -40,7 +40,7 @@ public:
             .withParameterOfType("Event", "event", (void*)&event)
             .withParameter("timeOutInSeconds", timeOutInSeconds);
     }
-    virtual void notifyRegistration(EventObserver* newObserver)
+    virtual void notifyRegistration(EventObserver* newObserver) _override
     {
         mock().actualCall("notifyRegistration").onObject(this).withParameter("newObserver", newObserver);
     }
@@ -49,11 +49,11 @@ public:
 class EventComparator : public MockNamedValueComparator
 {
 public:
-    virtual bool isEqual(const void* object1, const void* object2)
+    virtual bool isEqual(const void* object1, const void* object2) _override
     {
         return ((const Event*)object1)->type == ((const Event*)object2)->type;
     }
-    virtual SimpleString valueToString(const void* object)
+    virtual SimpleString valueToString(const void* object) _override
     {
         return StringFrom(((const Event*)object)->type);
     }
@@ -67,12 +67,12 @@ TEST_GROUP(EventDispatcher)
     ObserverMock observer2;
     EventComparator eventComparator;
 
-    void setup()
+    void setup() _override
     {
         dispatcher = new EventDispatcher;
         mock().installComparator("Event", eventComparator);
     }
-    void teardown()
+    void teardown() _override
     {
         delete dispatcher;
         mock().removeAllComparatorsAndCopiers();
--- a/cmake/Modules/CppUTestWarningFlags.cmake
+++ b/cmake/Modules/CppUTestWarningFlags.cmake
@@ -64,12 +64,13 @@ else (MSVC)
 
     if (DEFINED CMAKE_CXX_STANDARD AND NOT CMAKE_CXX_STANDARD EQUAL 98)
         set(WARNING_CXX_FLAGS
-           ${WARNING_CXX_FLAGS}
-           Wno-c++98-compat
-           Wno-c++98-compat-pedantic
-           Wno-c++14-compat
-           Wno-inconsistent-missing-destructor-override
-           )
+            ${WARNING_CXX_FLAGS}
+            Wno-c++98-compat
+            Wno-c++98-compat-pedantic
+            Wno-c++14-compat
+            Wno-inconsistent-missing-destructor-override
+            Wsuggest-override
+        )
     endif ()
 
     check_and_append_c_warning_flags(${WARNING_C_FLAGS})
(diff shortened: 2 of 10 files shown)


offa (approved):

basvodde:
I don't agree with the typedef being easier to read, I personally prefer not using a typedef here. But I'm curious about the warnings that it resolved ?

thetic:
```
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:95:20: error: â€˜virtual void (* MockCheckedActualCall::returnFunctionPointerValue())()â€™ can be marked override [-Werror=suggest-override]
   95 |     virtual void (*returnFunctionPointerValue())() _override;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:96:20: error: â€˜virtual void (* MockCheckedActualCall::returnFunctionPointerValueOrDefault(void (*)()))()â€™ can be marked override [-Werror=suggest-override]
   96 |     virtual void (*returnFunctionPointerValueOrDefault(void (*)()))() _override;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:216:20: error: â€˜virtual void (* MockActualCallTrace::returnFunctionPointerValue())()â€™ can be marked override [-Werror=suggest-override]
  216 |     virtual void (*returnFunctionPointerValue())() _override;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:217:20: error: â€˜virtual void (* MockActualCallTrace::returnFunctionPointerValueOrDefault(void (*)()))()â€™ can be marked override [-Werror=suggest-override]
  217 |     virtual void (*returnFunctionPointerValueOrDefault(void (*)()))() _override;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:292:20: error: â€˜virtual void (* MockIgnoredActualCall::returnFunctionPointerValue())()â€™ can be marked override [-Werror=suggest-override]
  292 |     virtual void (*returnFunctionPointerValue())() _override { return NULLPTR; }
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/CppUTestExt/../../include/CppUTestExt/MockCheckedActualCall.h:293:20: error: â€˜virtual void (* MockIgnoredActualCall::returnFunctionPointerValueOrDefault(void (*)()))()â€™ can be marked override [-Werror=suggest-override]
  293 |     virtual void (*returnFunctionPointerValueOrDefault(void (*value)()))() _override { return value; }
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

See also https://github.com/cpputest/cpputest/pull/1604

basvodde:
Oh, that is strange. Is it because it makes it a pointer to member function? Lemme have a look at this later. Thanks for all your contributions, I'll gradually be merging most of them.

thetic:
Yeah, it's definitely strange. It _feels_ like a gcc bug, but I reproduces across multiple versions of both gcc and clang.

thetic:
Have you had a chance to look into this yet?

basvodde:
What do we do with the function pointer typedef? I really do not want to use a typedef for all the FPs, personally I find it harder to read as it hides so much info.

thetic:
https://isocpp.org/wiki/faq/pointers-to-members#typedef-for-ptr-to-memfn

basvodde:
"Yea, right, I know: you are different. You are smart. You can do this stuff without a typedef. Sigh. I have received many emails from people who, like you, refused to take the simple advice of this FAQ"

haha.

thetic:
> Yeah, it's definitely strange. It _feels_ like a gcc bug, but I reproduces across multiple versions of both gcc and clang.

I was wrong on this one.

GCC expects:
```cpp
void (*member() override)();
```

Clang expects:

```cpp
void (*member())() override;
```

basvodde:
Wow, and it is not standardized ?

To me the gcc expectation feels wrong...

examples/AllTests/EventDispatcherTest.cpp
```diff
@@ -67,12 +67,12 @@ TEST_GROUP(EventDispatcher)
     ObserverMock observer2;
     EventComparator eventComparator;
 
-    void setup()
+    TEST_SETUP()
```
basvodde, line 70: Could we keep the "void setup ()" instead? The "TEST_SETUP" macro is there mostly for backwards compatability as the "void setup" makes it clearer what is happening.

basvodde:
Considering the different syntax for gcc and clang, I guess I'll need to give up and go with a typedef.

Could we at least try a slightly more descriptive name as FunctionPointer and move the typedef outside of the class space so that it doesn't have to be declared on the class in the cpp file ?

thetic:
> Could we at least try a slightly more descriptive name as FunctionPointer and move the typedef outside of the class space so that it doesn't have to be declared on the class in the cpp file ?

I can't think of a good name that is more descriptive for `void (*)()`. I am open to suggestions.

My original intentions for putting the typedef in the class was to avoid naming collisions with client code.

examples/AllTests/EventDispatcherTest.cpp
thetic: sorry. lost track of this one. done.

basvodde:
Ok. could you refer to is as cpputest_functionPointerRetrunValue or cppUTestFunctionPointerReturnValue or FunctionPointerReturnValue.

The reason is that it seems it is only in this context that the typedef is used. Other function pointers that are passed as parameter are still written in the 'normal' form.

If you strongly disagree with it, then I can also merge it as is... the usage of the typedef is quite localized.

thetic:
OK. I've renamed the typedef to `FunctionPointerReturnValue` and reduced the change set to the bare minimum required to satisfy `-Wsuggest-override`.

basvodde:
Thanks!

PR #1607, opened 2022-07-08
GitHub actions

thetic:
This gets at least some of the old travis builds running again using GitHub actions. There's still a bunch of follow on work we'll need to take on after this:

- C++98 support is broken because newer features have been introduced in TestFailure.cpp, CompatiblityTests.cpp, SimpleStringTest.cpp for example.
- The old version of GTest/GMock that travis used seems to depend on python2 which is EOL.
- DOSBox has trouble running in GitHub Actions' terminal.
- There are a bunch of other configs from travis that need to migrate too.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -38,6 +38,18 @@ jobs:
           ./configure
           make tdd
 
+  mac-cmake:
+    runs-on: macos-latest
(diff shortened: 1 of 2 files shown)


offa (approved):

basvodde:
Thanks!

PR #1610, opened 2022-07-09
Automatically disable C++ standard library usage when exceptions are disabled

thetic:
Depends on #1620

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,8 +60,8 @@ cmake_dependent_option(CPPUTEST_BUILD_TESTING "Compile and make tests for CppUTe
 option(CPPUTEST_TESTS_DETAILED "Run each test separately instead of grouped?" OFF)
 cmake_dependent_option(CPPUTEST_TEST_DISCOVERY "Build time test discover"
   ON "CPPUTEST_BUILD_TESTING;CMAKE_CROSSCOMPILING_EMULATOR OR NOT CMAKE_CROSSCOMPILING" OFF)
-
-option(CPPUTEST_EXAMPLES "Compile and make examples?" ${PROJECT_IS_TOP_LEVEL})
+cmake_dependent_option(CPPUTEST_EXAMPLES "Compile and make examples?"
+  ${PROJECT_IS_TOP_LEVEL} "NOT CPPUTEST_STD_CPP_LIB_DISABLED" OFF)
 option(CPPUTEST_VERBOSE_CONFIG "Print configuration to stdout during generation" ${PROJECT_IS_TOP_LEVEL})
 
 cmake_dependent_option(CPPUTEST_LIBNAME_POSTFIX_BITSIZE "Add architecture bitsize (32/64) to the library name?"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,15 @@ include(CTest)
 
 include(CMakeDependentOption)
 option(CPPUTEST_STD_C_LIB_DISABLED "Disable the standard C library")
+
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+  "int main(int argc, char ** argv) { throw 20; }"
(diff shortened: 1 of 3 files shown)


basvodde:
So far, we've assumed that "CPPUTEST_USE_STD_CPP_LIB" also includes using exceptions and not created a separate #define for that. Is there a benefit of making a seperate #define for that?

thetic:
I think so. In some of my embedded use cases I disable exceptions, but still use the standard library:

- `#include <cstdint>`
- `#include <algorithm>`
- `#include <array>`

basvodde:
Does standard C++ library work without exceptions ?!?!

thetic:
I'm sure it varies between implementations, but in my experience many parts of the standard library (like the components I listed above) work without exceptions.

basvodde:
But can you compile them with the compiler set to not recognizing exceptions?

I mean, we can make this #define also explicit. So far, I had gone for trying to avoid that and include the no exceptions in no stdc++ lib. The more #defines, the more builds we'll end up and the harder the maintenance.

I'm just surprised about this one :) StdC++ Lib ought to be standard and according standard it uses exception handling... so was my thinking.

thetic:
This is my motivation:

Currently, this won't work:

```console
$ CXXFLAGS="-fno-exceptions" cmake # ...
```

unless we turn off the standard library:

```console
$ CXXFLAGS="-fno-exceptions" cmake -DSTD_CPP=OFF # ...
```

CMake can't reliably detect use of the C++ standard library in any practical way, but it can detect exception support as demonstrated in this diff. What if, instead of adding another macro, we automatically turn off `STD_CPP` when CMake detects that exceptions are disabled?

basvodde:
I would be more comfortable with that.

That said, if you are saying that -fno-exceptions actually works with StdC++ library, then I'm super surprised. Lemme check this a bit further.

basvodde:
I cannot find anything about it... but AFAIK exception handling is an integral part of the Std C++ library, so I would be much more comfortable to assume you don't use stdc++ when you turn off exception handling. So, lets do that

thetic:
> What if, instead of adding another macro, we automatically turn off `STD_CPP` when CMake detects that exceptions are disabled?

I've switched to this approach, which ended up being simpler. This exposed a couple latent unused variables in the NAN tests.

I put my changes on top of #1620 to avoid conflicts. This should be good to merge right after that one.

basvodde:
Hmm, I merged the other one but the diff still shows the changes that are already merged to master. Any idea how to not have that?

thetic:
> Hmm, I merged the other one but the diff still shows the changes that are already merged to master. Any idea how to not have that?

np, rebased

CMakeLists.txt
```diff
@@ -28,8 +28,15 @@ include(CTest)
...
+
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+  "int main(int argc, char ** argv) { throw 20; }"
+  CPPUTEST_HAVE_EXCEPTIONS
+)
+
 cmake_dependent_option(CPPUTEST_STD_CPP_LIB_DISABLED "Use the standard C++ library"
-  OFF "NOT CPPUTEST_STD_C_LIB_DISABLED" ON)
+  OFF "NOT CPPUTEST_STD_C_LIB_DISABLED;CPPUTEST_HAVE_EXCEPTIONS" ON)
```
basvodde, line 39: Is this one accidentally still here ?

thetic: I'm not sure I understand your question. This line automatically disables the C++ standard library if:
- the C standard library is disabled, or
- exceptions are not supported.

basvodde: Sorry, what I mean is that here is now the only place there still in CPPUTEST_HAVE_EXCEPTIONS. I wondered if that is still needed?

But... now I see I was wrong. Above this, the CPPUTEST_HAVE_EXCEPTIONS is still checked also. So, no more comment :)

PR #1612, opened 2022-07-10
Work around Apple Clang's quirks

thetic:
Depends on:

- #1601 by @offa.
- #1605 by me.

Resolves warnings from #1604.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -99,3 +99,15 @@ jobs:
         run: cmake --build build
       - name: Test
         run: ctest --test-dir build
+
+  xcode:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+      - name: Configure
+        run: cmake -B cpputest_build -S . -G Xcode -DWERROR=ON
+      - name: Build
+        run: cmake --build cpputest_build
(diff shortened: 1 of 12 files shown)


thetic:
I don't have access to apple clang outside of github actions. Best I can tell, some versions of apple clang support `override` as an extension to C++98.

mtfurlan:
I refuse to use a mac so I'm testing with https://github.com/sickcodes/Docker-OSX/
It's matched my coworkers actual hardware with everything so far, but I can make a coworker run this tomorrow.

mtfurlan:
Coworker gets the same `'override' keyword is a C++11 extension` warnings building your branch.

thetic:
are you providing your own flags are are these all defaults? It's rather odd to build with `-Weverything`, especially without setting a language standard.

mtfurlan:
Building your cpputest branch with `./autogen.sh && ./configure && make -j8`, that warning shows up as a warning.

When using MakefileWorker.mk with clang, `-Weverything` is set here: https://github.com/cpputest/cpputest/blob/master/build/MakefileWorker.mk#L209=
It was added in https://github.com/cpputest/cpputest/commit/7ce5f6428a3a05c590aa0531a742de208a198086

PR #1614, opened 2022-07-14
Remove `-Weverything` flag

thetic:
> Since `-Weverything` enables every diagnostic, we generally don’t
> recommend using it. `-Wall -Wextra` are a better choice for most projects.
> Using `-Weverything` means that updating your compiler is more difficult
> because you’re exposed to experimental diagnostics which might be of
> lower quality than the default ones.

https://clang.llvm.org/docs/UsersManual.html?highlight=weverything#diagnostics-enable-everything

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -206,8 +206,8 @@ ifeq ($(COMPILER_NAME),$(CLANG_STR))
 # -Wno-c++98-compat-pedantic -> Incompatibilities with C++98, these are happening through #define.
 # -Wno-reserved-id-macro -> Macro uses __ in MINGW... can't change that.
 # -Wno-keyword-macro -> new overload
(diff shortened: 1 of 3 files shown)


basvodde:
Did this actually cause problems?

I understand that everything is a lot and is not necessary recommended. So far, our attitude has been that we can try to fix even experimental warnings.... but still want all warning (and errors)

thetic:
It flags missing `override`s in C++98.

https://github.com/cpputest/cpputest/runs/7927116794?check_suite_focus=true#step:6:19

basvodde:
And there is no other way to fix that ?

thetic:
The behavior `-Weverything` is going to be different from version to version and between Apple clang and "normal" clang. As I see it, we can support `-Werror` and require that all enabled warnings are addressed, or `-Weverything` and depend on manual review of build logs to minimize violations, but we cannot practically use both unless someone wants to audit every supported version of both clang variants and commit to addressing every future version.

Personally, I think `-Werror` is much more useful than `-Weverything` and the documentation is pretty clear that use of `-Weverything` is not recommended. I think we'd be better off continuing to identify and enable useful warnings individually.

basvodde:
Hmm. Ok. You've convinced me.

PR #1617, opened 2022-08-16
Fix linking libraries with circular dependencies.

AlanRosenthal:
explaination of `--start-group` flag: https://stackoverflow.com/a/5651895/1639019
info about flag not existing with clang: https://stackoverflow.com/questions/28009290/linker-ld-on-os-x-how-to-use-wl-start-group-and-end-group#comment100304006_55856699

### DIFF
--- a/build/MakefileWorker.mk
+++ b/build/MakefileWorker.mk
@@ -502,11 +502,19 @@ flags:
 TEST_DEPS = $(TEST_OBJS) $(MOCKS_OBJS) $(PRODUCTION_CODE_START) $(TARGET_LIB) $(USER_LIBS) $(PRODUCTION_CODE_END) $(CPPUTEST_LIB) $(STDLIB_CODE_START)
 test-deps: $(TEST_DEPS)
 
+ifeq ($(COMPILER_NAME),$(CLANG_STR))
+LD_START_GROUP = -Wl,--start-lib
+LD_END_GROUP = -Wl,--end-lib
+else
+LD_START_GROUP = -Wl,--start-group
+LD_END_GROUP = -Wl,--end-group
+endif
+
 $(TEST_TARGET): $(TEST_DEPS)
 ifndef MORE_SILENCE
 	@echo Linking $@
 endif
-	$(SILENCE)$(CXX) -o $@ $^ $(LD_LIBRARIES) $(LDFLAGS)
+	$(SILENCE)$(CXX) -o $@ $(LD_START_GROUP) $^ $(LD_LIBRARIES) $(LD_END_GROUP) $(LDFLAGS)
 
 $(TARGET_LIB): $(OBJ)
 ifndef MORE_SILENCE


basvodde:
Could you say something about when this happened? Usually you'd avoid circular dependencies by not having them in the first place...

AlanRosenthal:
I'm testing out another change, will expand on `the why` once CI passes :)

AlanRosenthal:
For example. let's take two libs as an example, liba.a an libb.a .

liba.a uses a symbol (fn_b) in libb.a
libb.a uses a symbol (fn_a) in liba.a

aka, there's a circular dependency on the libraries.

During the linking process, the linker does a linear pass, and discards symbols it hasn't seen used it.

`ld liba.a libb.a` (list libs once)
linker scans liba.a, sees that no one is using fn_a and discards it. sees fn_b is unresolved.
linker scans libb.a, finds fn_b, marks it as found. sees fn_a is unresolved.
linker generates an error, cannot find fb_a

`ld liba.a libb.a liba.a libb.a` (list libs twice)
linker scans liba.a, sees that no one is using fn_a and discards it. marks fn_b is unresolved.
linker scans libb.a, finds fn_b, marks it as found. marks fn_a is unresolved.
linker scans liba.a (again), finds fn_a, and marks it as found.
linker scans libb.a (again), finds no new sybmols
linker has resolved all symbols, generates binary

`ld --start-group liba.a libb.a --end-group` (use gcc's start/end group)
linker scans liba.a.
linker scans libb.a.
linker resolves all symbols, across all items in group
linker discards any unused symbols, across all items in group

note: lld (clang's linker) uses `--start-lib` and `--end-lib` as flags instead of `--start-group` and `--end-group`

now you might ask what is a real world example of this?

I'm creating a library of a lot of files, and adding it as a USERLIB (libusercode.a).
libusercode.a has some unresolved symbols, which will be added as mocks: `int get_data(void)` and `void set_data(int)` . As included in this library are public functions: `void start_process(void)` and `void cancel_process(void)`.

CppUTest creates a library of the test file (test_mytest.cpp), mock files, (mock_data.cpp), as `lib/libMYTEST.a`. The test file also include calls to `start_process` and `cancel_process`, to exercise the tests.

we have 
`libusercode.a`:
symbols: start_process, cancel_process
unresolved symbols: get_data, set_data

`lib/libMYTEST.a`:
symbols: get_data, set_data
unresolved symbols: start_process, cancel_process

`libusercode.a` is dependent on `lib/libMYTEST.a`, and `lib/libMYTEST.a` is dependent on `libusercode.a`.

By passing in the flags `--start-group`, we tell the compiler to treat the libraries as one giant library. Which allows all symbols to resolved.

Let me know if you have any questions.

https://linux.die.net/man/1/ld (ctrl+f `--start-group`)
https://manpages.ubuntu.com/manpages/impish/man1/ld.lld.1.html (ctrl+f `--start-lib`)

basvodde:
Thanks.

I've been considering this... but I really do not want to merge this. (sorry). The reasons are: (1) circular dependencies are usually a sign of bad design and are best solved by changing your library structure (e.g. moving the start_process and cancel_process is separate libraries), and (2) this likely makes linking slower (MSVC traditionally did this automatically and that caused linking to slow down significantly).

When I merge this, then the linker will not give a warning anymore when I accidentally create a circular dependency. Thus, I prefer not merging it.

In your case, can you not solve this simply by adding the library that has circular dependencies twice to the list of libraries to link? AFAIK that would solve this also without any change and it will make the circular dependency visible in your Makefile.

PR #1620, opened 2022-08-20
Clean up CMake configuration variables

### LINKED ISSUE #1616: CMake options not "scoped" to CppUTest
I'm trying to integrate CppUTest into my project using [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake), but unfortunately the CMake [options specified in `CMakeLists.txt`](https://github.com/cpputest/cpputest/blob/v4.0/CMakeLists.txt#L60-L77) are not scoped to the CppUTest project (e.g. prefixed with `CPPUTEST_`).
I haven't checked if there are other "global" CMake variables/things that CppUTest uses that should be scoped, but I can imagine that they exist as well.

A problem that might arise is that another CMake-based project also define an option named `TESTS` to enable/disable its tests (or any of the other options CppUTest has defined).
That means that I will no longer be able to disable the tests for CppUTest but enable them for the other project.

Currently there is not yet a problem for me and as long as others do scope their options correctly there won't be a "collision". However, I think it would be a good idea to address this in a future release (even though this will be a breaking change).

For reference, also see the [Preparing projects for CPM.cmake](https://github.com/cpm-cmake/CPM.cmake/wiki/Preparing-projects-for-CPM.cmake) wiki page.

Although this might sound CPM.cmake specific, in reality this can also lead to problems if using `FetchContent` or Git submodules in a project that also uses CMake.

thetic:
- Prefix CMake configuration variables with `CPPUTEST_` to avoid collisions and clarify meaning in users' projects. (Closes #1616).
- Use `cmake_dependant_option` to hide options where they don't apply (e.g. disabling the C standard library implicitly disables the C++ standard library, disabling testing disables test discovery, etc).
- Remove backwards compatibility shims for old `HAS_INF`, `HAS_NAN`, `C++11` variables to consolidate breaking changes.
- Build examples when CppUTest is the root project to prevent regressions.

This depends on quite a few open PRs:

- #1614 
- #1584 
- #1588 
- #1579
- #1611


basvodde:
Failing Appveyer build.

basvodde:
Unfortunately, this one cannot be merged anymore

scripts/appveyor_ci_build.ps1
```diff
@@ -77,7 +77,7 @@ switch -Wildcard ($env:Platform)
         {
             $toolchain_filename = Get-ClangToolchainFilename
             $toolchain_path = (Join-Path (Split-Path $MyInvocation.MyCommand.Path) "..\cmake\$toolchain_filename")
-            $toolchain = "-DCMAKE_TOOLCHAIN_FILE=$toolchain_path"
+            $toolchain = "-DCMAKE_TOOLCHAIN_FILE=$toolchain_path -DCPPUTEST_WERROR=OFF"
```
thetic, line 80: MinGWClang throws some very confusing warnings around `_check_format_` that I could not figure out.

basvodde:
This one, that I had to merge first, now has merge conflicts.

PR #1625, opened 2022-09-02
Define and enforce formatting standard

### LINKED ISSUE #293: Create a development guideline
I think it would be a good idea to have a development guideline, and maybe some scripts that checks the code for this guidelines (like GStreamer).

I had this idea seeing that MockSupport.h has its indentation completely messed up, at least this (indentation :-) should have a clear guideline :-).

### LINKED ISSUE #1267: Clang-format support
How about supporting clang-format and use it to format the code base?

thetic:
I tried to align the config closely with existing code. I set the column limit at a generous 120 characters.

clang-format can be run directly against all checked in code with `clang-format -i $(git ls-files '*.cpp' '*.c' '*.h')`, with any number of editor integrations, or using a pre-commit hook.

I added a `.git-blame-ignore-revs` file to keep these changes out of blames. Enable it with `git config blame.ignoreRevsFile .git-blame-ignore-revs`.

Closes #293.
Closes #1267.


basvodde:
That is a lot of change.

include/CppUTest/CppUTestConfig.h
```diff
@@ -61,34 +61,33 @@
...
-  #define CPPUTEST_USE_STD_CPP_LIB 0
- #else
-  #define CPPUTEST_USE_STD_CPP_LIB 1
- #endif
+#ifdef CPPUTEST_STD_CPP_LIB_DISABLED
+#define CPPUTEST_USE_STD_CPP_LIB 0
+#else
+#define CPPUTEST_USE_STD_CPP_LIB 1
+#endif
 #endif
```
basvodde, line 78: I don't like these changes, this makes it less readable.

basvodde:
I know I'm potentially asking more work... but could we do this one in smaller chunks. E.g. first the Application, then the Ext etc. Then it is easier to see where the formatting got worst (like the #defines) and see if we can do something about that.

basvodde:
I will try to merge the smaller ones as soon as possible.

include/CppUTest/CppUTestConfig.h
thetic: Agreed. I'll set `IndentPPDirectives: AfterHash` to follow the existing style.

thetic:
> I know I'm potentially asking more work... but could we do this one in smaller chunks. E.g. first the Application, then the Ext etc. Then it is easier to see where the formatting got worst (like the #defines) and see if we can do something about that.

I've pared things down to juts the examples so we can hash out a config before we apply it more broadly.

examples/AllTests/AllTests.cpp
```diff
@@ -31,22 +31,18 @@
...
 public:
-    virtual bool isEqual(const void* object1, const void* object2)
-    {
+    virtual bool isEqual(const void* object1, const void* object2) {
         return object1 == object2;
     }
 
-    virtual SimpleString valueToString(const void* object)
-    {
+    virtual SimpleString valueToString(const void* object) {
```
basvodde, line 40: I think, right now, almost everywhere in CppUTest it has the opening bracket of functions (and TEST macros) on a new line. Personally I do not see a benefit of changing this? Can we configure it so to use that style ? Anything against that?

examples/AllTests/FEDemoTest.cpp
```diff
@@ -30,19 +30,20 @@
 #include "CppUTest/TestRegistry.h"
 
 #if CPPUTEST_HAVE_FENV
-#include "CppUTestExt/IEEE754ExceptionsPlugin.h"
+    #include "CppUTestExt/IEEE754ExceptionsPlugin.h"
```
basvodde, line 33: These are minor issues. But this one didn't needed to be indented, but we can't avoid that, I assume ?

basvodde:
Ok, two minor comments left, then I'm ready to merge this one.

examples/AllTests/FEDemoTest.cpp
basvodde: For now, I assume this one cannot be different as it is hard to figure out when to indent the #include and when not to.

PR #1631, opened 2022-09-07
Format extensions source

thetic:
Another round of review on the formatting config and output. Applied to CppUTestExt this time.

Follow up to #1625.


include/CppUTestExt/GTestConvertor.h
```diff
@@ -247,45 +244,49 @@ class GTestResultReporter : public ::testing::ScopedFakeTestPartResultReporter
...
     }
 
     void testBody()
     {
         try {
             test_->TestBody();
         }
-        catch (CppUTestFailedException& ex)
-        {
+        catch (CppUTestFailedException& ex) {
```
thetic, line 271: TODO: break between `catch` and brace

src/CppUTestExt/MemoryReporterPlugin.cpp
```diff
@@ -57,9 +54,8 @@ bool MemoryReporterPlugin::parseArguments(int /* ac */, const char *const *av, i
 MemoryReportFormatter* MemoryReporterPlugin::createMemoryFormatter(const SimpleString& type)
 {
     if (type == "normal") {
-        return  new NormalMemoryReportFormatter;
-    }
-    else if (type == "code") {
+        return new NormalMemoryReportFormatter;
+    } else if (type == "code") {
```
thetic, line 58: TODO: break between brace and `else`

src/CppUTestExt/MockActualCall.cpp
```diff
@@ -371,22 +370,25 @@ bool MockCheckedActualCall::hasFailed() const
...
         potentiallyMatchingExpectations_.resetActualCallMatchingState();
         return;
     }
 
     if (potentiallyMatchingExpectations_.hasFinalizedMatchingExpectations())
-        FAIL("Actual call is in progress, but there are finalized matching expectations when checking expectations. This cannot happen."); // LCOV_EXCL_LINE
+        FAIL(
+            "Actual call is in progress, but there are finalized matching expectations when checking expectations. "
+            "This cannot happen."
+        ); // LCOV_EXCL_LINE
```
thetic, line 391: TODO: Tell clang-format that `LCOV_*` comments are special.

include/CppUTestExt/GTestConvertor.h
thetic: This does not appear possible. Clang-format lumps `catch` in with the other controls statements (`if`, `for`, etc.).

include/CppUTestExt/CodeMemoryReportFormatter.h
```diff
@@ -47,21 +47,26 @@ class CodeMemoryReportFormatter : public MemoryReportFormatter
     virtual void report_test_start(TestResult* result, UtestShell& test) _override;
     virtual void report_test_end(TestResult* result, UtestShell& test) _override;
 
-    virtual void report_alloc_memory(TestResult* result, TestMemoryAllocator* allocator, size_t size, char* memory, const char* file, size_t line) _override;
-    virtual void report_free_memory(TestResult* result, TestMemoryAllocator* allocator, char* memory, const char* file, size_t line) _override;
+    virtual void report_alloc_memory(
```
basvodde, line 50: Can these stay on one line?

thetic:
> Can these stay on one line?

The column limit is currently set to 120 characters. We can increase or remove the limit. What would you prefer?

basvodde:
My first thought is... remove.

But if we remove it, then it will change manual line break to one-lines? So, perhaps increase it to a significant amount.

thetic:
I thought 120 was a significant amount, lol. I'll see how it handles manual breaks when the limit is removed.

thetic:
> A column limit of `0` means that there is no column limit. In this case, clang-format will respect the input’s line breaking decisions within statements unless they contradict other rules.

basvodde:
Thinking it over a bit. I do like that the limit actually puts some lines on one line that were previously too short.

Testing the columns a bit on my (way too large) monitor, I suggest setting the limit to 180.

thetic:
> I suggest setting the limit to 180.

done

examples/AllTests/EventDispatcherTest.cpp
```diff
@@ -106,16 +98,8 @@ TEST(EventDispatcher, DifferentEventWithRegistrationDoesNotResultIntoCallback)
...
-        .expectOneCall("notify")
-        .onObject(&observer)
-        .withParameterOfType("Event", "event", &event)
-        .withParameter("timeOutInSeconds", 10);
-    mock()
-        .expectOneCall("notify")
-        .onObject(&observer2)
-        .withParameterOfType("Event", "event", &event)
-        .withParameter("timeOutInSeconds", 10);
+    mock().expectOneCall("notify").onObject(&observer).withParameterOfType("Event", "event", &event).withParameter("timeOutInSeconds", 10);
```
basvodde, line 101: Is there no way to keep these formatted as they were?

thetic: This expression is less that 180 columns, so the formatter puts it on a single line.

basvodde: This isn't so much about the length of the line but about the formatting of method chaining. I checked online also though and it seems to be a missing feature in the clang formatter at the moment. Which sucks. I'll need to think about this a bit... but my initial thought is probably that this is big enough to anyways turn the max line off completely and leave it to the choice of the author.

include/CppUTestExt/MockActualCall.h
```diff
@@ -42,79 +42,118 @@ class MockActualCall
...
-    virtual const void * returnConstPointerValue()=0;
-    virtual const void * returnConstPointerValueOrDefault(const void * default_value)=0;
-
-    virtual void (*returnFunctionPointerValue())()=0;
-    virtual void (*returnFunctionPointerValueOrDefault(void (*default_value)()))()=0;
-
-    virtual MockActualCall& onObject(const void* objectPtr)=0;
+    virtual MockActualCall& withName(const SimpleString& name) = 0;
+    virtual MockActualCall& withCallOrder(unsigned int callOrder) = 0;
+    MockActualCall& withParameter(const SimpleString& name, bool value)
```
basvodde, line 47: Ouch. Could we set AllowShortFunctionsOnASingleLine  to SFS_Inline so these don't get expanded?

thetic:
I'm not willing to spend any more time on this. I thought autoformatting would help normalize style and improve the readability of the all the long, dense lines, but I don't think that's how this is panning out.

basvodde:
Ok, then I'll pick it up from here as I do think it is worth it.

thetic:
I think what you're looking for is

```yaml
AllowShortFunctionsOnASingleLine: Inline
ColumnLimit: 0
```

thetic:
Any progress on finding a config you're happy with? If we're not going to pursue formatting, we should probably remove the config and CI step.

PR #1632, opened 2022-09-08
Switch from putchar to fputs

thetic:
When semihosting, repeatedly calling `fputs` for each character drives very high syscall overhead.

Since this was the only use of `PlatformSpecificPutchar`, I've removed it entirely.


basvodde:
Ok. Could we remove PlatformSpecificStdOut from the interface completely? A platform specific global variable makes me a bit uncomfortable. So then PlatformSpecificPuts would just have 1 parameter, or if really needed, 2 parameters with the second one defaulting to NULL (meaning stdout in implementation)

thetic:
This is a C interface, so we can't use default parameters. Since it's a function pointer, I don't think that would work in C++ either.

thetic:
I think it's worth noting that PlatformSpecificFunctions_c.h is full of mutable global variables already.

thetic:
To clarify, we do need the file argument: https://github.com/cpputest/cpputest/blob/master/src/CppUTest/JUnitTestOutput.cpp#L318

basvodde:
Hrm, ok

PR #1633, opened 2022-09-08
Omit tests from coverage

thetic:
This is in line with behavior from before CI broke: https://coveralls.io/builds/44525666


basvodde:
Ah, finally 100% back again.

PR #1642, opened 2022-09-15
Add vcpkg installation instructions

MonicaLiu0311:
Cpputest is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for cpputest and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build cpputest, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/cpputest/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,13 +64,9 @@ TEST(FirstTestGroup, FirstTest)
 
 You can build and install cpputest using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
 
-* git clone https://github.com/Microsoft/vcpkg.git
-* cd vcpkg
-* ./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
-* ./vcpkg integrate install
-* ./vcpkg install cpputest
-
-The cpputest port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
+```console
+$ vcpkg install cpputest (More information: https://github.com/microsoft/vcpkg)
+```
 
 
 ## Command line switches
--- a/README.md
+++ b/README.md
@@ -62,6 +62,17 @@ TEST(FirstTestGroup, FirstTest)
 }
 ```
 
+You can build and install cpputest using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+* git clone https://github.com/Microsoft/vcpkg.git
+* cd vcpkg
+* ./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
+* ./vcpkg integrate install
+* ./vcpkg install cpputest
+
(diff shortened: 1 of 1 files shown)


thetic:
I'd prefer if we could just [add a badge](https://github.com/microsoft/vcpkg/discussions/17783) instead of duplicating your own docs.

thetic:
I don't think [this patch](https://github.com/microsoft/vcpkg/blob/master/ports/cpputest/fix-arm-build-error.patch) is necessary since #1629.

MonicaLiu0311:
> I don't think [this patch](https://github.com/microsoft/vcpkg/blob/master/ports/cpputest/fix-arm-build-error.patch) is necessary since #1629.

Okay, thanks for the notification. You can submit an issue or PR in [vcpkg](https://github.com/Microsoft/vcpkg/). Alternatively, we will remove this patch in the next stable release.

MonicaLiu0311:
> I'd prefer if we could just [add a badge](https://github.com/microsoft/vcpkg/discussions/17783) instead of duplicating your own docs.

I will close this PR, thanks for your suggestion, we will consider it carefully.

MonicaLiu0311:
Users prefer [adding a badge](https://github.com/microsoft/vcpkg/discussions/17783) instead of docs. 

Badge: Like a CI badge or coverage badge, to allow library owners on GitHub (or other communities) to notify their users that this library is available in vcpkg.

basvodde:
Wait wait

basvodde:
I don't think the decision is made yet. We also have other package managers in the help, so @thetic why not add this one ?

MonicaLiu0311:
> I don't think the decision is made yet. We also have other package managers in the help, so @thetic why not add this one ?

Anyway, thank you for reopening this PR, we don't have a representative Badge yet, would you like to add instructions for vcpkg?

basvodde:
@thetic Do you have any reason why this shouldn't be merged (independent of the badge) ?

thetic:
GItHub won't show my inline comments for some reason. I suggested, in the README.md additions, to replace

```
* git clone https://github.com/Microsoft/vcpkg.git
* cd vcpkg
* ./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
* ./vcpkg integrate install
* ./vcpkg install cpputest
```
with
````suggestion
```console
$ vcpkg install cpputest
```
````
to avoid having to maintain instructions on installing vcpkg itself.

**Comment by MonicaLiu0311** [comment]
@thetic 
It can be modified as you said, but would you like to add a link to vcpkg? As follows:

```console
$ vcpkg install cpputest (More information: https://github.com/microsoft/vcpkg)
```

basvodde:
Ok, can you change it in the PR?

MonicaLiu0311:
> Ok, can you change it in the PR?

Sorry, it was my oversight, thank you for your reminder.

PR #1644, opened 2022-09-18
Fix QEMU compabiltity

thetic:
This lets tests for ARM/GCC run with `-DCMAKE_CROSSCOMPILING_EMULATOR=qemu-arm-static;-cpu;cortex-m4`.

I can run this locally with qemu-arm v7.1, but the version of the `qemu-user-static` package available in CI (6.2 from Ubuntu) does not pass command line arguments to the executable.


basvodde:
Thanks again

PR #1645, opened 2022-09-19
Fix compilation on C++20

offa:
Remove deprecated usage of `volatile` in C++20.

### DIFF
--- a/examples/AllTests/FEDemoTest.cpp
+++ b/examples/AllTests/FEDemoTest.cpp
@@ -44,8 +44,6 @@ extern "C" {
 
     #include <limits>
 
-static volatile float f;
-
 TEST_GROUP(FE_Demo)
 {
     void setup()
@@ -56,13 +54,13 @@ TEST_GROUP(FE_Demo)
 
 IGNORE_TEST(FE_Demo, should_fail_when__FE_DIVBYZERO__is_set)
 {
-    f = 1.0f;
+    float f = 1.0f;
     CHECK((f /= 0.0f) >= std::numeric_limits<float>::infinity());
 }
 
 IGNORE_TEST(FE_Demo, should_fail_when__FE_UNDERFLOW__is_set)
 {
-    f = 0.01f;
+    float f = 0.01f;
     while (f > 0.0f)
         f *= f;
     CHECK(f == 0.0f);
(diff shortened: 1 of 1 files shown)


thetic:
can/should we add C++20 to the basic CI builds?

offa:
Sounds good to me :+1:. We should use a recent version of GCC / Clang though.

offa:
Indeed, didn't notice there's GCC 10 too.

basvodde:
Hmm, I remember having a discussion about the volatile here. But it seems they are now IGNORE_ anyways, so I guess this change will not actually do much harm.

thetic:
`volatile` itself isn't deprecated; using assignment operators (e.g `*=`, `/=`) on `volatile`s is.

offa:
Regardless of `volatile`, why even use a global here?

thetic:
> why even use a global here?

I agree, but I think the use of `volatile` may still be necessary to prevent the loops from being optimized out.

basvodde:
Yes, I checked the history of this code. The volatile was added to prevent the -O2 optimization from optimizing the code away.  I suggest we'll put that back anyways, unless it causes problems.

thetic:
So, rather than removing volatile, I suggest replacing `f *= f;` with `f = f * f;`.

examples/AllTests/FEDemoTest.cpp
```diff
@@ -56,21 +54,21 @@ TEST_GROUP(FE_Demo)
...
+    float f = 1.0f;
     CHECK((f /= 0.0f) >= std::numeric_limits<float>::infinity());
 }
 
 IGNORE_TEST(FE_Demo, should_fail_when__FE_UNDERFLOW__is_set)
 {
-    f = 0.01f;
+    float f = 0.01f;
     while (f > 0.0f)
         f *= f;
```
thetic, line 65:
```suggestion
    volatile float f = 0.01f;
    while (f > 0.0f)
        f = f * f;
```

examples/AllTests/FEDemoTest.cpp
```diff
@@ -56,21 +54,21 @@ TEST_GROUP(FE_Demo)
...
     CHECK(f == 0.0f);
 }
 
 IGNORE_TEST(FE_Demo, should_fail_when__FE_OVERFLOW__is_set)
 {
-    f = 1000.0f;
+    float f = 1000.0f;
     while (f < std::numeric_limits<float>::infinity())
         f *= f;
```
thetic, line 73:
```suggestion
    volatile float f = 1000.0f;
    while (f < std::numeric_limits<float>::infinity())
        f = f * f;
```

basvodde:
And should we keep the variable global, assuming there was a reason for that?

thetic (approved):

basvodde:
Ok, we'll see whether there was a good reason for it to be global.

PR #1648, opened 2022-09-20
Fix use with `long long` disabled

thetic:
~I broke this.~ I think this has been broken for a while. I've added it to CI to prevent future regressions.

### DIFF
--- a/.github/workflows/extended.yml
+++ b/.github/workflows/extended.yml
@@ -155,3 +155,19 @@ jobs:
         run: cmake --build cpputest_build
       - name: Test
         run: ctest --test-dir cpputest_build
+
+  no-long-long:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+      - name: Configure
+        run: >
+          cmake
+          -B cpputest_build
+          -D CPPUTEST_USE_LONG_LONG=NO
+          -D CMAKE_CXX_COMPILER=clang++
+      - name: Build
+        run: cmake --build cpputest_build -j
+      - name: Test
+        run: ctest --test-dir cpputest_build -j
--- a/.github/workflows/extended.yml
+++ b/.github/workflows/extended.yml
@@ -156,7 +156,19 @@ jobs:
       - name: Test
         run: ctest --test-dir cpputest_build
 
-  no-long-long:
+  automake-no-long-long:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+      - name: Configure
+        run: |
+          autoreconf -i .
+          ./configure --disable-longlong
+      - name: Build and test
+        run: make tdd
+
+  cmake-no-long-long:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -274,10 +274,11 @@
  * LLONG_MAX is set in limits.h. This is a crude attempt to detect long long support when no configure is used
  *
  */
-
-#if !defined(CPPUTEST_LONG_LONG_DISABLED) && !defined(CPPUTEST_USE_LONG_LONG)
-#if defined(CPPUTEST_HAVE_LONG_LONG_INT) || defined(LLONG_MAX)
+#ifndef CPPUTEST_USE_LONG_LONG
+#if !defined(CPPUTEST_LONG_LONG_DISABLED) && (defined(CPPUTEST_HAVE_LONG_LONG_INT) || defined(LLONG_MAX))
 #define CPPUTEST_USE_LONG_LONG 1
+#else
+#define CPPUTEST_USE_LONG_LONG 0
 #endif
 #endif
 
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -296,27 +296,22 @@ typedef unsigned long long cpputest_ulonglong;
 #define CPPUTEST_SIZE_OF_FAKE_LONG_LONG_TYPE 8
 #endif
 
-struct cpputest_longlong
-{
 #if defined(__cplusplus)
-  cpputest_longlong() {}
-  cpputest_longlong(int) {}
+extern "C" {
 #endif
-  char dummy[CPPUTEST_SIZE_OF_FAKE_LONG_LONG_TYPE];
-};
 
-struct cpputest_ulonglong
+typedef struct
 {
-#if defined(__cplusplus)
-  cpputest_ulonglong() {}
-  cpputest_ulonglong(int) {}
-#endif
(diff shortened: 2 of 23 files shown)


include/CppUTest/CppUTestConfig.h
```diff
@@ -281,7 +281,7 @@
 #endif
 #endif
 
-#if CPPUTEST_USE_LONG_LONG
+#ifdef CPPUTEST_USE_LONG_LONG
```
basvodde, line 284: Could we keep this as an #if. In general we prefer #if over #ifdef. In this case, it would likely require a change at the previous #if block to include an #else ?

thetic: All other checks are `#ifdef`s right now which is the cause of this regression. I can change them all if that's what you prefer.

basvodde: All other checks on the CPPUTEST_USE (which can be used outside the Config file) are always done with  #if. That one is set based on different conditions. So, in this case, the block before this one (which checks whether CPPUTEST_USE_LONG_LONG should be set to 1) probably ought to change so that it will set it to 0 is the case no long long is used. That way it will be consistent with the rest of the file.

thetic: I made the change. I ran into trouble with GCC's '-Wuninitialized` and with DOS trying to get `cpputest_longlong` and `cpputest_ulonglong` handled. I can't figure out how they're useful, so I removed them.

basvodde: The reason is that we cannot have #if in the header files to change the v table structure. It leads to very very unexpected errors.

thetic: I'm not sure I follow. That would only happen if different values of `CPPUTEST_USE_LONG_LONG` were defined for different translation units, which would be a problem no matter what.

Anyway, I've rolled back my removal of `cpputest_longlong` and `cpputest_ulonglong`. They break the build.

thetic: I think I've found a compromise. See my latest commit. There's no way to set up `cpputest_longlong` and `cpuutest_ulonglong` so that assignment works _and_ C linkage works. I managed to retain ABI stability. The only interface change required is to disable the `LONGLONGS_EQUAL`, `UNSIGNED_LONGLONGS_EQUAL`, etc. macros when `CPPUTEST_USE_LONG_LONG` is unset.

basvodde: Thanks, I'll have a look at it. It has happened in the past that the CPPUTEST_USE_LONG_LONG was set for different translation units (set differently when compiling CppUTest itself than for the tests). The big problem with it is that it doesn't give an error but behaves very surprising :)

As a result, we avoid #if in headers and especially for virtual functions.

.github/workflows/extended.yml
```diff
@@ -155,3 +155,31 @@ jobs:
         run: cmake --build cpputest_build
       - name: Test
         run: ctest --test-dir cpputest_build
+
+  automake-no-long-long:
```
thetic, line 159: Both of these configs failed without my changes.

include/CppUTest/UtestMacros.h
```diff
@@ -226,6 +226,7 @@
 #define UNSIGNED_LONGS_EQUAL_LOCATION(expected, actual, text, file, line)\
   do { UtestShell::getCurrent()->assertUnsignedLongsEqual((unsigned long)expected, (unsigned long)actual, text, file, line); } while(0)
 
+#if CPPUTEST_USE_LONG_LONG
```
basvodde, line 229: I guess this #if is not needed ?

tests/CppUTest/TestUTestMacro.cpp
```diff
@@ -226,30 +226,6 @@ IGNORE_TEST(UnitTestMacros, UNSIGNED_LONGLONGS_EQUAL_TEXTWorksInAnIgnoredTest)
...
-    fixture.runTestWithMethod(LONGLONGS_EQUALFailsWithUnsupportedFeatureTestMethod_);
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("\"CPPUTEST_USE_LONG_LONG\" is not supported");
-}
-
-TEST(UnitTestMacros, UNSIGNED_LONGLONGS_EQUALFailsWithUnsupportedFeature)
-{
-    fixture.runTestWithMethod(UNSIGNED_LONGLONGS_EQUALFailsWithUnsupportedFeatureTestMethod_);
-    CHECK_TEST_FAILS_PROPER_WITH_TEXT("\"CPPUTEST_USE_LONG_LONG\" is not supported");
-}
-
```
basvodde, line 252: These can also stay when the earlier #define it removed.

I do have a preference to that as it keeps the interface the same.

basvodde:
Ok, no major changed needed. Will wait for your comment on the small points above then merge this. If you feel it is not needed, then that is ok too and I'll anyways merge it then :)

include/CppUTestExt/MockSupport_c.h
```diff
@@ -61,7 +61,7 @@ typedef struct SMockValue_c
         unsigned int unsignedIntValue;
         long int longIntValue;
         unsigned long int unsignedLongIntValue;
-#ifdef CPPUTEST_USE_LONG_LONG
+#if CPPUTEST_USE_LONG_LONG
```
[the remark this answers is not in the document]
thetic: This is a union.

tests/CppUTest/TestUTestMacro.cpp
thetic: They cannot.That's the whole problem this MR is trying to solve. These are the constraints we're trying to place on the `cpputest_[u]longlong` types:
 1. same size as native `long long` types to avoid ABI breakages.
 2. C linkage so that C++ and C code can interoperate.
 3. assignable from an integral constant so `[UNSIGNED_]LONGLONGS_EQUAL` can be called without native `long long` support.

Based on everything I've tried, we have to pick two; we can't accomplish all three. I chose to drop support for 3, since these calls fail at runtime anyway.

include/CppUTestExt/MockSupport_c.h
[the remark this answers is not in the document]
basvodde: Don't think line 68 has an impact on it... but the union-ness does :) Sorry, I was wrong. If it is in a union then it isn't a problem. Thanks!

tests/CppUTest/TestUTestMacro.cpp
basvodde: Yes, it could be solved by declaring a cpputest_longlong in the test since it is anyways inside a #if and it doesn't need to be given a value for the test. It would then still properly test that it would give a failure. And I guess it might be worth it...

That said, in practice, because of (3), the same code will never work that works with longlong support will not compile for not longlong support. So, it is probably not that useful to keep it.

I'll merge it as is. If you feel it still useful to add these tests to test the proper failure, then they can be returned in a separate PR.

PR #1654, opened 2022-10-29
Move development settings out of CMake project

thetic:
There is a lot of stuff in the CMake project that doesn't affect users and makes for a confusing interface. This change set moves those settings into a [presets files](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) that contains all the settings used for development without trying to propagate them to users. The CMakeLists.txt files describe what the project is while the presets define different ways the project is used. Use of the presets file requires v3.21 or greater, but the project can used without them maintaining compatibility with v3.8.


thetic:
How are the GccNoStdC and Iar platforms tested?

thetic:
I think GccNoStdC was broken when the string cache was added.
I get the following backtrace when trying to run:
```
#0  0x0000000000000000 in ?? ()
#1  0x0000555555704bad in checkedMalloc (size=<optimized out>) at src/CppUTest/TestMemoryAllocator.cpp:35
#2  TestMemoryAllocator::alloc_memory (this=<optimized out>, size=<optimized out>) at src/CppUTest/TestMemoryAllocator.cpp:164
#3  0x0000555555703c5e in SimpleStringInternalCache::createInternalCacheNodes (this=this@entry=0x7fffffffbe08) at src/CppUTest/SimpleStringInternalCache.cpp:63
#4  0x00005555557042d6 in SimpleStringInternalCache::SimpleStringInternalCache (this=0x7fffffffbe08) at src/CppUTest/SimpleStringInternalCache.cpp:47
#5  GlobalSimpleStringCache::GlobalSimpleStringCache (this=this@entry=0x7fffffffbe00) at src/CppUTest/SimpleStringInternalCache.cpp:249
#6  0x000055555566b302 in main (ac=5, av=0x5) at tests/CppUTest/AllTests.cpp:37
```

thetic:
See #1656

basvodde:
Is this ready to be merged ?

thetic:
I think so unless you or @offa had any feedback.

.github/workflows/extended.yml
```diff
@@ -79,51 +72,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    name: Disable libc
+    steps:
+      - name: Checkout
+        uses: actions/checkout@main
+      - name: Configure
+        run: cmake --preset no-std-c
+      # FIXME: This is broken.
+      # - name: Build
+      #   run: cmake --build cpputest_build --verbose
```
basvodde, line 111: Shouldn't the build be enabled here ?

basvodde:
Ok, just one minor feedback, the rest looks fine @offa  you want to check this still ?

basvodde:
I assume @offa has no more comment and merging this now.

offa:
Sorry, I was unavailable the last days. But your assumptions are right – good PR! :+1:

PR #1655, opened 2022-10-29
Drop confusing *LibName variables

thetic:
These make the code very confusing and I cannot understand the use case. Multiple architectures cannot exist in the same CMake project. When installing, multiple architectures should probably be installed to different locations, not have different names. The native [`CMAKE_<CONFIG>_POSTFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIG_POSTFIX.html) variable can be used to append arbitrary suffixes to installed libraries.

This removes changes from #1538. I may not have all the context here. So I'd appreciate any insight @jgonzalezdr has on what this was originally intended to support.


offa (approved):

basvodde:
@jgonzalezdr Any comments on this ?

basvodde:
As no feedback, I'll merge this for now.

PR #1657, opened 2022-10-30
Simplify C++98 compat

thetic:


basvodde:
Interesting.

I'm a bit worried about it as I remember some pre-processors giving a warning when #defining a language keyword. I assume these warnings can be turned off when they do happen, was that what you were thinking? I guess I'm ok with that.

The advantage is that it is easier to read, the IDE detects it, and people are less likely to forget the override?
The trade-off is the above warning?

(it also seems to be in conflict now after merging the other commits, it probably needs to be rebased)

thetic:
These macros are only ever defined for language versions in which they _aren't_ keywords so I don't expect that to arise.

basvodde:
I was thinking this over. This PR will lead to a potential other problem in that in old C++ code, using override and nullptr as variable names will suddenly lead to very weird errors (I've had to debug errors in the past where variables were changed by a 3rd party library #define). I assume it won't happen a lot but it is an additional drawback of this PR.

thetic:
This was just an idea. We don't have to pursue it. There are a couple tweaks in dd7f7c6ab290520f5775f368547cc6b2b7493b19 that I'd like to cherry-pick if we don't want like this.

thetic:
Closed in favor of #1665

PR #1660, opened 2022-10-31
Proposal: Drop GTest/GMock support

thetic:
The compatible GTest/GMock versions are effectively dead.


basvodde:
Why are they dead? We don't support the latest versions... but I've build and used the versions that are there now. I've always found it important because some companies ended up with both the test frameworks (historically) and this allowed them to make one executable for both of them. So, I'd like to keep the support ;)

Thanks for considering and thinking about it though! I do appreciate that.

thetic:
OK, I'll try to get a working build in CI and then look into compatibility with more recent versions. It looks like 1.6 is the most recent version expected to work, can you confirm?

basvodde:
According to the Makefile.am it ought to work on 1.5, 1.6, 1.7, 1.8.

They ought to be build with the check_all autotools build command. That one ought to be also done in the github build I guess. That is the most thorough check that we had so far.

PR #1661, opened 2022-10-31
Improve performance of CHECK_COMPARE_LOCATION.

dmitrykos:
Improve performance of CHECK_COMPARE_LOCATION by moving comparison result check above the SimpleString instantiation.

Performance drop of the test suite will occur if CHECK_COMPARE_LOCATION() is called inside some tight loop which is checking many multiple variants of values. SimpleString instantiation and filling with error message takes time and in case of a large loop may block execution for a long time. Proposed improvement eliminates this problem while keeping logic of the check intact.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -153,11 +153,15 @@
   CHECK_COMPARE_LOCATION(first, relop, second, text, __FILE__, __LINE__)
 
 #define CHECK_COMPARE_LOCATION(first, relop, second, text, file, line)\
- do { SimpleString conditionString;\
-      conditionString += StringFrom(first); conditionString += " ";\
-      conditionString += #relop; conditionString += " ";\
-      conditionString += StringFrom(second);\
-      UtestShell::getCurrent()->assertCompare((first) relop (second), "CHECK_COMPARE", conditionString.asCharString(), text, __FILE__, __LINE__);\
+ do {\
+      bool success = (first) relop (second);\
+      if (!success) {\
+          SimpleString conditionString;\
+          conditionString += StringFrom(first); conditionString += " ";\
+          conditionString += #relop; conditionString += " ";\
+          conditionString += StringFrom(second);\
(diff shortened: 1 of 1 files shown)


dmitrykos:
Hi @basvodde !

Thank you for accepting my PR, I do not mind to elaborate on the usage scenario. Consider the following test case:
```cpp
TEST(MATH_Common, ipow)
{
    enum { ERROR_THRESHOLD = 384 };
    char str[256] = { 0 };

    for (int32_t i = -4; i <= 4; ++i)
    {
        for (int32_t j = 0; j < 31; ++j)
        {
            uint64_t _v = (int64_t)pow((double)i, (double)j);
            uint64_t v = my_ipow(i, j);
            uint64_t e = (_v > v ? _v - v : v - _v);

            //if (e > ERROR_THRESHOLD)
            //{
                sprintf(str, "i = %d, j = %d", i, j);
                CHECK_COMPARE_TEXT(e, <=, ERROR_THRESHOLD, str);
            //}
        }
    }
}
```

Here function under a test is `my_ipow()` with some custom implementation and its result is comparted against standard `pow()`. In this scenario `CHECK_COMPARE_TEXT` will be called 279 times. The defensive programming approach would be to add `if (e > ERROR_THRESHOLD)` check to surround `sprintf` and `CHECK_COMPARE_TEXT` but with the proposed PR `SimpleString` is not allocated and filled on every pass that improves performance of the test suite, especially if it has similar tests like that.

basvodde:
Yes, usually you do not want to have loops or if inside a test case.

Every test case ought to be ONE scenario. Therefore it should be known upfront what the one scenario is... and hence no need for for and if statements

In your case, you'd probably write one TEST for threshold -1. One seperate one for threshold and perhaps one for threshold +1. You know the ought to know the values of i and j when making the test case.

dmitrykos:
Yes, I agree. Normally I do write corner test cases and there is one test per case but in this example, due to importance of the correct output in the whole range there is no way out. Sometimes corner cases can pass tests while unexpectedly failing in the middle (could be some saturation issue, flawed conditional logic, ...). It all depends on the logic of the tested function basically. In this case safer would be to test whole range to be sure it never fails on any supported platform.

PR #1662, opened 2022-10-31
Allow redirection of PlatformSpecificStdOut.

dmitrykos:
When putchar support was dropped by 80d807c1de79ee7307beae923ba2ad382bd17cd5 commit it became impossible to redirect log output to some arbitrary logging function without removing the `const` specifier of `PlatformSpecificStdOut`.

Consider the following usage example (pseudocode):
```cpp
static void LogRedirectorPuts(const char* str, PlatformSpecificFile file)
{
#ifdef _UTEST_IMMEDIATE_LOG
    Log.Warning("%s", str);
#else
    ((std::string *)file)->append(str);
#endif
}

static void RunTests()
{
    std::string log_buffer;

    PlatformSpecificStdOut = &log_buffer;
    PlatformSpecificFPuts  = LogRedirectorPuts;

    char **tmp = NULL;
    int result = RUN_ALL_TESTS(0, tmp);

    if (result != 0)
    {	
        Log.ErrorSrc(TEST_LOG, "failure:\n--//--\n%s--//--", log_buffer.c_str());
    }
    else
    {
        Log.VerboseSrc(TEST_LOG, "success:\n--//--\n%s--//--", log_buffer.c_str());
    }

    PlatformSpecificStdOut = NULL;
}
```

This PR makes log redirection possible again and drops `const` specifier of `PlatformSpecificStdOut` for all platform implementations.

### DIFF
--- a/include/CppUTest/PlatformSpecificFunctions_c.h
+++ b/include/CppUTest/PlatformSpecificFunctions_c.h
@@ -62,7 +62,7 @@ extern int (*PlatformSpecificAtExit)(void(*func)(void));
 /* IO operations */
 typedef void* PlatformSpecificFile;
 
-extern const PlatformSpecificFile PlatformSpecificStdOut;
+extern PlatformSpecificFile PlatformSpecificStdOut;
 
 extern PlatformSpecificFile (*PlatformSpecificFOpen)(const char* filename, const char* flag);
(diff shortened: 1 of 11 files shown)


basvodde:
Ok. Seems good. Thanks

PR #1663, opened 2022-11-02
Add GTest integration to CI

thetic:
I checked [v1.10](https://github.com/google/googletest/releases/tag/release-1.10.0) compatibility and didn't get very far.


basvodde:
Ok, excellent. Seems to work fine still or am I interprettng this wrong ?

thetic:
Yup. I had to install Python 2 and I haven't got this working in CMake or with more recent versions yet, but I think it's worth getting under test to avoid regressions.

PR #1665, opened 2022-11-03
Keep C++98 compatibility macros out of C

thetic:
- Keep `NULLPTR`, `_override`, and `_destructor_override` macros out of
  C. We had some leak into TestHarness_c.h.
- Drop Green Hills workaround for `override` with function pointers added
  in #1239. This was mitigated in #1605.

### DIFF
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -316,29 +316,24 @@ typedef struct
 
 #endif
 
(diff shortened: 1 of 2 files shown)


basvodde:
Ok, though I liked the other one... I do think this one is safer than the other one. I was surprised you already closed the other one, but am ok with this. Thanks again. Your contributions are much appreciated!

PR #1666, opened 2022-11-04
Test with recent GTest versions

thetic:
I am very skeptical of these results. I don't claim to understand autotools. Could you verify that this is exercising everything it should? I noticed some related code was commented out in:
 - f5c2367ac28a8fa9f9148bbadc79d53284e80fd9
 - b6899f75cffdae8c44d4689890bf6152c4a2b73a

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -77,6 +77,9 @@ jobs:
           - name: GTest 1.8
             os: ubuntu-20.04
             target: check_gtest18
+          - name: GTest 1.10
+            os: ubuntu-20.04
+            target: check_gtest110
           - name: Disable long long
             os: ubuntu-latest
             configure_args: --disable-longlong
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -80,6 +80,9 @@ jobs:
           - name: GTest 1.10
             os: ubuntu-20.04
             target: check_gtest110
+          - name: GTest 1.11
+            os: ubuntu-20.04
+            target: check_gtest111
           - name: Disable long long
             os: ubuntu-latest
             configure_args: --disable-longlong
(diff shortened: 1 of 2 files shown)


basvodde:
I think so. 2014 is a while ago, but if I look at these and the comments (and think back on it) I think these were my unfinished attempts to have the GMock tests themselves run in CppUTest. I vague remember that that led to some incompatabilities (e.g. things like the death tests in GMock that we don't have (don't want to have) in CppUtest).

I am not sure the new GTest versions will be working though. In the past, every gtest version there was some work (read: hacking) to get that to work.

thetic:
Are CppUTest's tests insufficient to validate the GTest integration?

basvodde:
They are invalid to test is all of GTest functionality works within CppUTest.

What the CppUTest/GTest integration does is that it tries to discover the GTests that are linked in the same binary and wraps them around a CppUTest TestShell. It also takes over the GMock assertion handler if I'm right. That way it can execute GTest cases as if they are CppUTest cases.

However, that does not test all the Gtest/GMock functionality within CppUTest. In order to do that, it would be better to take the GTest self-tests and run those in CppUTest. That failed on some functionality, which means that any basic usages of GTest within CppUTest will work... but that some functionality might cause problems.

For me, at the time of writing it, that was good enough and I didn't continue it further. It was also hard since GTest isn't really build to be part of another test runner, so it is definitively a hack on the GTest-side, which is why it has traditionally not  been easy to adopt a new version. That said, I guess that part of Gtest doesn't change that often...

thetic:
Is there any value in merging this given your points?

thetic:
> [Existing automation] does not test all the Gtest/GMock functionality within CppUTest. In order to do that, it would be better to take the GTest self-tests and run those in CppUTest. That failed on some functionality, which means that any basic usages of GTest within CppUTest will work... but that some functionality might cause problems.

It sounds like we could use some integration tests that exercise the subset of functionality that we expect to work.

basvodde:
Yes, I think so. The GTestTests ought to have some GTest tests that ought to be run in CppUTest. If they pass then the basic GTest/GMock functionality ought to work in CppUTest.

In my experience, for most Gtest-test-sets, they don't use more than the basic functionality anyways, so it is fine.

PR #1668, opened 2022-11-05
Don't guess CHAR_BIT

thetic:
This change requires use-cases without the standard library to provide their appropriate byte size. Given the commitment to maintain support for use without standard C, I think we need to be more prudent about the assumptions we're making about those environments.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -88,8 +88,8 @@
       "name": "no-std-c",
       "inherits": ["GNU"],
       "environment": {
-        "CXXFLAGS": "-Werror -nostdinc",
-        "CFLAGS": "-Werror -nostdinc"
+        "CXXFLAGS": "-DCHAR_BIT=8 -Werror -nostdinc",
+        "CFLAGS": "-DCHAR_BIT=8 -Werror -nostdinc"
       },
       "cacheVariables": {
         "CMAKE_SYSTEM_NAME": "Generic",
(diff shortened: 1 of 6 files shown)


basvodde:
Interesting. Before merging, why did you chose to remove the CPPUTEST_CHAR_BIT? The same could be accomplished by adding the #error to the #else in de #if STDCLIB ? Just curious.

thetic:
I didn't see much value in the indirection. When the std C library isn't used,  we don't have to worry about collisions. A quick scan of the Linux kernel shows some use of (a custom-defined) `CHAR_BIT` already, so my hope was to avoid impact on such use cases.

basvodde:
Ok, for me, it feels as a design inconsistency. So far, we've decided to e.g. keep the StdC functions out of the CppUTest code by wrapping them. Alternatively, we could have provided the missing standard C functions when they were not there. There we chose for explicitly wrapping them rather than replacing them.

This one means we are now creating a dependency on the StdC interface in the CppUtest code. It is not a big problem (and I do intent do merge the PR) but it does feel a bit inconsistent with the general design.

thetic:
Ok. I'll restore use of `CPPUTEST_CHAR_BIT`, but drop the old "guess."

basvodde:
Ok. I would have merged it :) Since you've been doing a lot of contributions, I'll try to explain more often the reasoning on why things are the way they are.

thetic:
Thanks! I don't mind doing rework to keep things consistent.

PR #1670, opened 2022-11-07
Fixed compilation when exceptions are enabled but RTTI not.

dmitrykos:
This PR fixes regression introduced by 3756d31e09e039662ee81e4672634ba8d7b15bea by excluding `typeid()` RTTI function from the compilation if compiler settings did not enable RTTI support that is a fairly common case for Embedded development when C++ exceptions are enabled while RTTI not to balance between functionality and binary size.

### DIFF
--- a/src/CppUTest/TestFailure.cpp
+++ b/src/CppUTest/TestFailure.cpp
@@ -395,6 +395,7 @@ UnexpectedExceptionFailure::UnexpectedExceptionFailure(UtestShell* test)
 
 static SimpleString getExceptionTypeName(const std::exception &e)
 {
+#if defined(_CPPRTTI) || defined(__GXX_RTTI)
     const char *name = typeid(e).name();
 #if defined(__GNUC__) && (__cplusplus >= 201103L)
     int status = -1;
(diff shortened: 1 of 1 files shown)


src/CppUTest/TestFailure.cpp
```diff
@@ -395,6 +395,7 @@ UnexpectedExceptionFailure::UnexpectedExceptionFailure(UtestShell* test)
 
 static SimpleString getExceptionTypeName(const std::exception &e)
 {
+#if defined(_CPPRTTI) || defined(__GXX_RTTI)
```
basvodde, line 398: Ok. I'll merge this... but these defines will need to go to the CppUTestConfig and not here like this. This will not be sustainable on the long run.

dmitrykos:
@basvodde thank you for merging! Although I could force update this PR I created a new one (#1671) with the proposed change.

src/CppUTest/TestFailure.cpp
dmitrykos: Ok, made changes via #1671.

PR #1671, opened 2022-11-07
Refactored RTTI feature detection and moved it to the config header.

dmitrykos:
In response to #1670 cleans up the fix by moving RTTI detection into the config header.

### DIFF
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -239,6 +239,17 @@
 #endif
 
 /*
+ * Detection of run-time type information (RTTI) presence
+ */
+#ifndef CPPUTEST_HAVE_RTTI
+  #if defined(_CPPRTTI) || defined(__GXX_RTTI) || defined(__RTTI) || (defined(__cpp_rtti) && __cpp_rtti)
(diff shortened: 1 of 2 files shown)


dmitrykos:
Had to cleanup PR due to outdated source on my side, now looks fine.

include/CppUTest/CppUTestConfig.h
```diff
@@ -238,6 +238,17 @@
 #endif
 #endif
 
+/*
+ * Detection of run-time type information (RTTI) presence
+ */
+#ifndef CPPUTEST_HAVE_RTTI
+  #if defined(_CPPRTTI) || defined(__GXX_RTTI)
```
thetic, line 245: please add `__RTTI` for green hills

thetic:
RTTI support can be directly inspected with C++20 https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros

dmitrykos:
@thetic thank you for the proposal, added `__cpp_rtti`. I did not add a check for its value but checked for presence only similar to GCC's `libstdc++`.

basvodde:
Thanks, that looks much better.

PR #1673, opened 2022-11-12
Refresh CMake test discovery

thetic:
The existing implementation does not handle multi-config setups (generators that can build Debug, Release, etc. from the same project like Visual Studio).

Consider the following example. No tests should run because _Debug_ hasn't been built. The current discovery automation, however, just runs whatever was built last.

```console
$ cmake --preset MSVC
$ cmake --build cpputest_build --config Release
$ ctest --test-dir cpputest_build -C Debug
```


cmake/Modules/CppUTestBuildTimeDiscoverTests.cmake
```diff
@@ -4,6 +4,10 @@ set(_DISCOVER_SCRIPT "${CMAKE_CURRENT_LIST_DIR}/../Scripts/CppUTestBuildTimeDisc
 
 # Create target to discover tests
 function (cpputest_buildtime_discover_tests tgt)
```
thetic, line 6: Can we delete this file? What's the workflow for breaking changes?

cmake/Modules/CppUTestNormalizeTestOutputLocation.cmake
```diff
@@ -1,10 +0,0 @@
-# Override output properties to put test executable at specificied location
```
thetic, line 1: This created collisions between configs, by using the same output path for all of them:

```sh
cmake --preset MSVC
cmake --build cpputest_build --config Release
cmake --build cpputest_build --config Debug # Overwrites Release outputs
ctest --test-dir cpputest_build -C Release # Actually runs Debug tests
```

basvodde:
Uhm, the PR has a conflict

cmake/Modules/_CppUTestDiscovery.cmake
```diff
@@ -0,0 +1,58 @@
...
+foreach(test_name IN LISTS discovered_tests)
+    add_command(
+        add_test
+        "${test_name}"
+        ${EMULATOR}
+        "${EXECUTABLE}"
+        ${ARGS}
+        ${select_arg}
+        ${test_name}
+
```
thetic, line 54:
```suggestion
```

Issue #1634, opened 2022-09-08
Compile on Windows with Clang 14.0.6


itavero:
I'm getting a similar error as in #1405 when trying to build the current `master` branch on my Windows machine with Clang 14.0.6 (via CMake / ninja).
Example of the error I got:
```
[build] D:/gitrepo/cpputest/src/Platforms/Gcc/UtestPlatform.cpp:248:11: error: 'fopen' is deprecated: This function or variable may be unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [-Werror,-Wdeprecated-declarations]
[build]    return fopen(filename, flag);
[build]           ^
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\stdio.h:212:20: note: 'fopen' has been explicitly marked deprecated here
[build]     _Check_return_ _CRT_INSECURE_DEPRECATE(fopen_s)
[build]                    ^
[build] C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\vcruntime.h:269:55: note: expanded from macro '_CRT_INSECURE_DEPRECATE'
[build]         #define _CRT_INSECURE_DEPRECATE(_Replacement) _CRT_DEPRECATE_TEXT(    \
[build]                                                       ^
[build] C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\vcruntime.h:259:47: note: expanded from macro '_CRT_DEPRECATE_TEXT'
[build] #define _CRT_DEPRECATE_TEXT(_Text) __declspec(deprecated(_Text))
[build]                                               ^
```

I fixed the compilation itself by changing the pre-processor check for `fopen_s` and `localtime_s` from:
```c
#if defined(_WIN32) && defined(MINGW_HAS_SECURE_API)
```
to
```c
#if defined(_WIN32) && (defined(__clang__) || defined(MINGW_HAS_SECURE_API))
```

The next failure I run into is:
```
[build] CMake Error at D:/gitrepo/cpputest/cmake/Scripts/CppUTestBuildTimeDiscoverTests.cmake:49 (message):
[build]   Executable "D:/gitrepo/cpputest/build/tests/CppUTest/CppUTestTests.exe -lg"
[build]   failed with output:
[build] 
[build]   Access violation
[build] 
[build]   Please check that the excutable was added.
[build] Call Stack (most recent call first):
[...]

Issue #1622, opened 2022-08-25
FAIL: CppUTestExtTests.exe


javier-acosta-m:
============================================================================
Testsuite summary for CppUTest 4.0
============================================================================
# TOTAL: 2
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
Please report to https://github.com/cpputest/cpputest
============================================================================
make[3]: *** [Makefile:3283: test-suite.log] Error 1
make[3]: Leaving directory '/home/jacosta/cpputest/cpputest_build'
make[2]: *** [Makefile:3391: check-TESTS] Error 2
make[2]: Leaving directory '/home/jacosta/cpputest/cpputest_build'
make[1]: *** [Makefile:3613: check-am] Error 2
make[1]: Leaving directory '/home/jacosta/cpputest/cpputest_build'
make: *** [Makefile:4107: check_coverage] Error 2

javier-acosta-m:
[test-suite.log](https://github.com/cpputest/cpputest/files/9428546/test-suite.log)

basvodde:
Trunk version ?

javier-acosta-m:
$ git show
commit 7c53161fd236f8a518965cf8c8bbd5a594a9d369 (HEAD -> master, origin/master, origin/HEAD)
Merge: fb5e9c4a d5f36280
Author: Bas Vodde <basv@odd-e.com>
Date:   Thu Aug 25 16:28:09 2022 +0200

    Merge pull request #1583 from thetic/cmake-macros

    Simplify config macros

basvodde:
Ok, what is your environment? This works for me and in the automated build, but it is something environment specific.

javier-acosta-m:
I recompiled on MSYS2 mingw64

basvodde:
Using CMake or configure ?

javier-acosta-m:
CMake

thetic:
I suspect the automatic fenv support check isn't working as intended here. I'll get a windows/gcc build going in CI and see if I can repro.

thetic:
Oddly, I can reproduce this in MSYS2 with autotools, but not CMake. #1623

thetic:
> I recompiled on MSYS2 mingw64

[...]

Issue #1381, opened 2020-05-09
Fix the cmake windows travis build


basvodde:
cmake_windows in travis_ci_build.

Compilation passes, but it causes an access violation during test discovery.

Issue #1279, opened 2020-03-06
Memory leak tables corrupted when calling DLL


uecasm:
I haven't worked out precisely why as yet, but given the following code (on Windows VS2017 with Boost 1.68 using DLLs for both the CRT and Boost, and in Debug config):

```
#include <sstream>
#include <boost/filesystem.hpp>

#include <CppUTest/TestHarness.h>
#include <CppUTestExt/MockSupport.h>
#include <CppUTest/CommandLineTestRunner.h>
#include <CppUTest/TestRegistry.h>
#include <CppUTestExt/MockSupportPlugin.h>
#include <CppUTestExt/MemoryReporterPlugin.h>
#include <CppUTest/MemoryLeakDetector.h>

#include <CppUTest/MemoryLeakDetectorNewMacros.h>
#include <CppUTest/MemoryLeakDetectorMallocMacros.h>

TEST_GROUP(TestTest)
{
};

TEST(TestTest, Test)
{
    for (auto i = 0; i < 100; ++i)
    {
        std::ostringstream str;
        str << "hello world " << i;

        boost::filesystem::path p(str.str());
        if (exists(p))
        {
            FAIL("why did that exist?");
        }
    }
}

int main(int argc, char *argv[])
{
    std::string x("yay");   // force a memory allocation prior to calling the below; see https://github.com/cpputest/cpputest/issues/1278

    MemoryLeakWarningPlugin::turnOnThreadSafeNewDeleteOverloads();

    MockSupportPlugin mockPlugin;
    MemoryReporterPlugin memoryPlugin;
    TestRegistry::getCurrentRegistry()->installPlugin(&mockPlugin);
    TestRegistry::getCurrentRegistry()->installPlugin(&memoryPlugin);
    return CommandLineTestRunner::RunAllTests(argc, argv);
}
```

Running this causes the `MemoryLeakDetectorTable` to get corrupted somehow and it ends up with the internal linked lists pointing into deleted memory, which will eventually cause a crash (often with some entirely unrelated code in the call stack, which made finding this difficult).

(Note that it tends to run successfully in Release builds, presumably because the Debug CRT poisons deleted memory.  It's also less noticeable when there are fewer allocations, which is why it's run in a loop.)

[...]

PR #1676, opened 2022-11-12
Remove test of undefined behavior

thetic:
This hangs in ClangCL.  Calling `delete` on a pointer that wasn't created by a `new` expression is undefined behavior.[1]  There are no valid assertions about post conditions here. Removing this test allows us to remove the hack added in #1641 and restore support for memory leak detection.

[1]: https://en.cppreference.com/w/cpp/language/delete.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,14 +29,10 @@ include(CTest)
 include(CMakeDependentOption)
 option(CPPUTEST_STD_C_LIB_DISABLED "Disable the standard C library")
 
-if((CMAKE_CXX_COMPILER_ID STREQUAL "Clang") AND (CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC"))
-  set(is_clang_cl TRUE)
-endif()
-
 cmake_dependent_option(CPPUTEST_STD_CPP_LIB_DISABLED "Use the standard C++ library"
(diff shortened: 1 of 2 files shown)


basvodde:
Why? The operator delete is overridden in this case where it will check whether the memory is allocated or not. This does not seem like something we want to remove, right?

thetic:
Oof. I think something is wonky with memory on Windows. I don't have a Windows system anymore to debug any of this.

* #1634
* #1622
* #1381
* #1279

thetic:
I can't find anything that would indicate this behavior becomes defined when using custom `delete` or `new` implementations.

thetic:
In this test, `delete` ends up calling `free()` on the invalid pointer.

> The behavior is undefined if the value of ptr does not equal a value returned earlier by `malloc()`,` calloc()`, `realloc()`, or `aligned_alloc()` (since C11). 

https://en.cppreference.com/w/c/memory/free

thetic:
> _[The argument passed to `delete`]_ must be a pointer _[...]_ to a non-array object created by a new-expression, or a pointer to a base subobject of a non-array object created by a new-expression. _[...]_ If expression is anything else, _[...]_ the behavior is undefined.

https://en.cppreference.com/w/cpp/language/delete

PR #1679, opened 2022-11-14
Call abort instead of invoking undefined behavior

thetic:
This started hanging instead of crashing when built with Clang 14.0.6 on my box. Dereferencing a null pointer is not a valid or reliable way to induce a crash.

I also added back the no-std-c build which I guess got lost somewhere. This shook out a bug in my `CPPUTEST_CHAR_BIT` changes.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -130,6 +130,10 @@ jobs:
             os: ubuntu-latest
             cpp_version: 98
             preset: no-rtti
+          - name: No C Standard Library
+            os: ubuntu-latest
+            cpp_version: 98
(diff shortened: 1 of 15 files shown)


basvodde:
I haven't checked this myself,  but does it lead to a proper crash that you can debug (e.g. in gdb you can properly get the stack trace on where it crashed/called abort)? I assume so. That is the intention of the crash method.

thetic:
gdb breaks on signals by default and `abort` raises `SIGABRT`.

basvodde:
Ok, that's what I thought. Just wanted to make sure. Then this is fine!

PR #1682, opened 2022-11-15
Decouple C++ standard library and exception support

thetic:
Like RTTI, exceptions can be disabled independently of the C++ standard library. It is common to disable one or both of RTTI and exceptions while continuing to use remaining supported parts of the standard library in space- or time-constrained embedded systems.


basvodde:
Yes, you are right. This is probably better.

When writing the original support, we've tried to minimize the amount of configurations and hence lumped them all together. In practice, they ought to be seperated.

PR #1683, opened 2022-11-16
Remove backticks

thetic:
This is really dumb, but the `` ` `` character is not technically supported by either the C[1] or C++[2] languages.  Some especially strict compilers and linters will complain.

[1]: https://en.cppreference.com/w/c/language/charset
[2]: https://en.cppreference.com/w/cpp/language/charset


basvodde:
Now, that is an interesting change :)

PR #1685, opened 2022-11-16
Cleanup outdated files

offa:
Removes some outdated / unneeded files.


thetic:
Note that CMake can generate an Eclipse project: https://cmake.org/cmake/help/latest/generator/Eclipse%20CDT4.html.

thetic (approved):

basvodde:
I would like to keep the .travis file though for now. I'm good with removing the Eclipse files. At least i haven't used Eclipse for quite a while now.

basvodde:
I cannot think of a convincing reason, yet I'd prefer keeping it for a while still. I'm impressed with the github actions system that @thetic has spend a lot of time on! In the future, one day, I hope to have time to explore whether to continue the CircleCI direction that I had started or just drop it because Github Actions is reliable enough and has no continuity problem. For that, I suspect that I would like to have a look at the .travisci also. Now I do not have time for that, so I would prefer to keep the travisci config file until we decided to continue/discontinue CircleCI next to the excellent github actions that we have now.

thetic:
I'm not married to any CI system, but the thing I really like about GItHub Actions is that contributors can run the pipeline in their own forks _before_ submitting a PR. I rely on this for iterating on systems or tool chains I can't run locally.

basvodde:
Oh yes, I was not convinced of Github actions at all... before you (@thetic) started making the changes you did. Now I'm more and more convinced that that was the right thing to do!

I do really hope to spend one day just looking at the github actions and CircleCI and other CI systems to see if there is something of value to add beyond the Github Actions. The different processor support that travis had comes to mind as something we might want to add from a different CI system... or just protecting ourselves from Github putting the actions behind a paywall.

Anyways, no concrete plans and because the github actions seem to work fine, no urgency anymore either! But... I would like to keep the travis config file :P

thetic:
Agreed. It definitely has some limitations, but its great for a bunch of fast, parallel checks.

offa:
I have updated the PR :man_shrugging:

PR #1686, opened 2022-11-18
Save map files

thetic:
I was trying to save map files from the ClangCL build to try to debug the hang in `BasicBehavior.deleteWillNotThrowAnExceptionWhenDeletingUnallocatedMemoryButCanStillCauseTestFailures`. Unfortunately ClangCL's linker (link-lld) can't generate map files (and neither can Mac's ld), but I figured it's worth saving them for the builds that can to make future errors easier to diagnose.

See saved map files listed under "Artifacts" here: https://github.com/cpputest/cpputest/actions/runs/3495321337.

Without a Windows system, I won't be able to pursue the problem with `BasicBehavior.deleteWillNotThrowAnExceptionWhenDeletingUnallocatedMemoryButCanStillCauseTestFailures` and ClangCL any further.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -210,6 +210,11 @@ jobs:
         run: cmake --preset ${{ matrix.preset }} ${{ matrix.cmake_args }}
       - name: Build
         run: cmake --build cpputest_build --verbose -j
+      - name: Save map files
+        uses: actions/upload-artifact@v3
+        with:
+          name: "${{ matrix.name }} map files"
+          path: cpputest_build/**/*.map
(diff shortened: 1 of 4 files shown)


basvodde:
Interesting. Is there a way I can see the artifacts for the github actions web UI? I am either too tired to find them or just dumb :)

thetic:
Lol. You're not dumb, They're tucked away: https://github.com/cpputest/cpputest/actions/runs/3557026688

basvodde:
Oh, interesting. Wonder why they didn't put the artifacts at the specific build that produces them

PR #1697, opened 2022-11-25
Improve floating point exception support check

thetic:
Compiler checks are not a reliable indicator of support for floating point exception status flags.  Neither is checking that a program can _compile_ with `fenv.h`; linking can still fail (e.g. Open Watcom). This change checks `__STDC_IEC_559__` instead.  This macro "expands to 1 if IEC 60559 is supported".

Some context/history: #862


basvodde:
Ok... reading this (and not knowing a lot about it), I would have a worry that this is a C99 define (according google) and that it might be implemented in some compilers in the StdC library (noticed also some notifications of that in google). That said, I will merge the pull request.

PR #1702, opened 2022-12-05
Pin Ubuntu version

thetic:
See https://github.com/actions/setup-python/issues/543.


offa:
Isn't this a good change to EOL Python 2.7 here too?

thetic:
That would require EOL for the (old) GTest integration as well.

offa:
> That would require EOL for the (old) GTest integration as well.

Why not head towards CppuTest 5.0 then? :smile:

thetic:
I'm inclined to agree. The GTest integration is very fragile and out of date. I've been struggling to figure out if/how it can be updated and made robust.

offa:
If it's still requires Python 2.7 it's another reason to drop it. While this PR will work for some time, it'll likely break in the future, eg. Python action may work now, but the next Action Image update could drop support.

basvodde:
So far, in CppUTest, we've always tried to keep old things working (hence, e.g. a DOS build) because it tends to be used in larger and often embedded software, which unfortunately often is with old versions of things. So, if the only thing we need to do is to ping the Ubuntu version than I'm for that rather than dropping the support of old GTest versions.

PR #1710, opened 2022-12-07
Mark exception as unused

offa:
If compiled without RTTI `e` is unused and issues a warning. Casting to void marks it as intentionally unused.

Should we create eg. an "unused()" for improved documentation (cast-to-void might not be known to everyone)?


thetic (approved):

thetic:
There's [this](https://github.com/cpputest/cpputest/blob/dd8a04474abaf05f5095644f81a04dbf8be82f75/include/CppUTest/TestHarness_c.h#L237-L249) but it only works in the parameter list and only for a couple compilers.

PR #1713, opened 2022-12-08
Remove mock crash on fail (use the UtestShell crash on fail instead)

CiderMan:
As promised, this removes the Mock-specific crash-on-failure mechanism, using the new core mechanism to support the existing API.

### DIFF
--- a/include/CppUTest/Utest.h
+++ b/include/CppUTest/Utest.h
@@ -184,8 +184,6 @@ protected:
 
     virtual SimpleString getMacroName() const;
     TestResult *getTestResult();
-    static const TestTerminatorWithoutExceptions normalTestTerminatorWithoutExceptions_;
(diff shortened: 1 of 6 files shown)


basvodde:
I'm not sure this is wanted. When I set the crash on failure on CppUMock then it doesn't mean that I want the tests to crash on failure.

I don't use crashOnFailure in the tests a lot... I'm not completely clear why that is needed. However, the crashOnFailure in mock is important to be able to get a stack trace...

CiderMan:
OK, sorry, I read our conversation to mean that you wanted me to remove this from Mock. If that is not the case, I clearly have the wrong end of the stick and you can reject this PR. I don't use this anyway - I was only doing it because I thought you had asked for it! 🤣

basvodde:
Yeah, I think we do not want this as it is removing some flexibility.

If you want, you can fix the issue related to the deallocation memory not crashing.

PR #1717, opened 2022-12-10
Memory Sanitizer

thetic:
> [MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html) is a detector of uninitialized reads.

* We obviously can't run msan with leak detection enabled.
* We have to disable the C++ standard library for msan. Unless we build it ourselves, it isn't instrumented, so generates false positives.
* I had to disable all the MemoryLeakDetector tests for msan. I was surprised to discover that code is active, even when leak detection is disabled.


basvodde:
Why close?

PR #1718, opened 2022-12-10
Split

thetic:


basvodde:
What are you trying to do here? :)

thetic:
This is just something I've been toying around with: segregating the include paths so that we can't `#include "CppUTestExt/..."`  in CppUTest code. This also prevents use of extension headers through the `CppUTest`  CMake target, turning confusing link errors into  compile errors (which are faster and typically clearer).

PR #1719, opened 2022-12-10
Improve Makefile Worker CI

thetic:
* Call _all_ of the makefile setup
* Move logic out of workflow YAML back to the Makefile

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -5,7 +5,7 @@ name: Basic builds
   push:
   pull_request:
   schedule:
-    - cron: '0 5 * * 3'
+    - cron: "0 5 * * 3"
 
 jobs:
   clang-format:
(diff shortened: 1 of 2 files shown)


.github/workflows/basic.yml
```diff
@@ -344,78 +344,39 @@ jobs:
...
-            make_args: >
-              CPPUTEST_USE_GCOV=Y
-              everythingInstall
-            test_cmd: >
-              make -f Makefile_using_MakefileWorker gcov &&
-              make -f Makefile_CppUTestExt gcov &&
-              make -C examples gcov
+            make_target: test_junit_output
+          # - name: gcov
+          #   make_target: test_gcov
```
thetic, line 369: Something is broken with this: https://github.com/cpputest/cpputest/actions/runs/3666346383/jobs/6198003504

PR #1720, opened 2022-12-12
Start porting Travis stuff

thetic:


tests/CppUTest/MemoryOperatorOverloadTest.cpp
```diff
@@ -136,9 +136,10 @@ CPPUTEST_DO_NOT_SANITIZE_ADDRESS
 static void freeInvalidatesMemory()
 {
     unsigned char* memory = (unsigned char*) cpputest_malloc(sizeof(unsigned char));
+    size_t const address = (size_t)memory;
     *memory = 0xAD;
     cpputest_free(memory);
-    CHECK(*memory != 0xAD);
+    CHECK(*(unsigned char*)address != 0xAD);
```
thetic, line 142: This is a hack to avoid reading from a freed pointer, which is undefined behavior. This test has been flakey.

thetic: It didn't work.

basvodde:
haha, this way @offa will still get the travis files deleted like he proposed earlier. You are just doing it gradually. :)

I would prefer keeping them as is for now if that is ok. Yes, that is probably not my rational self and yes they are anyways still in version control... still :)

thetic:
> I would prefer keeping them as is for now if that is ok. 

done

basvodde:
Thanks!

PR #1722, opened 2022-12-12
Hack around undefined use after free

thetic:
This has been flaky lately.


thetic:
Still not working. I don't think this is a valid test.

thetic:
#1725

PR #1723, opened 2022-12-12
Port Docker builds from Travis

thetic:
https://github.com/marketplace/actions/build-and-push-docker-images

If we run into rate limits, we may have to set up a Docker Hub account. I've set it up to conditionally do a login when the credentials are set.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -488,3 +488,20 @@ jobs:
           ./configure
       - name: Build and test
         run: make tdd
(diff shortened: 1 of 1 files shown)


basvodde:
I have tried to create an open source docker hub account in the past and had failed to do so. I couldn't find a good way to do that in the time that I had spend on it :(

offa:
Why not use Github Container Registry instead? It's already included, eg. for this project.

PR #1725, opened 2022-12-14
Remove undefined behavior

thetic:
This is undefined behavior and has been failing very frequently on CI. The custom `delete` implementation calls `free()`.

> The behavior is undefined if after `free()` returns, an access is made through the pointer.

https://en.cppreference.com/w/c/memory/free

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -265,7 +265,9 @@ jobs:
       ### SETUP
 
       - name: Install packages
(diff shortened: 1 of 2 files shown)


offa (approved):

basvodde:
Hrm, I can see why it probably needs to go... it is too bad, the test is very clear

thetic:
Is there anything needed before we merge this? These CI failures are pretty frustrating.

PR #1728, opened 2022-12-15
Remove scheduled workflows

### LINKED ISSUE #1727: Cron are scheduled on forks too
I have noticed that the cron builds are scheduled on forks too.

This should be solved according to [this issue](https://github.com/laravel/framework/issues/34356), but it still happens.
There's a workaround mentioned, but this would require an `if` to each job.

Since disabling the workflow isn't an option, I don't have a solution yet. :man_shrugging:

thetic:
Closes #1727

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -4,8 +4,6 @@ name: Basic builds
 "on":
   push:
   pull_request:
-  schedule:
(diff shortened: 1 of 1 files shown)


offa (approved):

basvodde:
This one is in conflict state now.

PR #1729, opened 2022-12-15
Add support for asnprintf to the MemoryLeakDetector

AlanRosenthal:
see: https://man7.org/linux/man-pages/man3/asprintf.3.html

### DIFF
--- a/src/CppUTest/TestHarness_c.cpp
+++ b/src/CppUTest/TestHarness_c.cpp
@@ -33,6 +33,7 @@
 
 extern "C"
 {
+#include <stdio.h>
 
 void CHECK_EQUAL_C_BOOL_LOCATION(int expected, int actual, const char* text, const char* fileName, size_t lineNumber)
 {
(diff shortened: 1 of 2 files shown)


src/CppUTest/TestHarness_c.cpp
```diff
@@ -33,6 +33,7 @@
 
 extern "C"
 {
+#include <stdio.h>
```
thetic, line 36: Direct reliance on the C standard library is not allowed in CppUTest.

thetic (changes requested):
Clang-tidy warnings need to be addressed.
Build without C standard library needs to be repaired.
We need to get test coverage for this before we can consider merging.

src/CppUTest/TestHarness_c.cpp
thetic: My intuition is that this feature does not belong in CppUTest, but rather in your own force include. I don't see anything in the implementation that requires it to be tightly integrated with CppUTest itself.

`asnprintf` is non-standard; its a GCC/Linux extension. CppUTest is focused on portability across old, embedded, and resource-constrained systems. `asnprintf` isn't portable and I don't see it as especially valuable for CppUTest's target use-cases.

That's just my opinion though.@basvodde, did you have thoughts?

basvodde:
@AlanRosenthal Why close this?

PR #1736, opened 2022-12-29
Abandon attempts to auto-tag and release

thetic:
None of the available GitHub actions allow us to "move" an existing `latest-passing-build` tag. GitHub expects tags to go unmutated. It seems we've been trying to use a tag as a branch.

I think this is fine. We shouldn't be merging anything that doesn't build; master effectively serves the same purpose.


offa (approved):

basvodde:
It is ok for now. I had this working at some point in the far past... but had written a small tool to help me with that. I didn't have time/priority to keep that up to date. So, let's abandon it for now. Thanks for the effort

PR #1738, opened 2023-01-05
Ensure that exception specifications match for delete overloads

### LINKED ISSUE #1737: Build failure on CentOS 7
Firstly some background:

We use C++ exceptions for exploding fakes. The exception is caught in `main()` and a backtrace is produced to help the developer understand the code path that has now caused the exploding fake to be called.

We are updating to a recent version of CppUTest but have to build it with `CPPUTEST_HAVE_EXCEPTIONS` defined as `0`; otherwise the portion of the stacktrace inside the code under test is lost and the backtrace only ever traces down to the `catch(...)` in UTest.cpp.

The problem:

One of the platforms that where we need to use CppUTest is CentOS 7 (yes, it is old but still supported until June 2024) with GCC 4.8.5. When trying to build CppUTest without exceptions, it seems that GCC of that vintage is requiring the correct exception specification for the standard delete overloads. For example:

```
/cue/tm_win32_tools/cpputest/cpputest/src/CppUTest/MemoryLeakWarningPlugin.cpp: In function 'void operator delete(void*)':
/cue/tm_win32_tools/cpputest/cpputest/src/CppUTest/MemoryLeakWarningPlugin.cpp:330:31: error: declaration of 'void operator delete(void*)' has a different exception specifier
 void operator delete(void* mem) UT_NOTHROW
                               ^
In file included from /cue/tm_win32_tools/cpputest/cpputest/include/CppUTest/MemoryLeakWarningPlugin.h:32:0,
                 from /cue/tm_win32_tools/cpputest/cpputest/include/CppUTest/TestHarness.h:45,
                 from /cue/tm_win32_tools/cpputest/cpputest/src/CppUTest/MemoryLeakWarningPlugin.cpp:28:
/cue/tm_win32_tools/cpputest/cpputest/include/CppUTest/MemoryLeakDetectorNewMacros.h:70:10: error: from previous declaration 'void operator delete(void*) noexcept (true)'
     void operator delete(void* mem) UT_NOTHROW;
          ^
```

The solution?

Looking in CppUTestConfig.h, it looks like Clang might have a similar requirement:

```cxx
  #if CPPUTEST_HAVE_EXCEPTIONS
    [...snip...]
  #else
    #define UT_THROW(exception)
    #ifdef __clang__
      #define UT_NOTHROW throw()
    #else
      #define UT_NOTHROW
    #endif
  #endif
```

If I extend the pattern to include GCC, it then builds successfully and cleanly for both the old version of GCC and more recent versions:

```cxx
  #if CPPUTEST_HAVE_EXCEPTIONS
     [...snip...]
  #else
    #define UT_THROW(exception)
    #if defined(__clang__) || defined(__GNUC__)
      #define UT_NOTHROW throw()
    #else
      #define UT_NOTHROW
    #endif
  #endif
```

I will create a PR with this change so that it can be discussed and/or accepted as appropriate.

Thanks,

Steve.

CiderMan:
Fixes #1737

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -281,7 +281,7 @@ jobs:
       - name: Install QEMU
         # The version in the ubuntu repositories (6.2) is broken.
         run: |
-          wget -nv http://ftp.de.debian.org/debian/pool/main/q/qemu/qemu-user-static_7.1+dfsg-2+b3_amd64.deb -O qemu.deb
+          wget -nv http://ftp.de.debian.org/debian/pool/main/q/qemu/qemu-user-static_7.1+dfsg-2~bpo11+3_amd64.deb -O qemu.deb
           sudo dpkg --install qemu.deb
           rm -f qemu.deb
         if: ${{ matrix.name == 'GCC ARM embedded' }}
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -281,7 +281,7 @@ jobs:
       - name: Install QEMU
         # The version in the ubuntu repositories (6.2) is broken.
         run: |
-          wget -nv http://ftp.de.debian.org/debian/pool/main/q/qemu/qemu-user-static_7.1+dfsg-2~bpo11+3_amd64.deb -O qemu.deb
+          wget -nv http://ftp.de.debian.org/debian/pool/main/q/qemu/qemu-user-static_7.1+dfsg-2+b3_amd64.deb -O qemu.deb
           sudo dpkg --install qemu.deb
           rm -f qemu.deb
         if: ${{ matrix.name == 'GCC ARM embedded' }}
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -217,8 +217,12 @@
     #endif
   #else
     #define UT_THROW(exception)
-    #ifdef __clang__
-      #define UT_NOTHROW throw()
+    #if defined(__clang__) || defined(__GNUC__)
+      #if defined(__cplusplus) && __cplusplus >= 201103L
+        #define UT_NOTHROW noexcept
+      #else
+        #define UT_NOTHROW throw()
+      #endif
     #else
       #define UT_NOTHROW
     #endif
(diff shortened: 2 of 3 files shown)


CiderMan:
@basvodde Is there a problem with the 'CMake GCC ARM embedded (pull_request)' build? It seems to have failed with a 404 error long before getting close to the CppUTest code but that has resulted in a load of other (also relevant) tests being cancelled...

CiderMan:
> @basvodde Is there a problem with the 'CMake GCC ARM embedded (pull_request)' build? It seems to have failed with a 404 error long before getting close to the CppUTest code but that has resulted in a load of other (also relevant) tests being cancelled...

It seems like there has been work in the QEMU area in the last week or two, meaning that the URL for the .deb is no longer valid. I've picked what looks like the closest and it now runs. I am happy to revert the change, if you would like to deal with updating the workflows separately. The new failure (CMake Linux Clang 14) fails on the current master branch so is not due to my change.

basvodde:
@thetic What do you think? Ack?

CiderMan:
Just to update, the failure was not the same as on master (one was Apple and the other Linux and similar, but different, tests) but I think that it is the same root cause. The tests are potentially racey as they access memory after it is freed so I wonder if that is what causes occasional failures. I did notice that there was a typo in one of the tests so have fixed that while in there but I don't not think that the typo was the cause of the failure.

thetic:
#1725

CiderMan:
> #1725

Should I revert the commits addressing the Qemu and the typo in the test so that this PR can be delivered or are you happy to deliver this as-is and subsequently deal with them on `master`?

CiderMan:
> I would roll back your qemu changes and cherry-pick #1739. It has proven too fragile to try to maintain.
> 
> I'm not sure about #1725. It's been open for a while.

OK. I've reverted the Qemu change - I'll leave #1739 to be delivered separately as it is basically unrelated. I've left the typo fix in place as it is correct (I believe) until such time as the tests are removed or changed by #1725. Hopefully that will allow this PR to be merged.

basvodde:
Ok, thanks

PR #1739, opened 2023-01-05
Remove qemu from CI

thetic:
Our workaround to download a new version on Ubuntu keeps breaking because the Debian URL is not stable.

- #1700
- #1738


basvodde:
Ok. Wouldn't it then preferred to just update the URL regularly ? Or does it change that often?

thetic:
it seems to change roughly once a month

PR #1756, opened 2023-07-16
Minor readme fix for autoconf install instructions

MitchellThompkins:
I copied/pasted the install instructions and it failed. Seems like there's just a typo in the readme.


basvodde:
It should be cpputest_build... it is now fixed though. Thanks!

PR #1763, opened 2023-10-28
Ci repair

thetic:
- Remove gtest builds. They depend on Python 2 which is no longer available in GitHub actions.
- Suppress ClangCL warning for "unsafe buffer usage" that flags all the raw string operations.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -63,27 +63,6 @@ jobs:
             cc: clang
             cxx: clang++
             make_args: tdd
-          - name: GTest 1.5
-            os: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


offa (approved):

basvodde:
Thanks for still paying attention after my lack of attention the last months. Is there no better solution than this?

thetic:
Maybe, but I'd like to get CI passing in the meantime. I don't have much interest in trying to resurrect Python 2.

SteelPh0enix:
can someone merge it please? would love to see a new build in releases section!

basvodde:
Ok....

PR #1779, opened 2024-03-31
Include headers in clang-tidy checks

thetic:
I discovered that clang-tidy hadn't been checking headers. This PR adjusts settings to make sure they are and resolves all the newly uncovered warnings.


basvodde:
That is a huge PR.

basvodde:
Thanks

PR #1783, opened 2024-04-04
Suppress Sanitization Warning on Linux When Memory Leak Detection is Disabled

hkpeprah:
### Overview

There is a warning in the `CppUTestConfig.h` on Linux platforms when building with ASAN. The warning says to disable memory leak detection in order to get rid of the warning with `-DCPPUTEST_MEM_LEAK_DETECTION_DISABLED`. However, providing said define does not get rid of the warning. This prevents compilation on Linux with ASAN + `-Werror`.

### Proposed Change

1. If `CPPUTEST_MEM_LEAK_DETECTION_DISABLED` do not display warning about address sanitization.

### DIFF
--- a/include/CppUTest/CppUTestConfig.h
+++ b/include/CppUTest/CppUTestConfig.h
@@ -146,7 +146,7 @@
 #endif
 
 #if CPPUTEST_SANITIZE_ADDRESS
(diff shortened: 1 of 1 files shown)


basvodde:
Ah ok. That makes sense.

PR #1785, opened 2024-04-23
Avoid missing include directory warnings

### LINKED ISSUE #1784: Can't compile CppUTest with GCC when `-Werror=missing-include-dirs` is set
Hi,
it seems that CppUTest is not able to compile with GCC when `-Werror=missing-include-dirs` is set, as the directory `include/Platforms/Gcc` does not exist:
```
cc1plus.exe: error: C:/mydir/.build/gcc-x86_64-w64-mingw32/Release/_deps/cpputest-src/src/CppUTest/../../include/Platforms/Gcc: No such file or directory [-Werror=missing-include-dirs]
```
(built with Windows 10, GCC from w64devkit 1.17.0, cmake 3.27.4)

As far as I see, there are two possible solutions:
1. Remove the missing directory from the following places:
  - https://github.com/cpputest/cpputest/blob/49bae1944af67053f7560df3e99cbefc3e120b83/scripts/templates/ProjectTemplate/ProjectMakefile#L35
  - https://github.com/cpputest/cpputest/blob/49bae1944af67053f7560df3e99cbefc3e120b83/platforms/Eclipse-Cygwin/.project#L1188
  - Find a workaround for CMake: https://github.com/cpputest/cpputest/blob/49bae1944af67053f7560df3e99cbefc3e120b83/src/CppUTest/CMakeLists.txt#L58
  - Possible workaround for CMake: only reffering to this directory if it exists
2. Create a dummy folder `include/Platforms/Gcc`  with an empty `.gitkeep` file in it (to ensure that it is kept by git)
  - Do this for the other major platforms mentioned in https://github.com/cpputest/cpputest/blob/49bae1944af67053f7560df3e99cbefc3e120b83/CMakeLists.txt#L68 as well

thetic:
Only c2000 even uses this include path anymore, so we can remove it for everything else.

Closes #1784

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -78,7 +78,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@main
-      - run: brew install automake
+      - run: brew install automake libtool
         if: ${{ startswith(matrix.os, 'macos') }}
       - if: ${{ matrix.cxx }}
         run: echo "CXX=${{ matrix.cxx }}" >> $GITHUB_ENV
--- a/CppUTest.vcproj
+++ b/CppUTest.vcproj
@@ -41,7 +41,7 @@
 			<Tool
 				Name="VCCLCompilerTool"
 				Optimization="0"
-				AdditionalIncludeDirectories=".\include\Platforms\VisualCpp,.\include"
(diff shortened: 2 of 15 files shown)


basvodde:
Thanks (again)!

PR #1800, opened 2024-07-31
Do not define __has_attribute in a public header

hwti:
This causes issues with gcc < 5 and gmacros.h from glib >= 2.69.
The glib header thinks the compiler supports __has_attribute, so it uses it instead of doing version checks.
Then macros like G_GNUC_UNUSED become empty, as if no attribute was supported.


thetic:
This is a good change. We shouldn't be defining reserved identifiers. Thank you.

PR #1805, opened 2024-09-02
Disable leak detection on Mac builds

thetic:
See https://github.com/cpputest/cpputest/actions/runs/10673028874/job/29581484547


basvodde:
Ah, this is the commit. Disabling something when it isn't working is not the right action.

thetic:
It's not working because [the implementation](https://github.com/cpputest/cpputest/blob/df7e2a9fb737300f642a9bc73b880398a992dcbd/include/CppUTest/MemoryLeakDetectorNewMacros.h#L87) invokes [undefined behavior](https://en.cppreference.com/w/cpp/preprocessor/replace#Reserved_macro_names). I don't know how to work around this.

basvodde:
I made a PR to fix it. It is caused by the realloc #define. The way around it is the same as in the manual for dealing with STL.

PR #1806, opened 2024-09-02
Remove excessive warning suppressions

thetic:
We removed the last use of reserved macro names in #1800 and have a local suppression for the single keyword macro.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -232,10 +232,10 @@ AC_MSG_CHECKING([whether CC support -Wreserved-id-macro])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wreserved-id-macro"], [AC_MSG_RESULT([no])])
 CFLAGS="$saved_cflags"
 
-# Flag -Wno-keyword-macro.
-CFLAGS="-Werror -Wkeyword-macro"
-AC_MSG_CHECKING([whether CC supports -Wkeyword-macro])
-AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wkeyword-macro"; CPPUTEST_CXXWARNINGFLAGS="${CPPUTEST_CXXWARNINGFLAGS} -Wno-keyword-macro" ], [AC_MSG_RESULT([no])])
+# Flag -Wreserved-identifier
+CFLAGS="-Werror -Wreserved-identifier"
+AC_MSG_CHECKING([whether CC supports -Wreserved-identifier])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wreserved-identifier"; CPPUTEST_CXXWARNINGFLAGS="${CPPUTEST_CXXWARNINGFLAGS} -Wreserved-identifier" ], [AC_MSG_RESULT([no])])
 CFLAGS="$saved_cflags"
 
 ##### C++ Warnings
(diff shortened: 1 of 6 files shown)


thetic:
Pipeline depends on fixes from #1805

.github/workflows/basic.yml
```diff
@@ -204,21 +204,23 @@ jobs:
             os: macos-latest
             preset: defaults
             install: true
+            # Undefined-behavior in leak detection breaks #include<list> on Mac
+            cmake_args: -DCPPUTEST_MEM_LEAK_DETECTION_DISABLED=YES
```
basvodde, line 208: I think disabling the memory leak detection here does not seem the right thing to do.

thetic: This change was from #1805. I needed it to get a passing pipeline. I've removed it from this MR.

configure.ac
```diff
@@ -226,6 +226,18 @@ AC_MSG_CHECKING([whether CC supports -Wsign-conversion])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wsign-conversion"], [AC_MSG_RESULT([no])])
 CFLAGS="$saved_cflags"
 
+# Flag -Wreserved-id-macro
+CFLAGS="-Werror -Wreserved-id-macro"
+AC_MSG_CHECKING([whether CC support -Wreserved-id-macro])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wreserved-id-macro"], [AC_MSG_RESULT([no])])
+CFLAGS="$saved_cflags"
+
+# Flag -Wno-keyword-macro.
```
basvodde, line 235: typo in comment.

Do we need to enable these explicitly? Aren't they included in some of the warning sets like -Wall ?

thetic: Good catch. I've switched to the broader, non-default `-Wreserved-identifier`.

configure.ac
```diff
@@ -232,10 +232,10 @@ AC_MSG_CHECKING([whether CC support -Wreserved-id-macro])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wreserved-id-macro"], [AC_MSG_RESULT([no])])
 CFLAGS="$saved_cflags"
 
-# Flag -Wno-keyword-macro.
-CFLAGS="-Werror -Wkeyword-macro"
-AC_MSG_CHECKING([whether CC supports -Wkeyword-macro])
-AC_COMPILE_IFELSE([AC_LANG_PROGRAM([])], [AC_MSG_RESULT([yes]); CPPUTEST_CWARNINGFLAGS="${CPPUTEST_CWARNINGFLAGS} -Wkeyword-macro"; CPPUTEST_CXXWARNINGFLAGS="${CPPUTEST_CXXWARNINGFLAGS} -Wno-keyword-macro" ], [AC_MSG_RESULT([no])])
+# Flag -Wreserved-identifier
```
basvodde, line 235: On this one. Let me add a bit more detail.

We often need to add the -Wno because the warning is included in one of the warning-set parameters, such as -Wall -Wpedantic etc. Because they are included in a set, we will need to exclude them explicitly.

However, when we do not need to exclude them then it doesn't mean we should include them (as that is default). So, if the -Wreserved-identifier is a part of the other warning collections, then I would suggest to remove the explicit checking and turning on of this warning and instead rely on the warning group. That means there is less (configure) code and it is more resistent to changes.

So, I'd suggest to check that. If they are included, then delete these checks and delete the explicit inclusion of these warnings.

Or am I missing something here?

thetic: `-Wreserved-identifier`[^1] is not enabled by default. It enables a set of other checks, only some of which are enabled by default.

[^1]: https://clang.llvm.org/docs/DiagnosticsReference.html#wreserved-identifier

basvodde: Oh yes, I'm not doubting whether it needs to be on, just wondering whether it is on via one of the other flags or not.

Diving into the clang manual suggests that of reversed-identifier, only the reserved-macro-identifier is not default and I couldn't find it as part of -Wall and -Wpadantic, so I guess the flag is needed and the check is needed. Thanks.

basvodde: Thinking this over more. Isn't this one still needed? We still do have a #define new in the code (for the memory leak detector) and that would still trigger an unwanted warning ?

thetic: Those suppressions are already in the code with minimal scope. https://github.com/search?q=repo%3Acpputest%2Fcpputest%20pragma&type=code

basvodde: Ah ✅

basvodde:
Thanks!

PR #1807, opened 2024-09-03
Fix realloc build failure with clang

basvodde:
Ensuring the list include works when overloading the realloc


basvodde:
Excellent, thanks to myself.

PR #1810, opened 2024-09-09
Revert "Encode newlines for -ojunit as their XML equivalent instead of {newline}"

basvodde:
Reverts cpputest/cpputest#1598


basvodde:
Broke the build in a non-trivial way

basvodde:
Fixed in other way. Closing revert

PR #1830, opened 2024-11-23
Resolve new CMake warning

thetic:
CMake 3.31 added a new warning for the added [CMP0175](https://cmake.org/cmake/help/latest/policy/CMP0175.html) policy which uncovered the previously ignored unused `DEPENDS` argument.


basvodde:
Tnx

PR #1833, opened 2024-12-28
Squelch deprecation warning

thetic:
Most recent CMake started complaining about the pending end of 3.8 support.

> ```
> CMake Deprecation Warning at CMakeLists.txt:4 (cmake_minimum_required):
>   Compatibility with CMake < 3.10 will be removed from a future version of
>   CMake.
>
>   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
>   to tell CMake that the project requires at least <min> but has been updated
>   to work with policies introduced by <max> or earlier.
> ```

We can retain support by adding a max version. This will not prevent use with newer versions, but indicates forward compatibility.[^1]

[^1]: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # 2.6.3 is needed for ctest support
 # 3.1 is needed for target_sources
 # 3.8 is needed for try_compile improvements (CMP0067)
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.8...3.31)
 
 project(CppUTest
   VERSION 4.0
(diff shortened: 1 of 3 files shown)


offa:
Why not use CMake > 3.8? 3.8 was released 2018 and there's barely a reason to stay on such an old version.

thetic:
> Why not use CMake > 3.8? 3.8 was released 2018 and there's barely a reason to stay on such an old version.

I considered that, but we're not using anything that requires anything newer, and this is easy enough to fix without breaking compatibility. I agree, however, that we should not hesitate to bump the minimum version if we ever want a newer CMake behavior.

PR #1834, opened 2024-12-28
Resolve new clang-tidy warning

thetic:
https://clang.llvm.org/extra/clang-tidy/checks/bugprone/casting-through-void.html

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,10 @@
 ---
-Checks: "clang-diagnostic-*,clang-analyzer-*,bugprone-*,-bugprone-easily-swappable-parameters"
+Checks:
+  - clang-diagnostic-*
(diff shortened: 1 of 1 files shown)


offa (approved):

thetic:
These changes are confusing lcov. I assume this changes the type of cast that is being made.

@basvodde is there a reason that C++-style casts aren't used anywhere? I think the explicit control would be rather useful, but I presume you had a reason to avoid them.

basvodde:
Yes, it is the same reason as not using templates and other 'newer' or 'advanced' C++ features. We want CppUTest to be able to work with old compilers because embedded systems often have old compilers or bad C++ support.

thetic:
Coverage build has rotted, but this repairs the tidy job.

PR #1835, opened 2025-01-02
 Add malloc.h into StandardCLibrary.h.

ahuo-dev:
Add malloc.h into StandardCLibrary.h. Some C code include malloc.h for malloc func.

### DIFF
--- a/include/CppUTest/StandardCLibrary.h
+++ b/include/CppUTest/StandardCLibrary.h
@@ -15,12 +15,13 @@
  #if CPPUTEST_USE_STD_CPP_LIB
   #include <cstdlib>
   #include <string>
+  #include <malloc.h>
  #endif
 #endif
 
 /* Needed for malloc */
 #include <stdlib.h>
-
+#include <malloc.h>
 /* Needed for std::nullptr */
 #ifdef __cplusplus
  #if CPPUTEST_USE_STD_CPP_LIB


thetic:
Per the standard, `malloc` is defined in `stdlib.h`[^1] which is already included.

[^1]: https://en.cppreference.com/w/c/memory/malloc

basvodde:
@ahuo-dev Could you share what environment you were trying to solve what problem, then we can look at whether this is the best fix.

ahuo-dev:
> Per the standard, `malloc` is declared in `stdlib.h`[1](#user-content-fn-1-e0e312372e8b66f074c3d8d5c80107c2) which is already included.
> 
> ## Footnotes
> 1. https://en.cppreference.com/w/c/memory/malloc [↩](#user-content-fnref-1-e0e312372e8b66f074c3d8d5c80107c2)



> @ahuo-dev Could you share what environment you were trying to solve what problem, then we can look at whether this is the best fix.

> @basvodde My project use amazonaws kinesis video stream code. I have a compile error when enable cpputest memory leak check . and I found amazonaws kinesis video stream code use malloc header with malloc.h. see this code line 523 at https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h

basvodde:
Ok, but the include of #malloc.h in the CommonDef.h shouldn't need to trigger you to include it in the SatndardCLibrary.h, right?

Could you share what the compile error is that you are getting?

ahuo-dev:
> Ok, but the include of #malloc.h in the CommonDef.h shouldn't need to trigger you to include it in the SatndardCLibrary.h, right?
> 
> Could you share what the compile error is that you are getting?

Compile error message :
"nclude/CppUTest/MemoryLeakDetectorMallocMacros.h:36:47: error: expected declaration specifiers or ‘...’ before string constant
   36 | #define malloc(a) cpputest_malloc_location(a, __FILE__, __LINE__)"

and run command "gcc -E " to expand all macro. I see malloc function prototype expand to cpputest_malloc_location, like this :
extern void *
# 38 "/usr/include/malloc.h"
            cpputest_malloc_location(
# 38 "/usr/include/malloc.h" 3 4
            size_t __size
# 38 "/usr/include/malloc.h"
            , "/usr/include/malloc.h", 38)
# 38 "/usr/include/malloc.h" 3 4
                                   __attribute__ ((__nothrow__ , __leaf__)) __attribute__ ((__malloc__))
     __attribute__ ((__alloc_size__ (1))) ;

Normally it should be like this:
extern void *malloc (size_t __size) __THROW __attribute_malloc__
     __attribute_alloc_size__ ((1)) __wur;

Issue #1822, opened 2024-09-27
old Visual C++ builds are broken.


thetic:
#1808 broke the old Visual C++ builds in AppVeyor. I didn't notice when I opened the MR. Since it merged all pipelines are failing.

I don't have a Windows machine or know Visual C++ well enough to understand why it broke or how to fix it.

https://ci.appveyor.com/project/basvodde/cpputest/builds/50614045

PR #1837, opened 2025-01-12
configure.ac: fix call of AC_LANG_PROGRAM & unexpected compilation er…

clan:
…rors

1. multiple line should be enclosed by [[ ]], otherwise compilation of conftest.cpp will failed with: error: extra tokens at end of #include directive [-Werror,-Wextra-tokens]
2. add #include <cstdint>
3. size_t -> std::size_t


basvodde:
Thanks

thetic:
> Is the test failure on Windows related with this PR?

No https://github.com/cpputest/cpputest/issues/1822

PR #1839, opened 2025-01-13
update README.md fetchcontent cmake section remove test disable

mtfurlan:
dd70ae595cdd7374bd8e160e5d53fe9510904478 renamed TESTS to
CPPUTEST_BUILD_TESTING and defaulted it to off, so we no longer need to
set it to off in the example

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,8 +248,6 @@ FetchContent_Declare(
     GIT_REPOSITORY https://github.com/cpputest/cpputest.git
     GIT_TAG        master # or use release tag, eg. v4.0
 )
-# Set this to ON if you want to have the CppUTests in your project as well.
-set(TESTS OFF CACHE BOOL "Switch off CppUTest Test build")
 FetchContent_MakeAvailable(CppUTest)
 ```
 


README.md
```diff
@@ -249,7 +249,7 @@ FetchContent_Declare(
     GIT_TAG        master # or use release tag, eg. v4.0
 )
 # Set this to ON if you want to have the CppUTests in your project as well.
-set(TESTS OFF CACHE BOOL "Switch off CppUTest Test build")
+set(CPPUTEST_BUILD_TESTING OFF CACHE BOOL "Switch off CppUTest Test build")
```
thetic, line 252: This is disabled by default no when used in other projects. I think we can delete this entirely.

thetic: https://github.com/cpputest/cpputest/blob/f8b2b8c28c1fe28c7558256cf7be083b79d82ff7/CMakeLists.txt#L42-L43

mtfurlan: Sounds good, removed it

thetic (approved):

PR #1841, opened 2025-01-20
Fix pipeline

thetic:
- Suppress new [bugprone-casting-through-void](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/casting-through-void.html) warning from clang-tidy. We can't fix this without using C++ casts.
- Workaround lcov regression
- Repair cygwin build


basvodde:
Nice, thanks!

PR #1845, opened 2025-03-05
fix expectedCallsLeft after actual call

blackvladimir:
I want to check if all expected calls were called without failing test. `expectedCallsLeft`  returns true between `actualCall` and `checkExpectations`:

```c++
mock().expectedCallsLeft(); // false
mock().expectOneCall("boo");
mock().expectedCallsLeft()); //true
mock().actualCall("boo");
mock().expectedCallsLeft()); //TRUE BUT SHOULD BE FALSE
mock().checkExpectations();
mock().expectedCallsLeft()); //false
```

This PR fixes it by evaluating last actuall call inside `expectedCallsLeft`.

### DIFF
--- a/src/CppUTestExt/MockSupport.cpp
+++ b/src/CppUTestExt/MockSupport.cpp
@@ -254,6 +254,7 @@ const char* MockSupport::getTraceOutput()
 
 bool MockSupport::expectedCallsLeft()
 {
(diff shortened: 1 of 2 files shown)


basvodde:
Oh, thanks. Interesting.

PR #1848, opened 2025-04-14
add DEF_SOURCE_LINE to ctests

malsyned:
GoogleTest sets this property on ctest tests, and vscode-cmake-tools will use it to dramatically improve the Test Explorer user experience.

CppUTest doesn't record file paths or line numbers for test groups, only for tests, so when `CPPUTEST_TESTS_DETAILED=0`, use the lowest line number of the tests in that group.


malsyned:
Just saw there were some new commits on `master` after pushing this PR, so I rebased and force-pushed it. It should be good to go now.

malsyned:
A note for anyone trying to use this with Visual Studio Code: you'll have to install the pre-release version of **CMake Tools** (`ms-vscode.cmake-tools`) for this to work for you, at least until they put out a release that contains [vscode-cmake-tools:3ab4e0dc](https://github.com/microsoft/vscode-cmake-tools/commit/3ab4e0dc87aef3935b0fa14f89939ee177e56ac2).

cmake/Modules/_CppUTestDiscovery.cmake
```diff
@@ -1,4 +1,4 @@
-set(script)
+set(script "" CACHE INTERNAL "")
```
thetic, line 1: why does this need to be added to the cache? It's overwritten every time `add_command` is called?

malsyned: When `add_command()` was only called directly from file scope, it could get away with using `set(PARENT_SCOPE)` to set this global variable. In my PR, it gets called from a new function, `add_test_to_script()`, so the changes it made to `script` were getting lost every time `add_test_to_script()` returned.

The only way that I've been able to find to have a true "global" variable in the way that `script` is, is to make it a `CACHE INTERNAL` variable. I'm somewhat new to CMake, though. If you know an easier or more idiomatic way to accomplish this, I'll happily update this PR (and several of my own personal `CMakeFiles` as well :laughing:).

cmake/Modules/_CppUTestDiscovery.cmake
```diff
@@ -41,17 +52,37 @@ if(NOT ${result} EQUAL 0)
...
-        "${test_name}"
-        ${EMULATOR}
-        "${EXECUTABLE}"
-        ${ARGS}
-        ${select_arg}
-        ${test_name}
-    )
-endforeach()
+
+set(LL_LINE_REGEX "^([^.]*)\\.([^.]*)\\.(.*)\\.([^.]*)\n")
```
thetic, line 56: nit: i think every regex warrants a comment explaining what it's capturing.

thetic:
I left a bit of feedback, but directionaly I think this is good and I like the idea. @basvodde are you able to kick off the workflow?

cmake/Modules/_CppUTestDiscovery.cmake
```diff
@@ -41,17 +52,37 @@ if(NOT ${result} EQUAL 0)
...
+string(REGEX MATCHALL "[^\n]+\n" discovered_test_lines "${discovered_tests}")
+if(TESTS_DETAILED)
+    foreach(line IN LISTS discovered_test_lines)
+        string(REGEX MATCH "${LL_LINE_REGEX}" __ign "${line}")
```
thetic, line 60: the meaning of `__ign` is not apparent to me

malsyned: Yeah, fair. I don't know why I didn't use the standard term "unused". Fixed.

cmake/Modules/_CppUTestDiscovery.cmake
malsyned: I like it. Fixed.

malsyned:
@thetic thanks for taking the time to review this. I think I've addressed all of the points you raised. I'm wrapping up for today. Let me know if there's anything else you want me to tweak about this PR, I'll get to it in the next day or so.

thetic:
It all looks good to me. Thanks for the contribution. We'll have to wait for Bas to merge.

basvodde:
Ok. I'll merge.

Looking at the CMake code. I wonder... is there a unit test framework for CMake also? It seems to include more and more logic...

malsyned:
> Looking at the CMake code. I wonder... is there a unit test framework for CMake also? It seems to include more and more logic...

CMake has a little tool that they use internally, [`Tests/RunCMake/RunCMake.cmake`](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Tests/RunCMake/RunCMake.cmake?ref_type=heads), that sets up a bare bones framework for using CTest to test CMake modules. It looks like at least one project, Qt6, is using it by copying that file into their own tree, along with a lightly modified form of the macro [`add_RunCMake_test()` from `Tests/RunCMake/CMakeLists.txt`](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Tests/RunCMake/CMakeLists.txt?ref_type=heads#L9). It's BSD-3-Clause, if you wanted you could just copy it into your source tree and use it to add tests of your CMake scripts to your existing CTest suite.

basvodde:
Might be interesting to consider. Part of being a unit test framework is showing that you can write unit tests for certain things. So, it could be an example for others.

Issue #1851, opened 2025-04-29
gcc 15.1 breaks leak detection


thetic:
https://gcc.gnu.org/onlinedocs/gcc-15.1.0/gcc/C_002b_002b-Dialect-Options.html#index-fassume-sane-operators-new-delete

```
 8/85 Test  #8: MemoryLeakWarningThreadSafe ....................................................***Failed    0.01 sec
TEST(MemoryLeakWarningThreadSafe, turnOnThreadSafeNewDeleteOverloads)
D:/a/cpputest/cpputest/tests/CppUTest/MemoryLeakWarningTest.cpp:497: error: Failure in TEST(MemoryLeakWarningThreadSafe, turnOnThreadSafeNewDeleteOverloads)
	LONGS_EQUAL(storedAmountOfLeaks + 4, MemoryLeakWarningPlugin::getGlobalDetector()->totalMemoryLeaks(mem_leak_period_all)) failed
	expected <9 (0x9)>
	but was  <5 (0x5)>

 - 0 ms
TEST(MemoryLeakWarningThreadSafe, turnOnThreadSafeNewDeleteOverloadsDebug) - 0 ms
TEST(MemoryLeakWarningThreadSafe, turnOnThreadSafeMallocFreeReallocOverloadsDebug) - 0 ms

Errors (1 failures, 880 tests, 3 ran, 16 checks, 0 ignored, 877 filtered out, 0 ms)
```

https://github.com/cpputest/cpputest/actions/runs/14722153889/job/41317813068?pr=1850#step:6:26

basvodde:
Is this one now fixed?

thetic:
it should be, yes.

PR #1852, opened 2025-04-29
Change codegen for gcc 15.1+

thetic:
Leak detection requires modification of global state during allocation which GCC 15.1 now optimizes out by default[^1]:

> The `-fassume-sane-operators-new-delete` option has been added and enabled by default. This option allows control over some optimizations around calls to replaceable global operators new and delete. **If a program overrides those replaceable global operators and the replaced definitions read or modify global state visible to the rest of the program, programs might need to be compiled with `-fno-assume-sane-operators-new-delete`.**

  This change detects GCC 15.1 and disables that optimization. The CMake build only affects the CppUTest project, not user builds. We have no way of injecting such a flag in users' projects.

Mitigates #1851.

[^1]: https://gcc.gnu.org/gcc-15/changes.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,17 @@ if(NOT PROJECT_IS_TOP_LEVEL)
   set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" PARENT_SCOPE)
 endif()
 
+
+if(
+  (NOT CPPUTEST_MEM_LEAK_DETECTION_DISABLED)
+  AND (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "15.1")
+)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -91,6 +91,17 @@ if(NOT PROJECT_IS_TOP_LEVEL)
   set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" PARENT_SCOPE)
 endif()
 
+
+if(
+  (NOT CPPUTEST_MEM_LEAK_DETECTION_DISABLED)
+  AND (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "13.1")
```
thetic, line 98:
```suggestion
  AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "15.1")
```

PR #1857, opened 2025-05-16
Add support for long int and unsigned long int in Mock setData functions

Steven-Butz:
For MockSupport getData functions, MockNamedValue and MockValue_c already support long int and unsigned long int.

However, there is no equivalent setData function that accepts these types. I have added support for long int and unsigned long int (in C++ and C), following the exact format and structure of the existing types.

Personally, this was useful when trying to pass a time_t as data to a mocked time() function, without running the risk of lossy typecasting / year 2038 problem.

### DIFF
--- a/include/CppUTestExt/MockSupport.h
+++ b/include/CppUTestExt/MockSupport.h
@@ -81,6 +81,8 @@ public:
     void setData(const SimpleString& name, bool value);
     void setData(const SimpleString& name, int value);
     void setData(const SimpleString& name, unsigned int value);
+    void setData(const SimpleString& name, long int value);
+    void setData(const SimpleString& name, unsigned long int value);
     void setData(const SimpleString& name, const char* value);
     void setData(const SimpleString& name, double value);
     void setData(const SimpleString& name, void* value);
--- a/include/CppUTestExt/MockSupport_c.h
+++ b/include/CppUTestExt/MockSupport_c.h
@@ -204,6 +204,8 @@ struct SMockSupport_c
     void (*setBoolData) (const char* name, int value);
     void (*setIntData) (const char* name, int value);
     void (*setUnsignedIntData) (const char* name, unsigned int value);
+    void (*setLongIntData) (const char* name, long int value);
+    void (*setUnsignedLongIntData) (const char* name, unsigned long int value);
     void (*setStringData) (const char* name, const char* value);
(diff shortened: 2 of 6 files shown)


thetic:
I like this, but we'll need to add getters too and then add this to the mock support tests.

Steven-Butz:
> I like this, but we'll need to add getters too and then add this to the mock support tests.

Oddly, there are already getters for these data types. Just no setters.

In MockNamedValue.cpp:

`long int MockNamedValue::getLongIntValue() const
{
    if(type_ == "int")
        return value_.intValue_;
    else if(type_ == "unsigned int")
        return (long int)value_.unsignedIntValue_;
    else
    {
        STRCMP_EQUAL("long int", type_.asCharString());
        return value_.longIntValue_;
    }
}

unsigned long int MockNamedValue::getUnsignedLongIntValue() const
{
    if(type_ == "unsigned int")
        return value_.unsignedIntValue_;
    else if(type_ == "int" && value_.intValue_ >= 0)
        return (unsigned long int)value_.intValue_;
    else if(type_ == "long int" && value_.longIntValue_ >= 0)
        return (unsigned long int)value_.longIntValue_;
    else
    {
        STRCMP_EQUAL("unsigned long int", type_.asCharString());
        return value_.unsignedLongIntValue_;
    }
}`

And for C, the data is just accessed as a union, that also already includes these types (MockSupport_c.h):
`typedef struct SMockValue_c
{
    MockValueType_c type;
    union {
        int boolValue;
        int intValue;
        unsigned int unsignedIntValue;
        long int longIntValue;
        unsigned long int unsignedLongIntValue;
#if CPPUTEST_USE_LONG_LONG
        cpputest_longlong longLongIntValue;
        cpputest_ulonglong unsignedLongLongIntValue;
#else
        char longLongPlaceholder[CPPUTEST_SIZE_OF_FAKE_LONG_LONG_TYPE];
#endif
        double doubleValue;
        const char* stringValue;
        void* pointerValue;
        const void* constPointerValue;
        void (*functionPointerValue)(void);
        const unsigned char* memoryBufferValue;
        void* objectValue;
        const void* constObjectValue;
    } value;
} MockValue_c;`

basvodde:
I know these functions are quite trivial.... yet they still miss test?

Could we have these before integrating?

Steven-Butz:
Unit tests are added for both C++ and C.

PR #1863, opened 2025-07-18
Add a configuration macro to disable defining new macro

notable-equivalent:
If the macro `CPPUTEST_DISABLE_NEW_KEYWORD_MACRO` is defined, do not define `new` as a macro.

In most cases it is useful to have the memory leak detector plugin enabled. However in some cases the `new` macro causes issues with standard library/other headers (mainly things that use placement-new). The suggested workarounds for including standard library headers before the `MemoryLeakDetectorNewMacros.h` file are not always feasible, since other headers can include standard library headers and the memory leaker detector header, and making sure we've imported every problematic header beforehand can be a burden (and difficult to maintain in the face of multiple compiler versions being used).

Other solutions considered:
* Just don't include `MemoryLeakDetectorNewMacros.h`
  * It is automatically included by certain files like `GTest.h` and `TestHarness.h` (through `MemoryLeakWarningPlugin.h`)
* Globally disable detection with something like `#define CPPUTEST_USE_MEM_LEAK_DETECTION 0` or `#define CPPUTEST_MEM_LEAK_DETECTION_DISABLED`
  * These macros actually disable the `new` and `delete` operator overloads which detect memory leaks, which is not what we want; We want to still be able to detect memory leaks, even if we can't get specific file and line information since we need to disable the macro.

### DIFF
--- a/include/CppUTest/MemoryLeakDetectorNewMacros.h
+++ b/include/CppUTest/MemoryLeakDetectorNewMacros.h
@@ -25,8 +25,8 @@
 #if CPPUTEST_USE_MEM_LEAK_DETECTION && defined(__cplusplus)
 
 /* This #ifndef prevents <new> from being included twice and enables the file to be included anywhere */
-#ifndef CPPUTEST_USE_NEW_MACROS
-
+#ifndef CPPUTEST_HAVE_ALREADY_DECLARED_NEW_DELETE_OVERLOADS
+    #define CPPUTEST_HAVE_ALREADY_DECLARED_NEW_DELETE_OVERLOADS 1
     #if CPPUTEST_USE_STD_CPP_LIB
         #ifdef CPPUTEST_USE_STRDUP_MACROS
             #if CPPUTEST_USE_STRDUP_MACROS == 1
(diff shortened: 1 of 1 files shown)


include/CppUTest/MemoryLeakDetectorNewMacros.h
```diff
@@ -25,8 +25,8 @@
 #if CPPUTEST_USE_MEM_LEAK_DETECTION && defined(__cplusplus)
 
 /* This #ifndef prevents <new> from being included twice and enables the file to be included anywhere */
-#ifndef CPPUTEST_USE_NEW_MACROS
```
notable-equivalent, line 28: We want the `CPPUTEST_USE_NEW_MACROS` macro to only be defined in situations where we actually defined the `new` macro. Since we include `<new>` even when we don't define the `new` macro (since we still want to declare the overloads of `operator new` and `operator delete`), we need to use a separate macro to guard the inclusion of `<new>`, in this case `CPPUTEST_HAVE_ALREADY_DECLARED_NEW_DELETE_OVERLOADS`

include/CppUTest/MemoryLeakDetectorNewMacros.h
```diff
@@ -69,27 +69,33 @@
...
+    #ifdef __clang__
+        #pragma clang diagnostic push
+        #if (__clang_major__ == 3 && __clang_minor__ >= 6) || __clang_major__ >= 4
+            #pragma clang diagnostic ignored "-Wkeyword-macro"
+        #endif
+    #endif
 
-#define new new(__FILE__, __LINE__)
+    #undef new
+    #define new new(__FILE__, __LINE__)
```
notable-equivalent, line 90: Use `#undef` so we don't cause an error if this header is included twice.

Could do `#ifndef new` instead, it depends on how we want to interact with any other library which might potentially define `new` as a macro (either we overwrite their definition with our definition by using `#undef`, or we keep their definition and don't define our own `new` macro by using `#ifdef`). I don't think it matters much since I don't think many other libraries are defining `new` as a macro, and because the previous behavior of this file would just cause an error in such a case anyway.

notable-equivalent:
Oh that's very confusing, because `CPPUTEST_USE_NEW_MACROS` already exists but it works differently from `CPPUTEST_USE_MALLOC_MACROS`.

I don't think we can safely change `CPPUTEST_USE_NEW_MACROS` to work the same way without breaking backwards compatibility. People rely on the ability to `#undef new` and then reinclude the header to define the macros again. See for example https://cpputest.github.io/manual.html#:~:text=Conflicts%20with%20my%20own%20overload 

[GTest.h](https://github.com/cpputest/cpputest/blob/4363c47a86125c8b3557e58de0032d16a2c7f7ea/include/CppUTestExt/GTest.h) is also relying on that behavior, although confusingly it appears to be trying to do the same thing for the `malloc` and `strdup` macros which actually don't work like that.

If we really want to use the same macro, there's a few options for how the `MemoryLeakDetectorNewMacros.h` file could work:
* If `CPPUTEST_USE_NEW_MACROS` is undefined or 1, define the `new` macro and set `CPPUTEST_USE_NEW_MACROS` to 1. If `CPPUTEST_USE_NEW_MACROS` is 0, don't define the `new` macro.
  * I think this is a bit overly complicated and could be confusing for users.
* Use a separate macro (something like `CPPUTEST_INTERNAL_ALREADY_DEFINED_NEW_MACROS`) to keep track of whether or not we defined the `new` macro, and if that macro is set then even if `CPPUTEST_USE_NEW_MACROS` is already set we still define the `new` macros again.
  * This one is still a bit convoluted, but at least to users it should appear the same so that's a plus. 

I think I favor option two out of those. But if we do that, we would probably also want to update `MemoryLeakDetectorMallocMacros.h` to work the same way.

What are your thoughts?

basvodde:
Interesting. I think I can live with these changes, but let's see if all the builds pass :)

PR #1865, opened 2025-08-12
Document -ri CLI arg

seanmil2:
# Summary

Add explanation for the run ignored tests cli argument.

I noticed that the -ri argument is in the print usage but not in the verbose help and I was not sure what this option did:

image

## Testing

image


notable-equivalent:
Can this be merged @basvodde ?

Issue #1855, opened 2025-05-12
state of CI and latest-passing-build


mtfurlan:
The last [latest passing build](https://github.com/cpputest/cpputest/releases/tag/latest-passing-build) was in 2021, and it looks like Travis [hasn't run](https://app.travis-ci.com/github/cpputest/cpputest/builds) since then.
I think we should make it start working again, because #1651 says use 4.0 (5 years ago) or latest passing build(4 years ago)

It looks like the current ci state is
* [github actions test report](https://github.com/cpputest/cpputest/blob/master/.github/workflows/test-report.yml)
  * makes a nice display for all the junit tests in the basic workflow
* [github actions basic workflow](https://github.com/cpputest/cpputest/blob/master/.github/workflows/basic.yml)
  * clang-format
  * codeql
  * autotools autoreconf, configure, and make
      * mac: make tdd with gcc & clang
      * ubuntu: make tdd with gcc, clang, and `--disable-longlong`
      * ubuntu: make dist and make install
      * windows-latest make tdd
  * cmake: cmake & cmake --build && ctest, and sometimes make --install && build examples against install
    * ubuntu:
      * clang CXX standards 98, 11, 14, 17, 20
      * gnu CXX standards 98, 14, 17, 20
      * gnu install with CXX 11
      * presets
        * No long long
        * Detailed
        * No C++ Standard Library
        * No RTTI
        * No Exceptions
        * No C Standard Library
        * Address Sanitizer
        * GCC ARM embedded
        * clang-tidy
        * mingw
        * arm-embedded
        * OpenWatcom DOS with dosbox
      * run lcov
    * macos
      * install
      * clang CXX standards 14, 17, 20
      * gcc CXX standard 14
    * windows:
      * Cygwin with preset GNU
      * ClangCL CXX standard 17
      * MSVC CXX standard 20
      * MSVC install
  * make: under ubuntu do `make -f Makefile_using_MakefileWorker` with the following targets
    * test_defaults
    * test_disable_std_cpp
    * test_disable_memory_leak_detection
    * test_disable_std_cpp_and_leak_detection
    * test_disable_debug
    * test_override_flags
    * test_disable_std_c
    * test_target_platform
    * test_override_and_disable
    * test_examples
    * test_junit_output
    * test_vpath
    * test_vpath_install
    * commented out: test_gcov
  * make_dos
    * build with dosbox and github.com/cpputest/watcom-compiler and platforms/Dos/Makefile
[...]

PR #1869, opened 2025-10-03
ci: latest-passing-build

mtfurlan:
bring back the latest-passing-build tag

addresses #1855
    
to see it in action before merging, see
https://github.com/mtfurlan/cpputest/releases/tag/latest-passing-build

### DIFF
--- a/.github/workflows/latest-passing-release.yml
+++ b/.github/workflows/latest-passing-release.yml
@@ -1,0 +1,61 @@
+name: 'Latest Passing Release'
+# run this once basic build passes tests on master branch
+# it will update the latest-passing-build tag, and update the github release
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


notable-equivalent:
The date was already not being updated back when this repo published the latest passing build (right now the "release date" of the latest passing build is April 2020, but the build is actually from November 2021). So given that this was already an existing problem, I think it's acceptable to still have that problem, at least for now.

EDIT: Apparently GitHub sorts releases in the feed based on the date of the tag/commit and not the "release date" of the GitHub releases itself, so I even more so feel that it's fine to leave it like this https://github.com/orgs/community/discussions/5447#discussioncomment-1555826

basvodde:
Thanks

Issue #1829, opened 2024-11-20
Can we rename HAVE_CONFIG_H?


mtfurlan:
I'm working with [asn1c](https://github.com/mouse07410/asn1c) and CppUTest, and both projects use `HAVE_CONFIG_H` to mean things, and CppUTest defining it as a public compile definition is causing issues for the asn1c code

I would really like if we could rename the one in this repo to `CPPUTEST_HAVE_CONFIG_H`
From the comments in the file, it looks like it isn't supposed to be user facing, but searching github issues shows that I personally have had to define it manually in the past

So can we deprecate `HAVE_CONFIG_H` in favour of `CPPUTEST_HAVE_CONFIG_H` but support both?
```
#if defined(HAVE_CONFIG_H) || defined(CPPUTEST_HAVE_CONFIG_H)
```

mtfurlan:
asn1c is renaming their HAVE_CONFIG_H to HAVE_ASN1C_CONFIG_H: https://github.com/mouse07410/asn1c/pull/202

mtfurlan:
Okay so it turns out it's not easy, `HAVE_CONFIG_H` is coming from [autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Configuration-Headers.html)

So I think the fix is to not expose the define publicly in [CMakeLists.txt](https://github.com/cpputest/cpputest/blob/master/CMakeLists.txt#L129-L132), that should be PRIVATE not PUBLIC

basvodde:
Yes, I think I looked at changing the HAVE_CONFIG_H in autoconf and I couldn't find it :( The proposed solution sounds good to me! Probably @thetic has an opinion too.

thetic:
The macro needs to be public. For example, TestHarness.h includes CppUTestConfig.h includes CppUTestGeneratedConfig.h which needs `HAVE_CONFIG_H` to include the generated header.

thetic:
I agree that the generic name is prone to collisions, but I think we would need to convince autoconf to use something else. I'm no expert, but i'll see what i can find in the docs.

PR #1872, opened 2025-11-15
Rename HAVE_CONFIG_H

thetic:
Address #1829.

I know roughly nothing about autoconf, so please double check my work.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,11 @@ EXTRA_PROGRAMS = CppUTestExtTests
 lib_LIBRARIES = lib/libCppUTest.a
 check_PROGRAMS = $(CPPUTEST_TESTS)
 
-# Change autoconf's hard-coded (and collision susceptible) HAVE_CONFIG_H macro.
+# HACK: Replace the HAVE_CONFIG_H macro with CPPUTEST_HAVE_GENERATED_CONFIG_H.
+# The name HAVE_CONFIG_H is a hard-coded in autoconf and cannot be changed in
+# any "normal" ways. This isn't typically a problem, but CppUTest exposes the
+# macro in its public headers. This leads to collisions with any other projects
+# built with autoconf.
 DEFS = $(filter-out -DHAVE_CONFIG_H,@DEFS@) -DCPPUTEST_HAVE_GENERATED_CONFIG_H
 
 if INCLUDE_CPPUTEST_EXT


Makefile.am
```diff
@@ -10,6 +10,9 @@ EXTRA_PROGRAMS = CppUTestExtTests
 lib_LIBRARIES = lib/libCppUTest.a
 check_PROGRAMS = $(CPPUTEST_TESTS)
 
+# Change autoconf's hard-coded (and collision susceptible) HAVE_CONFIG_H macro.
+DEFS = $(filter-out -DHAVE_CONFIG_H,@DEFS@) -DCPPUTEST_HAVE_GENERATED_CONFIG_H
```
basvodde, line 18: Whoa....

basvodde: After a bit more checking on this, perhaps we can try out AX_PREFIX_CONFIG_H in the configure.ac.

My AI friend recommended to do this:
AC_CONFIG_HEADERS([config.h:config.hin], [AM_CPPFLAGS="$AM_CPPFLAGS -DCPPUTEST_HAVE_CONFIG_H"])

Which also feels like a hack... but at least a hack in the right place as this really shouldn't be in the am file :)

thetic: > After a bit more checking on this, perhaps we can try out AX_PREFIX_CONFIG_H in the configure.ac.

I tried this and it only appears to affect the name of the resulting header, not the macro.

> My AI friend recommended to do this:
AC_CONFIG_HEADERS([config.h:config.hin], [AM_CPPFLAGS="$AM_CPPFLAGS -DCPPUTEST_HAVE_CONFIG_H"])

I'll need to double check, but I think this only adds the `CPPUTEST_HAVE_CONFIG_H` macro; it doesn't remove the `HAVE_CONFIG_H` macro which we don't want leaking into other libraries.

thetic: > My AI friend recommended to do this:
AC_CONFIG_HEADERS([config.h:config.hin], [AM_CPPFLAGS="$AM_CPPFLAGS -DCPPUTEST_HAVE_CONFIG_H"])

That doesn't build. I tried `AC_CONFIG_HEADERS([config.h], [AM_CPPFLAGS="$AM_CPPFLAGS -DCPPUTEST_HAVE_CONFIG_H"])` and confirmed that defines only `HAVE_CONFIG_H` with `DEFS = -DHAVE_CONFIG_H` in the generated makefile.

basvodde: Ok, thanks for checking. I did some more checking, and I tried the following in the configure.ac:

-AC_CONFIG_HEADERS([config.h])
+AC_CONFIG_HEADERS([CppUTestGeneratedConfig.h])

In a quick test, that seems to do the trick. But I didn't checked in throughly yet. Could you check that?

thetic: that does not work. The name of `HAVE_CONFIG_H` is hardcoded in autoconf[^1][^2]. I don't think config headers were ever intended to be exposed in libraries' public interfaces.

[^1]: https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Configuration-Headers.html
[^2]: https://github.com/autotools-mirror/autoconf/blob/8ac9edd1a71fe24ed94d70f1669e4fd6a81fa585/lib/autoconf/status.m4#L1267

basvodde: Yeah, the problem comes from building a unit testing framework as library.

I'd like to merge this then, but probably need a bit of a comment to it to explain what is happening, why, and a tip that if we find a better version to then remove this. Will you give that a try or shall I add it?

thetic: no problem. see 7f9c2455c5f5666805612626f1d648b2807a1109

basvodde:
Thanks for your patience!

PR #1874, opened 2025-11-17
Fix new clang warnings

thetic:
Found with clang 21.1.14:

```
error: variable 'out1' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
```


basvodde:
That's a fair warning.

basvodde:
Thanks! As always!

PR #1881, opened 2026-03-03
Fix autotools cmake package install for non-standard libdir

### LINKED ISSUE #1563: CppUTestConfig.cmake in libcpputest-dev can't find CppUTest_INCLUDE_DIRS
When trying to use `find_package(CppUTest REQUIRED)` using the `CppUTestConfig.cmake` file provided by `libcpputest-dev` fails on trying to locate the include directory. From what I can tell, the `PACKAGE_PREFIX_DIR` that is generated doesn't account for the `x86_64-linux-gnu` directory that is used by Debian multiarch.

It also looks like the `GNUInstallDirs.cmake` module attempts to accommodate Debian and append the architecture to the libdir. This doesn't appear to be happening in the release build. I think using `libdir` in `Makefile.am` doesn't account for debian multiarch as indicated here: https://wiki.debian.org/Multiarch/Implementation#autotools_with_no_helper

### LINKED ISSUE #1880: Defective cmake config if building with autotools
If the build is being done with autotools instead of cmake, and if the `--libdir` argument to to `configure` specifies something else than `${prefix}/lib`, then the resulting files `CppUTestConfig.cmake`, `CppUTestTargets.cmake` and `CppUTestTargets-relwithdebinfo.cmake` will contain invalid include-paths and library-paths.

The issue is that the autotools build uses the pre-created files in the `build/cmake_package_files` directory but does not adjust these if `libdir` is not `lib`. This configuration is common for Debian-based systems by now.

Example of how to create the problematic files (and correct files for comparision):

```sh
cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -D CMAKE_INSTALL_PREFIX=/usr -D CMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -B cmake-build
cmake --build cmake-build
cmake --install cmake-build --prefix ${PWD}/cmake-destdir/usr

autoreconf -i
(mkdir autotools-build && cd autotools-build &&  ../configure --prefix=/usr --libdir=\${prefix}/lib/x86_64-linux-gnu)
make -C autotools-build -j $(nproc)
make -C autotools-build install DESTDIR=${PWD}/autotools-destdir

diff -u *-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestConfig.cmake
diff -u *-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets.cmake
diff -u *-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets-relwithdebinfo.cmake
```

This would output (among other differences):

```diff
--- autotools-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestConfig.cmake
+++ cmake-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestConfig.cmake
@@ -3,4 +3,4 @@
 ####### Any changes to this file will be overwritten by the next CMake run ####
 ####### The input file was CppUTestConfig.cmake.install.in                            ########

-get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)
+get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../../" ABSOLUTE)
```

```diff
--- autotools-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets.cmake
+++ cmake-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets.cmake
@@ -43,6 +57,7 @@

 # Compute the installation prefix relative to this file.
 get_filename_component(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
+get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
```

```diff
--- autotools-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets-relwithdebinfo.cmake
+++ cmake-destdir/usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestTargets-relwithdebinfo.cmake
@@ -9,21 +9,21 @@
 set_property(TARGET CppUTest APPEND PROPERTY IMPORTED_CONFIGURATIONS RELWITHDEBINFO)
 set_target_properties(CppUTest PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELWITHDEBINFO "CXX"
-  IMPORTED_LOCATION_RELWITHDEBINFO "${_IMPORT_PREFIX}/lib/libCppUTest.a"
+  IMPORTED_LOCATION_RELWITHDEBINFO "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libCppUTest.a"
   )

 list(APPEND _IMPORT_CHECK_TARGETS CppUTest )
-list(APPEND _IMPORT_CHECK_FILES_FOR_CppUTest "${_IMPORT_PREFIX}/lib/libCppUTest.a" )
+list(APPEND _IMPORT_CHECK_FILES_FOR_CppUTest "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libCppUTest.a" )

 # Import target "CppUTestExt" for configuration "RelWithDebInfo"
 set_property(TARGET CppUTestExt APPEND PROPERTY IMPORTED_CONFIGURATIONS RELWITHDEBINFO)
 set_target_properties(CppUTestExt PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELWITHDEBINFO "CXX"
-  IMPORTED_LOCATION_RELWITHDEBINFO "${_IMPORT_PREFIX}/lib/libCppUTestExt.a"
+  IMPORTED_LOCATION_RELWITHDEBINFO "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libCppUTestExt.a"
   )

 list(APPEND _IMPORT_CHECK_TARGETS CppUTestExt )
-list(APPEND _IMPORT_CHECK_FILES_FOR_CppUTestExt "${_IMPORT_PREFIX}/lib/libCppUTestExt.a" )
+list(APPEND _IMPORT_CHECK_FILES_FOR_CppUTestExt "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libCppUTestExt.a" )

 # Commands beyond this point should not need to know the version.
 set(CMAKE_IMPORT_FILE_VERSION)
```

Example of what happens if a demo cmake-project tries to use this file:

```
CMake Error at /usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestConfig.cmake:11 (message):
  File or directory /usr/lib/include referenced by variable
  CppUTest_INCLUDE_DIRS does not exist !
Call Stack (most recent call first):
  /usr/lib/x86_64-linux-gnu/CppUTest/cmake/CppUTestConfig.cmake:27 (set_and_check)
  demoproject/CMakeLists.txt:123 (find_package)
```

thetic:
Fixes #1880.
Fixes #1563.

On Debian multiarch systems, `--libdir` is typically `lib/x86_64-linux-gnu` rather than `lib`. The pre-generated cmake package files assumed a fixed directory depth to compute the install prefix, breaking `find_package(CppUTest)` for any non-standard `--libdir`.

Rather than trying to detect the directory depth at cmake time, generate the cmake package files at install time with `%%PREFIX%%` and `%%LIBDIR%%` substituted directly from the autotools `$(prefix)` and `$(libdir)` make variables. This works correctly regardless of how deep `libdir` sits relative to the prefix.

While here, bring the autotools cmake install into parity with the CMake install: add missing cmake modules, use the `CppUTest::` target namespace, and reorganize all package/install templates into a single `pkg/` directory.

A new CI job (`Autotools Custom prefix install`) installs with a non-standard `--libdir` and verifies the result by building the examples against it with `find_package`.


basvodde:
Thanks. Sorry for the late merge.

PR #1886, opened 2026-05-09
CHECK_EQUAL: fix array != array compare under C++26

dale-stewart:
## Problem

C++26 [over.match.oper]/p10 makes `array == array` (and `array != array`) ill-formed. `CHECK_EQUAL_LOCATION` expanded its two side-effect-detection probes literally as `(operand) != (operand)`. As soon as either operand was an array — most commonly a string literal in `CHECK_EQUAL("text", x)` — the inner self-compare became a forbidden array-vs-array compare.

The same pattern is also a `-Wdeprecated-array-compare` error under C++20+ when `-Werror` is on. CppUTest's existing tests don't exercise the array-operand path, so this never showed up in CI; it surfaces immediately in any consuming project that does.

## Fix

A small `CppUTestPrivate::checkEqualSelfNotEqual<T>(T, T)` helper that takes both arguments by value, so any array decays to a pointer before the comparison runs. The two side-effect probes (`(actual) != (actual)` and `(expected) != (expected)`) route through it. Both arguments come from the same operand expression, so `T` is identical on both sides — `-Wsign-compare` cannot fire.

The outer `(expected) != (actual)` is intentionally left at the macro call site. Routing it through a function template would lose `-Wsign-compare`'s constant-fits exception that lets idioms like `CHECK_EQUAL(0, some_unsigned)` build clean today.

Behavior is unchanged for every non-array operand. For arrays, the resulting pointer-address self-compare matches the historical C++03 behavior (always false, since the same array's decayed pointer equals itself).

## Test

Added `CHECK_EQUAL_compilesWithArrayOperand` in `TestUTestMacro.cpp`. Without the fix, the test TU emits `-Wdeprecated-array-compare` under C++20+ with `-Werror`, and is ill-formed at C++26 regardless of warning flags. With the fix it compiles at every standard from C++98 through C++26.

## CI coverage

Verified by the existing `-Wall -Werror` jobs (Make Defaults, CMake Linux GNU 17/20/23, etc.) — `-Wdeprecated-array-compare` is on by default at C++20+ and gates the relevant condition. A future PR could add a C++26 job once compiler C++26 support stabilizes; this fix doesn't depend on that addition.

## Compatibility

Works at C++98 through C++26. No new headers included. The helper is a plain function template — no `<type_traits>`, no `if constexpr`, no forwarding refs.

### DIFF
--- a/include/CppUTest/UtestMacros.h
+++ b/include/CppUTest/UtestMacros.h
@@ -129,20 +129,11 @@
 #define CHECK_EQUAL_TEXT(expected, actual, text)\
   CHECK_EQUAL_LOCATION(expected, actual, text, __FILE__, __LINE__)
 
-namespace CppUTestPrivate
-{
-    template <typename T>
-    bool checkEqualSelfNotEqual(T a, T b)
-    {
-        return a != b;
-    }
-}
-
 #define CHECK_EQUAL_LOCATION(expected, actual, text, file, line)\
   do { if ((expected) != (actual)) { \
-      if (::CppUTestPrivate::checkEqualSelfNotEqual((actual), (actual))) \
+      if (StringFrom(actual) != StringFrom(actual)) \
       	  UtestShell::getCurrent()->print("WARNING:\n\tThe \"Actual Parameter\" parameter is evaluated multiple times resulting in different values.\n\tThus the value in the error message is probably incorrect.", file, line); \
-      if (::CppUTestPrivate::checkEqualSelfNotEqual((expected), (expected))) \
+      if (StringFrom(expected) != StringFrom(expected)) \
       	  UtestShell::getCurrent()->print("WARNING:\n\tThe \"Expected Parameter\" parameter is evaluated multiple times resulting in different values.\n\tThus the value in the error message is probably incorrect.", file, line); \
       UtestShell::getCurrent()->assertEquals(true, StringFrom(expected).asCharString(), StringFrom(actual).asCharString(), text, file, line); \
   } \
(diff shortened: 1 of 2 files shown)


basvodde:
Thanks for the PR. Since in CppUTest we don't use templates, I'll need to look for an alternative solution. But, I verified the problem.

basvodde:
I've checked this further. As we will not add templated code to CppUTest, this means there is no fix for this (afaik). However, the issue can be fixed relative easily by adding a cast in the call. So, I'm leaning towards just fixing this in the call towards CHECK_EQUAL and not in CppUTest itself. This should be easy as it gives a very explicit compiler error.

Will that be ok with you? Thanks for the PR anyways...

ps. I did added C++26 support in the build so we can find incompatibilities as early as possible.

basvodde:
I've checked this further. As we will not add templated code to CppUTest, this means there is no fix for this (afaik). However, the issue can be fixed relative easily by adding a cast in the call. So, I'm leaning towards just fixing this in the call towards CHECK_EQUAL and not in CppUTest itself. This should be easy as it gives a very explicit compiler error.

Will that be ok with you? Thanks for the PR anyways...

ps. I did added C++26 support in the build so we can find incompatibilities as early as possible.

dale-stewart:
Good catch.

Replaced it with a `StringFrom`-based probe, no template:

```c
if (StringFrom(actual) != StringFrom(actual))
```

Arrays decay to pointers at the call boundary, so no array-to-array comparison is formed, and `StringFrom` is already a documented requirement for `CHECK_EQUAL` operands. Net diff is +2/-2. Rationale, the two trade-offs, and the verification runs are in the commit message.

Verified at C++98/23/26 with `-Wall -Wextra -Werror` -- clean build, suite green.

dale-stewart:
I just noticed your second reply.  I've offered a partial solution, do with it whatever you like.  I no longer have the need I originally had for this change.

basvodde:
The old solution uses the operator != and it is documented to expect that. Doing the comparison with StringFrom breaks that, hence I do not prefer that (also it will make it a lot slower). I'll be not accepting this PR and instead expect it to be fixed at the call rather than in CppUTest

PR #1889, opened 2026-07-14
Fix unsigned underflow in SimpleString::subString() on empty strings

94xhn:
`SimpleString::subString(size_t beginPos, size_t amount)` rejects an out-of-range `beginPos` with:

```cpp
if (beginPos > size()-1) return "";
```

`size()` returns `size_t` (unsigned). When called on an empty string, `size()` is `0`, so `size()-1` underflows to `SIZE_MAX`. The check `beginPos > SIZE_MAX` is then false for essentially any `beginPos`, so the bounds check is silently defeated and the function falls through to:

```cpp
SimpleString newString = getBuffer() + beginPos;
```

which reads out of bounds of the (1-byte, null-terminator-only) buffer of an empty `SimpleString`.

Repro (confirmed with a guard-page allocator that places an empty SimpleString's buffer directly against a `PAGE_NOACCESS` page):

```cpp
SimpleString empty("");
SimpleString result = empty.subString(1, 1); // reads 1 byte past the buffer
```

This reliably crashes before the fix and returns `""` after it.

Fix: change the check to `beginPos >= size()`, which is mathematically identical to the original for every `size() >= 1` (so every existing non-empty-string call site is byte-for-byte unaffected) but no longer underflows when `size() == 0`.

Verified:
- Compiled a minimal repro harness against the unmodified sources: `empty.subString(1, 1)` segfaults before the fix.
- Recompiled the same harness against the fixed source: no crash, `subString(1, 1)` on an empty string correctly returns `""`.
- Rebuilt and ran the existing `SimpleString` test group (`AllTests -g SimpleString`, 130 tests / 238 checks) against both the original and fixed source — identical `OK` results before and after, so this is a pure bug fix with no behavior change for any existing test or non-empty-string caller.

### DIFF
--- a/src/CppUTest/SimpleString.cpp
+++ b/src/CppUTest/SimpleString.cpp
@@ -580,7 +580,7 @@ void SimpleString::padStringsToSameLength(SimpleString& str1, SimpleString& str2
 
 SimpleString SimpleString::subString(size_t beginPos, size_t amount) const
 {
-    if (beginPos > size()-1) return "";
+    if (beginPos >= size()) return "";
 
     SimpleString newString = getBuffer() + beginPos;
 


basvodde:
Thanks!

